PR #31, opened 2025-09-03
feat: auth(z) based on Service Account tokens

bartoszmajsak:
This PR proposes the use of Service Account tokens as "api keys".

Access control is handled directly through standard Kubernets `Role`s and `RoleBinding`s and uses built-in token generation mechanism (TokenRequest API) for ephemeral keys while using the same rate limit policies.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Service Account token authentication with tier-based namespaces, RBAC-driven authorization, and per-user rate limiting.
* **Tools**
  * Added scripts to create/retrieve SA tokens, decode JWTs, and run rate-limit load tests with example API calls.
* **Documentation**
  * Expanded guide on SA token auth, tier/user mappings, setup, testing, and switching from API keys.

### DIFF
--- a/deployment/kuadrant/README.md
+++ b/deployment/kuadrant/README.md
@@ -640,8 +640,7 @@ This solution replaces a secret-based key system with an approach that integrate
 
 ### User Accounts and Tiers
 
-For the demo, the tiers/groups are mapped as following:
-
+For the demo, the tiers/groups are mapped as follows:
 | Tier           | Users                          | Rate Limit  | Service Account Namespace           |
 |----------------|--------------------------------|-------------|-------------------------------------|
 | **Free**       | `freeuser1`, `freeuser2`       | 5 req/2min  | `inference-gateway-tier-free`       |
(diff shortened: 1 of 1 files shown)


deployment/kuadrant/service-account-tokens/05-tier-access.yaml
```diff
@@ -0,0 +1,52 @@
...
+kind: Role
+metadata:
+  name: models-user
+  namespace: llm
+rules:
+  - apiGroups: ["serving.kserve.io"]
+    resources: ["inferenceservices"]
+    # Can be limited to selected models if needed
+    # resourceNames: ["vllm-simulator"] 
+    verbs: ["post"]
```
pierDipi, line 30: we could create this as part of the llmisvc (as we're targeting that API) reconciliation (in odh-model-controller)

deployment/kuadrant/service-account-tokens/05-tier-access.yaml
```diff
@@ -0,0 +1,52 @@
+## Namespaces are modeling the tiered access.
+## Each Service Account (delegated identity) will be created only
+## in one of them - that means given user is subscribed to a single tier.
+
+apiVersion: v1
+kind: Namespace
```
jland-redhat, line 6: One question I had and don't totally follow is why do we have to create a namespace for each? 

Is it not possible to just have a single namespace and 3 different groups that we connect the RoleBindings too?

jland-redhat (commented):
Just had one question we can talk about on Monday. Sorry it took so long to get this review in.

PR #53, opened 2025-09-09
Feature/enhanced policy manager

noyitz:
policy manager shows nows full data of the policy.


yossiovadia (approved):
LGTM ( tested locally )

PR #80, opened 2025-09-15
Fix/remove hardcoded user references

yossiovadia:
Reming noyitz from few hardcoded files

## Summary by CodeRabbit

* New Features
  - Experimental QoS mode with single- and multi-tier load testing, QoS headers, and response badges.
  - App-wide Experimental Mode provider to toggle experimental UI/features.
  - Token management now associates tokens with the authenticated user; auto-selects the first available token.
* Improvements
  - Simulator traffic uses a configurable default user instead of a hardcoded value.
  - More flexible CORS configuration to ease connectivity across environments.
* Style
  - Added multi-size/favicon formats (ICO, PNG 32/16, SVG) for better browser support.


noyitz (approved):
thanks for the fix!

PR #81, opened 2025-09-15
Add an --all flag for installs to install all components

nerdalert:
Adds an `-all` flag for deployment and testing convenience.


jland-redhat (approved):
LGTM

PR #89, opened 2025-09-18
feat(llmisvc): sample deployment for llmisvc with simulator

bartoszmajsak:
The manifests in `maas-api` have been re-organized to provide:

- Core infra pieces (`deploy/infra`)
  - ODH Kserve installation
  - Openshift-specific GatewayClass to automatically install OCP-supported GW API machinery
  - Basic Kuadrant installation
- Sample polices (`deploy/policies`)
- [`llm-d-simulator`](https://github.com/llm-d/llm-d-inference-sim)-backed model for testing (`models/simulator`)

Follow updated `DEV.md` to test/validate it.

## Summary by CodeRabbit

- New Features
  - Infrastructure bundles for Gateway API, Kuadrant, and ODH; demo LLM simulator added.
  - Gateway-level token rate limiting by user tier.

- Changes
  - Authorization now derives namespace/name from request path; identity fields renamed/remapped.
  - Admin API key placeholder set; dev token lifetime extended to 4h.
  - Networking: Gateway introduced; legacy route and simulator inclusion removed.
  - RBAC and model resource names adjusted for LLM services.

- Documentation
  - Environment setup, deployment, and testing guides reorganized and clarified.

- Chores
  - Dependency install script: streamlined readiness checks (single operator wait, no fallback).

### DIFF
--- a/maas-api/deploy/policies/auth-policy.yaml
+++ b/maas-api/deploy/policies/auth-policy.yaml
@@ -45,6 +45,10 @@ spec:
               auth.identity.user.username.split(":")[3]
     authorization:
       tier-access:
+        when:
+          # Target models only (TODO: find a better way to do this)
+          - predicate: |
+              !request.path.startsWith("/maas-api")
         kubernetesSubjectAccessReview:
           user:
             expression: auth.identity.user.username
@@ -60,7 +64,7 @@ spec:
                 request.path.split("/")[1]
             name:
               expression: |
-                request.path.split("/")[1]
+                request.path.split("/")[2]
             verb:
               value: post
     response:


maas-api/deploy/models/simulator/simulated-model.yaml
```diff
@@ -0,0 +1,54 @@
...
+    containers:
+      - name: main
+        image: "ghcr.io/llm-d/llm-d-inference-sim:v0.5.1"
+        imagePullPolicy: Always
+        command: ["/app/llm-d-inference-sim"]
+        args:
+        - --port
+        - "8000" 
+        - --model
+        - facebook-opt-125m-single-simulated
```
bartoszmajsak, line 22: @israel-hdez I just realized this is bug in `/v1/models` listing. We should read  `.spec.model.name` from `LLMInferenceService`. Can be done as follow-up to #88

israel-hdez: It depends on UI expectation when we properly integrate to ODH. I remember I used the name of the resource, because that's the name the ODH Dashboard currently shows.  I wouldn't expect UI showing different names in different screens.

Although, I see that for LLMIsvc, we configure the HTTPRoute using the resource name ([ref](https://github.com/opendatahub-io/kserve/blob/fc411e66a20a4f2af72908eafd6e99e322e979f1/config/llmisvc/config-llm-router-route.yaml#L29)), while vLLM engine is configured with `.spec.model.name` ([ref](https://github.com/opendatahub-io/kserve/blob/fc411e66a20a4f2af72908eafd6e99e322e979f1/config/llmisvc/config-llm-template.yaml#L19-L20)) which would be reflected in the body of the request :thinking:....

This is not going to be a nice UX for `LLMIsvcs`, given for `ISVCs` users are used to take from the UI the name and consistently use the same string in both the URL and the request body. Is there any reason why we are having this split in the names used?

maas-api/deploy/infra/odh/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+metadata:
+  name: odh-infra
+
+# For now, install only Model Serving pieces
```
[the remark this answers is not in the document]
bartoszmajsak: We will, not yet

maas-api/deploy/models/simulator/simulated-model.yaml
pierDipi: Decoupling instance name from model name helps with traffic splitting https://docs.google.com/document/d/1x6aI9pbTF5oOsaEQYc9n4pBBY3_AuEY2X51VKxmBSnU/edit?tab=t.0#heading=h.2ct4hd2ndg7g, by default if `spec.model.name` is not specified, we'll fallback to use `metadata.name`

PR #99, opened 2025-09-24
chore(make): no need to check REPO variable

bartoszmajsak:
It is always defaulted and if explictly set to blank the push will fail regardless.


israel-hdez (approved):
I'm good with it.

PR #105, opened 2025-09-26
Updated CI to push image on merge

chaitanya1731:
Updated CI to push image on merge

## Description

Added logic for `maas-api` image push to Quay.io repo on every push event
 
## Merge criteria:

### DIFF
--- a/.github/workflows/maas-api-ci.yml
+++ b/.github/workflows/maas-api-ci.yml
@@ -1,6 +1,9 @@
 name: MaaS API
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
     paths:
       - 'maas-api/**'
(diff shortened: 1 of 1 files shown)


Jaland:
Can we add some logic around git tagging that pushes a specific tagged image where the image tag matches the git tag.

Jaland:
So this would not be on a merge event it would be on a tag even. And you should be able to save the `GITHUB_REF` environment variable.

Something like this:

```
on:
  push:
    tags: 
      - '*'
      
jobs:
  build-and-push-image:
  ...
      - name: build_and_push
        uses: docker/build-push-action@v5
        with:
          tags: quay.io/${{ github.repository }}:${{ github.ref_name }}
```

There is also something apparently called `docker/metadata-action` that does this as well

PR #106, opened 2025-09-26
Updated deployment to explicity set the PROVIDER env variable. Helpin…

jland-redhat:
Updated deployment to explicitly set the PROVIDER env variable.

This should be defaulted with the new image, but this makes it more obvious what the provider is set to and that the provider is a updatable value.

## Summary by CodeRabbit

* **Chores**
  * Updated deployment configuration to include a new provider environment setting.
  * Aligns runtime configuration across environments to reduce drift and improve stability.
  * Enhances operational reliability for token-related workflows under varying deployments.
  * No user-facing behavior changes; background configuration only.

### DIFF
--- a/maas-api/deploy/base/deployment.yaml
+++ b/maas-api/deploy/base/deployment.yaml
@@ -22,6 +22,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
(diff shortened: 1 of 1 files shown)


bartoszmajsak (approved):
/lgtm

PR #112, opened 2025-09-30
[RHOAIENG-34715] - Expose model listing endpoint as top-level

spolti:
Update base model endpoint to be MAAS_GATEWAY_URL/v1/models.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added a new top-level endpoint for fetching model listings at /v1/models and routing to support it alongside existing /maas-api endpoints.

* **Documentation**
  * Updated model discovery instructions to reference /v1/models for fetching available models and noted no changes to behavior or error handling.

### DIFF
--- a/deployment/base/maas-api/networking/httproute.yaml
+++ b/deployment/base/maas-api/networking/httproute.yaml
@@ -15,7 +15,6 @@ spec:
             value: /v1/models
       backendRefs:
         - name: maas-api
-          namespace: platform-services
           port: 8080
           weight: 100
     # All other maas-api endpoints remain under /maas-api prefix


jland-redhat (approved):
LGTM

deployment/base/maas-api/networking/httproute.yaml
```diff
@@ -8,6 +8,17 @@ spec:
...
       namespace: openshift-ingress
   rules:
+    # OpenAI-compatible /v1/models endpoint at top level
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /v1/models
+      backendRefs:
+        - name: maas-api
+          namespace: platform-services
```
bartoszmajsak, line 18:
```suggestion
```

The route is in the same namespace, no need to be explicit. And also this ns does not exist anymore :)

spolti:
Just had a quick chat with @bartoszmajsak, and keeping both endpoints is good enough.

jland-redhat:
So just a point of clarity here that I have been thinking about. This makes it closer to the standard but not actually follow it I don't think, right?

Like the normal flow is you hit the `/v1/models` endpoint then you can hit the `/v1/chat/completion` with the model info for chat completions. But with us you need to look up the `url` returned from `/v1/models` and hit that.

I like the direction but did just want to make sure that part was clear when we document this

PR #118, opened 2025-10-01
chore: Moving sample models into docs folder

jland-redhat:
Moved the sample models to the root directory. And simplified the readme

## Summary by CodeRabbit

* **Documentation**
  * Consolidated and reorganized sample model deployment docs; added a clearer samples guide with deployment steps and environment-variable usage and updated README links to point to the new samples location.
  * Removed an outdated duplicate sample models document.

* **Bug Fixes**
  * Fixed malformed YAML list entry in RBAC kustomization to ensure correct parsing during deployments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,4 +208,4 @@ Built with:
 For questions or issues:
 - Open an issue on GitHub
 - Check the [deployment guide](deployment/README.md) for troubleshooting
-- Review the [samples](samples/models/) for examples
+- Review the [samples](docs/samples/models) for examples


bartoszmajsak (changes requested):
I wonder if putting those models under top-level `docs/samples/models` wouldn't help in organizing user-facing content and examples?

I also think that without `rbac/` piece, which this PR removes, we can't use AuthPolicy (and subsequent (token)rate limiting) for these models, they're accessible for everyone.

bartoszmajsak (approved):
LGTM

PR #121, opened 2025-10-02
chore: simplifies http routing

bartoszmajsak:
We only need one route for maas-api, routes for models are handled by LLMInferenceService controller, so creating separate one can cause confusion.

## Summary by CodeRabbit

* New Features
  * Unified access under /maas-api with automatic URL rewrite for cleaner URLs.
  * Gateway-level authentication is enforced for maas-api.

* Refactor
  * Renamed the maas-api route and removed host-based routing; previous maas-api.<domain> URLs are no longer served—use the /maas-api path instead.
  * Removed domain-based routes for certain model endpoints (e.g., qwen3 and simulator), consolidating traffic through the primary gateway path.


jland-redhat (approved):
LGTM

PR #123, opened 2025-10-02
chore(kustomize): consolidate TokenRateLimitPolicy

bartoszmajsak:
Defines single TokenRateLimitPolicy to be consistently across the whole setup.

It's also aligned with other policies to be deployed in the same namespace where the Gateway is defined.


jland-redhat (approved):
LGTM

PR #125, opened 2025-10-02
docs: guides for subscription tiers and token mgmt

bartoszmajsak:
This PR introduces two new user-facing guides:

- **Subscription Tiers**: explains concepts, architecture, tier-to-group mappings, RBAC design, rate limiting and troubleshooting.

- **Token Management**: describes how users obtain, use, and revoke service account tokens, including lifecycle management.

Both guides include diagrams, examples, and verification steps to make setup and usage more approachable.


docs/tiers.md
```diff
@@ -0,0 +1,483 @@
...
+RBAC controls which tiers (via their namespaces) can access models. This is done by creating a `Role` with model access 
+permissions and a `RoleBinding` to link that `Role` to the tier namespaces.
+
+**Example (for all models in the namespace):**
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: model-user
```
pierDipi, line 340: nit: perhaps "model-access" or "model-post-access"?

docs/tiers.md
```diff
@@ -0,0 +1,483 @@
...
+
+---
+
+### Rate Limiting
+
+The `RateLimitPolicy` enforces different quotas per tier using the `auth.identity.tier` field that was enriched by the `AuthPolicy`.
+
+**Request-based limits:**
+
+```yaml
```
pierDipi, line 392: maybe a little description/note that this example applies rate limits to a specific model server but the same can be applied at the Gateway level ?

docs/tiers.md
```diff
@@ -0,0 +1,487 @@
...
+- Basic understanding of JWT authentication
+
+---
+
+## Concepts Overview
+
+### Core Terminology
+
+#### Tier  
+A subscription level that determines class of service (e.g., `free`, `premium`, `enterprise`).
```
nerdalert, line 49: Thought I remember small, medium large tshirt sizing, good with any of them just in case this was an autogen!

nerdalert (approved):
LGTM

jland-redhat (commented):
I think this is a really good first step and I am fine if we want to merge this.

But I do think we should break this document up with specific target audiences in mind, cause right now we have a lot of high level concepts mixed with really deep in the weeds info about how the solution works.

My thought for breaking up the docs would be:
* High Level Concepts
* Basic Usage
* Advance Usage
* Deeper Dives on how the different flows work.

docs/tiers.md
bartoszmajsak: I am well aware, but I do not think we have a decision about naming yet.

docs/tiers.md
```diff
@@ -0,0 +1,487 @@
...
+##### Mapping  
+Configuration, currently stored in a Kubernetes ConfigMap, that defines tiers and their associated Kubernetes groups. 
+The mapping is used by `maas-api` and serves as the single source of truth for resolving user groups to tiers.
+
+#### Instance Name
+
+A unique identifier for a Model-as-a-Service component, typically matching the name of the Gateway resource (e.g., openshift-ai-inference). 
+This name is used to ensure that configurations for different gateways do not conflict.
+
+#### Tier Namespace
```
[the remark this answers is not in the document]
bartoszmajsak: > Especially stuff like this Tier Namespace, I think we are trying to teach them how it works when I think this section should just be stuff like

I think this is an important concept than needs to be captured. Maybe in advanced doc - but we need to have it.

docs/tiers.md
```diff
@@ -0,0 +1,487 @@
+# Configuring Subscription Tiers
```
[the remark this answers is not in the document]
bartoszmajsak: Makes sense, I hope we can evolve these docs further.

docs/tiers.md
```diff
@@ -0,0 +1,487 @@
...
+1. [Configuration Steps](#configuration-steps)
+1. [Verifying Your Configuration](#verifying-your-configuration)
+1. [Common Configuration Mistakes](#common-configuration-mistakes)
+1. [Related Documentation](#related-documentation)
+
+---
+
+## Prerequisites
+
+Before configuring tiers, ensure you have:
```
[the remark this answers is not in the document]
bartoszmajsak: I would appreciate if you could add GH ```suggestion``` so that the link is to the exact doc you have in mind :)

bartoszmajsak:
> But I do think we should break this document up with specific target audiences in mind, cause right now we have a lot of high level concepts mixed with really deep in the weeds info about how the solution works.
> 
> My thought for breaking up the docs would be:
> 
> * High Level Concepts
> * Basic Usage
> * Advance Usage
> * Deeper Dives on how the different flows work.

Very good feedback, thanks @jland-redhat. I think we should evolve this documents further and I hope we can collaborate to shape them. I am merging this as a good basis and let's make it better!

PR #130, opened 2025-10-02
fix: readding maas-api service account and removing ref to non-existing secret

jland-redhat:
Somewhere along the line we lost the maas-api service account so adding it back in.

And [this pr](https://github.com/opendatahub-io/maas-billing/pull/124) removes the admin secret but we still had some references to it that needed to be removed

## Summary by CodeRabbit

- Chores
  - Added a dedicated ServiceAccount for the MaaS API and included it in the base deployment.
  - Removed automatic generation and injection of the ADMIN_API_KEY secret from the overlay and deployment.
  - Prepares clearer identity management and secret handling for deployments.
  - No user-facing changes; runtime features and functionality remain unchanged.

### DIFF
--- a/maas-api/deploy/overlays/secret/kustomization.yaml
+++ b/maas-api/deploy/overlays/secret/kustomization.yaml
@@ -9,7 +9,14 @@ resources:
 
 patches:
 - path: patches/secrets-provider.yaml
+- path: patches/api-key-env-var.yaml
 - path: patches/clusterrole.yaml
   target:
     kind: ClusterRole
     name: maas-api
+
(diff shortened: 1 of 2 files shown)


deployment/base/maas-api/kustomization.yaml
bartoszmajsak, line 1: Maybe we can remove this overlay entirely?

maas-api/deploy/overlays/secret/kustomization.yaml
bartoszmajsak, line 1: This overlay is unrelated, but I think we can remove it entirely

bartoszmajsak (approved):
LGTM

PR #132, opened 2025-10-02
refactor(kustomize): consolidate deployment manifests

bartoszmajsak:
> [!IMPORTANT]
> We should be using `ODH` version of KServe - especially since certain features are not upstreamed yet (or even released there). That's why `kserve` folder is removed - to avoid confusion.

This change introduces a single source of truth to make deployments simpler.

Existing `maas-api/deploy` manifests are promoted to top-level `deployment` folder, reducing duplication and improving the structure.

Installation script for OpenShift has been adjusted:

- Fixed OCP version check - using cluster info instead as it's more reliable
- Add helpers to wait for CRDs/pods and a version comparator;
- Auto-detect and patch AuthPolicy audience; apply gateway policies in sequence after infra becomes ready.
- Detect existing ODH/RHOAI KServe; deploy components only when missing, then proceed with retries.

## Summary by CodeRabbit

* **New Features**
  * Version-aware deployment flow with CRD/pod readiness checks, clearer status messages, and in-cluster AuthPolicy audience patching.

* **Refactor**
  * Streamlined deployment composition; added container resource limits, readiness/liveness probes and improved termination behavior.

* **Chores**
  * Removed legacy KServe/OpenShift configs, RBAC, gateway/policy manifests, and several sample model deployments/overlays.

* **Documentation**
  * Updated developer guidance for the new deployment flow and token/audience extraction.


deployment/base/maas-api/deployment.yaml
```diff
@@ -1,59 +1,36 @@
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: maas-api
-  namespace: maas-api
-  labels:
-    app: maas-api
-    version: v2
```
bartoszmajsak, line 9: Injected using kustomize

deployment/base/maas-api/deployment.yaml
```diff
@@ -1,59 +1,36 @@
...
-          value: "debug"
+        - name: PROVIDER
+          value: sa-tokens
+        resources:
+          requests:
+            memory: "64Mi"
+            cpu: "50m"
+          limits:
+            memory: "128Mi"
+            cpu: "200m"
```
bartoszmajsak, line 33: Maybe we can tweak it

maas-api/Makefile
```diff
@@ -119,17 +119,13 @@ define deploy
 		trap 'mv kustomization.yaml.backup kustomization.yaml 2>/dev/null || true' EXIT INT TERM && \
 		kustomize edit set image maas-api=$(FULL_IMAGE) && \
 		$(if $(PRE_DEPLOY_STEP),$(PRE_DEPLOY_STEP) &&) \
-		kustomize build . | kubectl apply -f -
+		kustomize build . --load-restrictor LoadRestrictionsNone | kubectl apply -f -
```
bartoszmajsak, line 122: Only needed for dev overlay as we refer to top-level `deployment` manifests

deployment/base/maas-api/deployment.yaml
```diff
@@ -1,59 +1,36 @@
...
         livenessProbe:
```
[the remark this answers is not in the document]
bartoszmajsak: They're injected by Kustomize when rendering - see kustomization.yaml

bartoszmajsak (commented):
General note: we should be using `ODH` version of KServe - especially that certain features are not upstreamed yet (or even released there). That's why `kserve` folder is removed - to avoid confusion.

deployment/base/maas-api/deployment.yaml
jland-redhat: This is small enough where I don't think it matters unless it is causing us issues.

jland-redhat (approved):
Ok so the idea is that our Deployment Pieces will live in the `/deployment` folder. And we include an overly in the `maas-api` folder for that references those pieces for development purposes if I understand everything correctly.

I don't know that we need the dev overlay in the `maas-api` folder. But I am fine with keeping it for now.

Approved but just want to check that we ran the script on a fresh cluster to validate deployment?

PR #133, opened 2025-10-03
chore: Add odh-model-controller as infra component

israel-hdez:
Strictly, `odh-model-controller` would not act as infrastructure
component. It will provide automation for creating the required Roles and
RoleBindings for giving access to the models enrolled as-a-service.
However, this seems to be the best place to put it in.

## How Has This Been Tested?

Run `kustomize build maas-api/deploy/infra/odh`

## Merge criteria:


bartoszmajsak:
Before we merge it we probably need two things

- Gateway segregation - to not use the default one LLMInferenceService will attache "other" instances to
- Ensure the policies are not conflicting

/hold

jland-redhat:
Just an additional thought

jland-redhat (approved):

PR #134, opened 2025-10-03
refactor(scripts): align test script with DEV.md authentication flow

bartoszmajsak:
Use MaaS API token endpoint and model discovery to match the documented developer workflow, making tests consistent with production usage and self-adapting to deployed models.


jland-redhat (approved):
Awesome!

I would like to integrate this script in to the deployment README at some point as well.

PR #139, opened 2025-10-03
feat: adds labels for better discovery

bartoszmajsak:
## Summary by CodeRabbit

- New Features
  - Standardized labels for token-related namespaces and service accounts for improved consistency and discoverability.
- Refactor
  - Replaced legacy label construction with dedicated helper functions and removed the obsolete label utility.
- Tests
  - Updated tests to use the test framework’s context and switched to the current fake Kubernetes client for test interactions.


dmytro-zaharnytskyi (approved):
Nothing to add other than Coderabbit comment on username, as option - replace username label with {user-id: <short-hash>} and move raw username to an annotation.

But from other side, we can assume that there is no invalid username exists in maas.opendatahub.io/ :)

israel-hdez (approved):
LGTM

jland-redhat (approved):
LGTM

PR #173, opened 2025-10-20
chore: setup cleanup - remove unneeded Routes

israel-hdez:
The `gateway-route` is not required because the Gateway API stack should properly expose the Gateway. The `maas-api` is also not needed, because there is an `HTTPRoute` exposing the maas-api through Gateway API.

Also, remove `deployment/overlays/README.md` because its contents are covered in other md files.

## Summary by CodeRabbit

* **Chores**
  * Removed OpenShift Routes configuration files and associated deployment automation. Updated overlay manifests and deployment scripts accordingly. Removed related platform-specific documentation.


jland-redhat (commented):
LGTM

PR #212, opened 2025-11-12
feat: added granite model

dmytro-zaharnytskyi:
Added IBM Granite model to the pull of testing models we ship with MaaS. 

Tested it by deploying to the cluster with MaaS byt using kustomize command -> verified model works and responds + we hit rate limits

## Summary by CodeRabbit

* **New Features**
  * Added IBM Granite 2B (GPU) as a deployable model option with GPU-backed serving and sample manifests.
  * Made the Qwen3 sample serve name dynamic to follow the selected model configuration.

* **Documentation**
  * Updated sample docs and README to list the new IBM Granite 2B GPU model and reflect it in the deployment example comments.

### DIFF
--- a/docs/samples/models/ibm-granite-2b-gpu/model.yaml
+++ b/docs/samples/models/ibm-granite-2b-gpu/model.yaml
@@ -26,34 +26,25 @@ spec:
         operator: Exists
     containers:
       - name: main
+        # Specific vLLM version for Tesla T4 compatibility
         image: vllm/vllm-openai:v0.6.3.post1
         command: ["python3", "-m", "vllm.entrypoints.openai.api_server"]
         args:
           - "--model=/mnt/models"
-          # Model name used in API calls (e.g., in the "model" field of requests)
-          # Uses the model name from spec.model.name
-          - "--served-model-name={{ .Spec.Model.Name }}"
+          - "--served-model-name=ibm-granite/granite-3.1-2b-instruct"
+          - "--port=8000"
+          - "--ssl-keyfile=/etc/ssl/certs/tls.key"
+          - "--ssl-certfile=/etc/ssl/certs/tls.crt"
+          # Granite-specific overrides
           - "--max-model-len=4096"
           - "--trust-remote-code"
-          - "--ssl-keyfile=/var/run/kserve/tls/tls.key"
-          - "--ssl-certfile=/var/run/kserve/tls/tls.crt"
-          - "--port=8000"
           - "--dtype=float16"
         env:
-          - name: VLLM_LOGGING_LEVEL
-            value: INFO
-          - name: HF_HUB_CACHE
-            value: /models-cache
+          # Granite-specific environment variables for Tesla T4 compatibility
           - name: VLLM_ATTENTION_BACKEND
             value: XFORMERS
           - name: VLLM_USE_V1
             value: "false"
-        volumeMounts:
-          - name: cache
-            mountPath: /models-cache
-          - name: tls-certs
-            mountPath: /var/run/kserve/tls
-            readOnly: true
         resources:
           limits:
             cpu: '2'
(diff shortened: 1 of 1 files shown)


docs/samples/models/ibm-granite-2b-gpu/model.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - name: maas-default-gateway
+          namespace: openshift-ingress
+  template:
+    containers:
+      - name: main
+        image: vllm/vllm-openai:v0.6.3.post1
+        command: ["python3", "-m", "vllm.entrypoints.openai.api_server"]
+        args:
+          - "--model=/mnt/models"
+          - "--served-model-name=granite-2b-instruct"
```
bartoszmajsak: We usually do in other examples for consistency. Is `granite-2b-instruct` what we want to use here or the name we define above is fine too?

```suggestion
          - "--served-model-name \"{{ .Spec.Model.Name }}\""
```

bartoszmajsak: By looking at our samples I wonder if we can follow the same approach and unify as follow-up?

docs/samples/models/ibm-granite-2b-gpu/model.yaml
```diff
@@ -0,0 +1,91 @@
...
+            readOnly: true
+        resources:
+          limits:
+            cpu: '2'
+            memory: 12Gi
+            nvidia.com/gpu: '1'
+          requests:
+            cpu: '500m'
+            memory: 8Gi
+            nvidia.com/gpu: '1'
```
bartoszmajsak: Do we want to ensure the GPU node too?

```
nodeSelector:
  nvidia.com/gpu.present: "true"
tolerations:
   - effect: NoSchedule
     key: nvidia.com/gpu
     operator: Exists
```

docs/samples/models/ibm-granite-2b-gpu/model.yaml
```diff
@@ -0,0 +1,103 @@
...
+  #     # Example: Custom GPU utilization metric (requires GPU monitoring like NVIDIA DCGM)
+  #     - type: Pod
+  #       pod:
+  #         metric:
+  #           name: gpu-utilization  # Replace with your actual metric name
+  #         target:
+  #           type: AverageValue
+  #           averageValue: "80"  # Scale when GPU utilization exceeds 80%
+  #   targetRequests: 100
+  #   scaleUpRate: 10
```
israel-hdez: There is no `autoscaler` field in `LLMInferenceService`. I'm not sure why this is here.

docs/samples/models/ibm-granite-2b-gpu/model.yaml
```diff
@@ -0,0 +1,56 @@
...
+    nodeSelector:
+      nvidia.com/gpu.present: "true"
+    tolerations:
+      - effect: NoSchedule
+        key: nvidia.com/gpu
+        operator: Exists
+    containers:
+      - name: main
+        # Specific vLLM version for Tesla T4 compatibility
+        image: vllm/vllm-openai:v0.6.3.post1
```
israel-hdez: It is better, but I still do not understand why we need to use this version.

The base templates are using Inference Server 3.2.2, which is using the more recent vLLM 0.10.1.1. So, I don't see a reason for overriding the `image` with an older one... unless the newer one doesn't support Tesla? But that would be weird. Also, I don't see why the need to override the `command`.

The `args` shouldn't be overriden that much. Any additional args should be passed, instead, via a `VLLM_ADDITIONAL_ARGS` environment variable (see [here](https://github.com/opendatahub-io/kserve/blob/release-v0.15/config/llmisvc/config-llm-template.yaml#L143))

dmytro-zaharnytskyi: updated, you are right, tested without args and it works, I downgraded an image - cause it didn't work first and I thought that were cuda version compatibility, but apparently i fixed the issue some other way. Thanks!

PR #216, opened 2025-11-12
test: ignore apps directory in Snyk Code

israel-hdez:
This adds a new `.snyk` file with a rule to omit the `apps/` directory from Code Analysis. This is because the content of this directory is not being delivered in builds.

Related to https://issues.redhat.com/browse/RHOAIENG-38235

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **Chores**
  * Updated Snyk configuration to explicitly exclude the apps directory from security code scans, clarifying the exclusion in the configuration and reducing noise from those scans.

PR #218, opened 2025-11-12
chore: remove apps directory

israel-hdez:
This:

- Reverts #216, which added a Snyk configuation to exclude `apps` from code analisys
- Removes the apps directory, as an alternative to #216, since we no longer require it. The MaaS front-end will be part of ODH dashboard.

Related to https://issues.redhat.com/browse/RHOAIENG-38356

## Release Notes

* **Revert**
  * Removed the backend application, frontend user interface, and QoS service along with their associated configurations, documentation, and deployment scripts.


jland-redhat (approved):
LGTM

PR #221, opened 2025-11-12
docs: added versioning to documentation

jland-redhat:
This PR configures the documentation site to build and deploy all versions to the gh-pages branch.

Once the first tag is cut, it becomes the "latest" and default version, and all subsequent tags are added to the version dropdown. The dev branch content remains accessible at the dedicated /dev path but is excluded from the dropdown.

⚠️ Action Required: The GitHub Pages settings must be updated to use the gh-pages branch as the source.

## Summary by CodeRabbit

* **New Features**
  * Documentation versioning system now supports latest, dev, and tagged versions
  * Enhanced documentation features including instant navigation, tabbed content, task lists, and diagram support

* **Documentation**
  * Added comprehensive guides for local documentation development setup
  * Expanded deployment documentation with version management workflow instructions

### DIFF
--- a/.github/workflows/docs-version.yml
+++ b/.github/workflows/docs-version.yml
@@ -3,7 +3,7 @@ name: Version Documentation on Tag
 on:
   push:
     tags:
-      - 'v*.*.*'  # Triggers on version tags like v1.0.0, v2.1.3, etc.
+      - '*.*.*'  # Triggers on version tags like 1.0.0, 2.1.3, etc.
   workflow_dispatch:
     inputs:
       version:


.github/workflows/docs-version.yml
```diff
@@ -0,0 +1,75 @@
+name: Version Documentation on Tag
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'  # Triggers on version tags like v1.0.0, v2.1.3, etc.
```
bartoszmajsak: We might want to reconsider using the `v` prefix. 

Unless we intend to publish parts of this repo as a Go library, a plain `x.y.z` version may communicate more clearly that the code isn’t meant for reuse. 

If someone really wants to depend on it, they can rely on Go pseudo-versions - with no support or guarantees. This is the approach Istio takes for its core components (which is a headache for consumer, but I do not see this happening here soon).

Thoughts?

PR #229, opened 2025-11-17
fix: remove RateLimitPolicy from the default deployment

nerdalert:
- Leaving the manifest and line in deployment/base/policies/kustomization.yaml as a reference if the user wants to enable it.
- Showcasing TRLP as the default is probably the most appropriate install.

** Note: ** leaving this in draft until we resolve the flapping of TRLP limits as the demos are basically working from RateLimitPolicy rather than TokenRateLimitPolicy.

Example output without RateLimitPolicy applied where you see the random flapping between 200s and 429s.

```
$ for i in {1..16}; do curl -sSk -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d "{\"model\": \"${MODEL_NAME}\", \"prompt\": \"Hello\", \"max_tokens\": 50}" "${MODEL_URL}"; done
429
429
200
200
200
200
200
200
200
200
200
200
200
200
429
429

$ for i in {1..16}; do curl -sSk -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d "{\"model\": \"${MODEL_NAME}\", \"prompt\": \"Hello\", \"max_tokens\": 50}" "${MODEL_URL}"; done
429
429
429
429
429
429
429
429
429
429
429
429
200
200
200
200

$ for i in {1..16}; do curl -sSk -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -d "{\"model\": \"${MODEL_NAME}\", \"prompt\": \"Hello\", \"max_tokens\": 50}" "${MODEL_URL}"; done
200
200
200
200
429
200
200
200
200
200
200
200
200
200
200
200
```

## Summary by CodeRabbit

* **Chores**
  * Disabled the rate-limiting policy by default in configuration and added clarifying comments about enabling it.
  * Noted that the rate-limiting policy can be enabled via configuration and may be run alongside token-based rate limiting when desired.

### DIFF
--- a/deployment/base/policies/kustomization.yaml
+++ b/deployment/base/policies/kustomization.yaml
@@ -8,4 +8,4 @@ resources:
 # Gateway policies - default for all models
 - gateway-auth-policy.yaml
 - token-limit-policy.yaml
-- rate-limit-policy.yaml
+# - rate-limit-policy.yaml - RateLimitPolicy can be run in tandem with TokenRateLimitPolicy if desired
--- a/deployment/base/policies/rate-limit-policy.yaml
+++ b/deployment/base/policies/rate-limit-policy.yaml
@@ -1,3 +1,4 @@
+# disabled by the default deployment, enable in  deployment/base/policies/kustomization.yaml
 apiVersion: kuadrant.io/v1
 kind: RateLimitPolicy
 metadata:


jland-redhat (approved):
I think this makes sense

bartoszmajsak:
@nerdalert @jland-redhat A general concern I have is that without request rate-limiting, we aren't adequately protecting the infrastructure. We need both layers of protection, also because the `maas-api` endpoint sits behind the same gateway.

We may also need to revisit how this policy is defined across different rate-limiting windows. I wonder if separating sustained request limits from short burst limits, or applying different per-minute and per-second thresholds to handle spikes without overwhelming the system will help.

nerdalert:
@bartoszmajsak great feedback. Im for leaving ratelimipolicy, I think it makes sense. We can improve docs around usage. I feel like we should nail down some user stories that are associated to the policy or some t-shirt sizing.Random story: User wants gateway-level request rate limits that enforce both burst and sustained caps per user/tier so that a sudden spike of inference calls can’t overwhelm Authorino/Limitador or backends. RateLimitPolicy on  maas-default-gateway with per-tier dual windows (e.g., 5–10 req/5s burst + current 2m limits).

PR #244, opened 2025-11-21
docs: adding token endpoint info

jland-redhat:
In Title

## Summary by CodeRabbit

* **Documentation**
  * Added "Token API Endpoints" documentation section detailing endpoints for model retrieval (GET /models, GET /v1/models), token management (POST/DELETE /v1/tokens), and tier lookup (POST /v1/tiers/lookup) with complete method descriptions, request bodies, and response information.

### DIFF
--- a/docs/content/user-guide/self-service-model-access.md
+++ b/docs/content/user-guide/self-service-model-access.md
@@ -6,6 +6,18 @@ This guide is for **end users** who want to use AI models through the MaaS platf
 
 The Model-as-a-Service (MaaS) platform provides access to AI models through a simple API. Your organization's administrator has set up the platform and configured access for your team.
 
+## Token API Endpoints
+
+The MaaS platform provides the following API endpoints:
+
+| Endpoint           | Method | Purpose                                | Request Body     | Response               |
+|--------------------|--------|----------------------------------------|------------------|------------------------|
+| `/models`          | GET    | List available Models                  | None             | OpenAI-compatible list |
+| `/v1/models`       | GET    | List available Models (OpenAI)         | None             | OpenAI-compatible list |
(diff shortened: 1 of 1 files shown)


docs/content/user-guide/self-service-model-access.md
```diff
@@ -6,6 +6,18 @@ This guide is for **end users** who want to use AI models through the MaaS platf
 
 The Model-as-a-Service (MaaS) platform provides access to AI models through a simple API. Your organization's administrator has set up the platform and configured access for your team.
 
+## Token API Endpoints
+
+The MaaS platform provides the following API endpoints:
+
+| Endpoint           | Method | Purpose                                | Request Body     | Response               |
+|--------------------|--------|----------------------------------------|------------------|------------------------|
+| `/models`          | GET    | List available Models                  | None             | OpenAI-compatible list |
```
bartoszmajsak: Maybe we don't need this? It will return all `InferenceService`s and not all are OpenAI-compatible. Moreover we don't have logic to ensure that. It's legacy endpoint. I wonder if we should remove it?


```suggestion
```

bartoszmajsak (changes requested):

jland-redhat:
> @jland-redhat Generally speaking I wonder if we really want this? Is it intended to be published for Dev Preview?
> 
> It is not a stable list and might get outdated. Are we planning some automation to keep it consistent with the codebase?

We had someone ask about how do we delete tokens and that info does not exist at all in our current docs.

I could make a section for that but I do think we need to maintain a table of our endpoints and what they do. I think a lot of the way folks are going to interact with our flow is through `curls`

bartoszmajsak:
Maybe we can point to openapi schema instead? But we would need automation for that one too :)

PR #255, opened 2025-11-25
feat: add Konflux Dockerfile with FIPS compliance

liangwen12year:
- Add Dockerfile.konflux for Konflux/Red Hat builds
- Enable FIPS 140-2 compliance (CGO_ENABLED=1, GOEXPERIMENT=strictfipsruntime)
- Use pinned SHA256 digests for base images (reproducibility)
- Include Red Hat labels for EULA compliance and metadata
- Aligns with main Dockerfile FIPS compliance changes

This Dockerfile is used by Konflux for building the maas-api container image with Red Hat-specific requirements.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated build and deployment configuration for improved container practices and security standards.

### DIFF
--- a/maas-api/Dockerfile.konflux
+++ b/maas-api/Dockerfile.konflux
@@ -1,0 +1,42 @@
+FROM registry.access.redhat.com/ubi9/go-toolset@sha256:84286c7555df503df0bd3acb86fe2ad50af82a07f35707918bb0fad312fdc193 AS builder
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+
+RUN go mod download
+
+COPY . .
+
+USER root
+RUN CGO_ENABLED=1 GOEXPERIMENT=strictfipsruntime GOOS=linux go build -trimpath -ldflags="-s -w" -o maas-api ./cmd/
+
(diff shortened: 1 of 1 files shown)


liangwen12year:
> I wonder if we need two Dockerfiles that are almost identical right now?
> 
> @israel-hdez do we also need some extra args for multi-arch? I see we inject some for [Kserve](https://github.com/red-hat-data-services/kserve/blob/ea1b3dd914fa0e9950368991529607df8e214206/Dockerfiles/controller.Dockerfile.konflux) builds. Doesn't need to be part of this PR, but we should plan follow-up if that's required.

From my point of view, we currently keep two Dockerfiles because they serve slightly different purposes and constraints: Dockerfile.konflux is tailored for the Konflux build pipeline, where we need pinned base image digests and a specific set of product/metadata labels to satisfy supply-chain, release, and compliance requirements, while the standard Dockerfile is optimized for local development, using tag-based images (e.g., :1.24, :latest) and a simpler layout so developers can iterate more easily without having to update digests or product labels on every change. Keeping them separate lets us satisfy the stricter reproducibility and labeling requirements in CI/CD without making local builds more brittle, even though the build logic inside them is intentionally kept as close as possible.

PR #266, opened 2025-11-26
feat: add comprehensive Kind local deployment for MaaS platform

noyitz:
- Complete Kind cluster setup with all required components
- LLMInferenceService-only architecture using clean simulators
- Dual test model deployment (model-a: free tier, model-b: premium tier)
- Comprehensive test suite for functionality and authentication
- Gateway API with Kuadrant integration for auth and rate limiting
- Simplified installation and validation scripts

## Summary by CodeRabbit

* **New Features**
  * Kind-based local development environment with automated install/uninstall and demo flows.
  * Local TLS-enabled gateway and Gateway API routing for services and test models.
  * Sample test models (free/premium tiers) and tiered RBAC/access controls.
  * End-to-end test scripts for auth, rate limits, token-based limits, and model inference.

* **Documentation**
  * Detailed architecture, setup, testing, and troubleshooting guides for local Kind development.

### DIFF
--- a/deployment/overlays/kind/kind-config.yaml
+++ b/deployment/overlays/kind/kind-config.yaml
@@ -35,7 +35,7 @@ nodes:
 
 # Networking configuration
 networking:
-  # Disable default CNI to allow manual CNI installation if needed
+  # Keep default CNI enabled for Kind cluster networking
   disableDefaultCNI: false
 
   # Pod and service subnets
--- a/deployment/overlays/kind/kustomization.yaml
+++ b/deployment/overlays/kind/kustomization.yaml
@@ -62,6 +62,9 @@ patches:
       - op: replace
         path: /spec/targetRef/name
         value: maas-gateway
+      - op: replace
+        path: /spec/rules/authorization/tier-access/kubernetesSubjectAccessReview/resourceAttributes/verb/value
+        value: create
 
   # Patch Gateway rate limit policies namespace and target to istio-system/maas-gateway
   - target:
--- a/deployment/overlays/kind/rbac/roles.yaml
+++ b/deployment/overlays/kind/rbac/roles.yaml
@@ -13,7 +13,7 @@ rules:
 - apiGroups: ["serving.kserve.io"]
   resources: ["inferenceservices", "llminferenceservices"]
   resourceNames: ["model-a"]
-  verbs: ["get", "list", "post"]
+  verbs: ["get", "list", "create"]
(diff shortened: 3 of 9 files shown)


deployment/scripts/kind/install.sh
```diff
@@ -0,0 +1,754 @@
...
+    # Wait for Gateway to be created and service to be provisioned
+    log_info "Waiting for Gateway to be provisioned..."
+    sleep 5
+    kubectl wait --for=condition=Programmed gateway/maas-gateway -n istio-system --timeout=300s >/dev/null 2>&1 || true
+
+    # Patch the Gateway service to use fixed NodePorts (30080, 30443)
+    # This is required for Kind's extraPortMappings to work
+    log_info "Configuring Gateway service NodePorts..."
+    kubectl patch svc maas-gateway-istio -n istio-system --type='json' -p='[
+      {"op":"replace","path":"/spec/ports/1/nodePort","value":30080},
```
liangwen12year: The script hardcodes NodePort values (30080, 30443) without validating they match the `kind-config.yaml` file. If someone changes the config file, the script will fail silently or patch incorrectly. I would recommend to add validation to extract NodePort values from `KIND_CONFIG` or verify they match before patching.

noyitz:
> > This is very cool and I like the idea.
> > But I almost wonder if it should live somewhere outside of the official repo.
> > Reason being is I feel like if we have a Kind Cluster setup here then that means we would need to support it and validate it works, meaning it really would need to be added to our e2e test.
> > I like the idea of having this one day but do we think this would make more sense to include in a BU or OCTO owned repository that we could just link to in our readme for extra examples?
> > Thoughts: @bartoszmajsak ?
> 
> I would like to understand better the goal of this work. Having `kind` support is convenient. Same as @jland-redhat I wonder how much extra effort it is to ensure we can run things smoothly there. Do we plan to have e2e tests using kind for each PR to be sure we don't break this deployment mode as the code evolves?

this is a PoC, to demonstrate a more k8s native option to deploy. it is not planned to be an official deployment we support (at least at this point), it is more for early adapters and future work as a reference. I also believe more PoCs like this will be created (for example multi tenant deployment that is in progress). I can move PoC works into octo repo, and maintain a .md file that refer to PoC and experiments. will that work?

jland-redhat:
Going to Close this as we are using it as a reference PoC.

Can always be re-opened later

PR #267, opened 2025-11-26
fix(build): ensures tools are properly resolved

bartoszmajsak:
Moves tools installation targets to `tools.mk` file and arranges variables correctly so that target prerequisites are resolved correctly.

## Summary by CodeRabbit

* **Chores**
  * Updated build system infrastructure to use a modular approach for managing development tooling dependencies.


maas-api/tools.mk
```diff
@@ -0,0 +1,28 @@
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
```
israel-hdez: Hmmm, isn't PROJECT_DIR inherited from the parent Makefile?

PR #287, opened 2025-12-02
perf(tier): adds caching for configmap lookups

bartoszmajsak:
Replaces direct calls using ConfigMapInterface with cached lister. This eliminates repeated API calls by caching configmap in local memory with automatic refresh every 8 hours.

This prevents auth pipeline from crashing when enriching metadata with tier lookup for a given user under load.

Ref [RHOAIENG-40668](https://issues.redhat.com/browse/RHOAIENG-40668)

## Summary by CodeRabbit

* **Refactor**
  * Tier lookups now use a cache-backed informer/lister for faster, cached resolution and simplified lookup calls.
  * Mapper initialization moved to be context-aware, reducing context passing at lookup sites.

* **Chores**
  * Added a configurable default resync period for background caching.

* **Tests**
  * Test helpers and fixtures updated to use the new mapper initialization and lookup behavior.

### DIFF
--- a/maas-api/internal/tier/mapper.go
+++ b/maas-api/internal/tier/mapper.go
@@ -36,16 +36,9 @@ func NewMapper(ctx context.Context, clientset kubernetes.Interface, tenantName, 
 	configMapInformer := informerFactory.Core().V1().ConfigMaps()
 	configMapLister := configMapInformer.Lister()
 
-	stopCh := make(chan struct{})
+	informerFactory.Start(ctx.Done())
 
-	go func() {
-		<-ctx.Done()
-		close(stopCh)
-	}()
-
-	informerFactory.Start(stopCh)
-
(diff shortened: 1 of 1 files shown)


maas-api/internal/tier/mapper.go
```diff
@@ -1,38 +1,53 @@
...
 	tenantName      string
-	configMapClient corev1typed.ConfigMapInterface
+	namespace       string
+	configMapLister corelisters.ConfigMapLister
 }
 
 func NewMapper(clientset kubernetes.Interface, tenantName, namespace string) *Mapper {
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		clientset,
+		30*time.Second,
```
pierDipi: This could even be O(hours)

bartoszmajsak: That makes sense, I think we can align with ctrl-runtime default resync to be 8h. I will fix it in other places too.

PR #291, opened 2025-12-03
feat: addding the ability to specify namespace

jland-redhat:
Updates to the deploy script to allow us to deploy to different namespaces for testing.

Note this also makes a change to the ODH Controller and adds the "TENANT_NAMESPACE" environment variable which is dependent on the following PR:
https://github.com/opendatahub-io/odh-model-controller/pull/616

## Summary by CodeRabbit

* **Chores**
  * Deployment now supports a configurable MaaS API namespace via an environment variable, with safer templating, namespace-aware apply/patch flow, rollout/readiness waits, guarded operator patching, cleanup and improved logging.

* **Bug Fixes**
  * Removed hard-coded namespace entries from manifests; routing, policy and config resources are now namespace-agnostic and AuthPolicy now targets the route by kind/name.

### DIFF
--- a/deployment/scripts/deploy-openshift.sh
+++ b/deployment/scripts/deploy-openshift.sh
@@ -251,10 +251,10 @@ else
     "$SCRIPT_DIR/install-dependencies.sh" --ocp --odh
 fi
 
-# Patch odh-model-controller deployment to set TENANT_NAMESPACE
+# Patch odh-model-controller deployment to set MAAS_NAMESPACE
 # This should be done whether ODH was just installed or was already present
 echo ""
-echo "   Setting TENANT_NAMESPACE for odh-model-controller deployment..."
+echo "   Setting MAAS_NAMESPACE for odh-model-controller deployment..."
 if kubectl get deployment odh-model-controller -n opendatahub &>/dev/null; then
     # Wait for deployment to be available before patching
     echo "   Waiting for odh-model-controller deployment to be ready..."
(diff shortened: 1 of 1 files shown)


liangwen12year:
We do not need the test case, as a manual dry-run or a simple CI check that runs the script with a custom namespace and verifies the output would be sufficient validation.

deployment/scripts/deploy-openshift.sh
```diff
@@ -309,7 +363,9 @@ kubectl wait --for=condition=Programmed gateway maas-default-gateway -n openshif
 echo ""
 echo "1️⃣1️⃣ Applying Gateway Policies..."
 cd "$PROJECT_ROOT"
-kustomize build deployment/base/policies | kubectl apply --server-side=true --force-conflicts -f -
+# Ensure MAAS_API_NAMESPACE is set with default (it should already be set above)
+export MAAS_API_NAMESPACE=${MAAS_API_NAMESPACE:-maas-api}
+kustomize build deployment/base/policies | envsubst '$MAAS_API_NAMESPACE' | kubectl apply --server-side=true --force-conflicts -f -
```
bartoszmajsak: Like here we really are super sure already :) Is there something messing up the value of the exported variable in the meantime?

Jaland: You never know when a bit flips half way thorugh a script :smile: 

This was an AI choice I should cleaned up but doing it now

deployment/scripts/deploy-openshift.sh
```diff
@@ -276,7 +316,17 @@ kubectl apply -f deployment/base/networking/odh/kuadrant.yaml
...
+TMP_DIR=$(mktemp -d)
+cp -r "$PROJECT_ROOT/deployment/base/maas-api"/* "$TMP_DIR/"
+# Replace hardcoded "namespace: maas-api" with "namespace: ${MAAS_API_NAMESPACE}" in kustomization.yaml
+sed -i "s|namespace: maas-api|namespace: \${MAAS_API_NAMESPACE}|g" "$TMP_DIR/kustomization.yaml"
+# Replace ${MAAS_API_NAMESPACE} placeholder with actual value
+envsubst '$MAAS_API_NAMESPACE' < "$TMP_DIR/kustomization.yaml" > "$TMP_DIR/kustomization.yaml.tmp"
+mv "$TMP_DIR/kustomization.yaml.tmp" "$TMP_DIR/kustomization.yaml"
+# Build and replace any remaining hardcoded namespace references in the output
+kustomize build "$TMP_DIR" | sed "s|namespace: maas-api|namespace: $MAAS_API_NAMESPACE|g" | kubectl apply -f -
+rm -rf "$TMP_DIR"
```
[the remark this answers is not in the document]
jland-redhat: Nah, I this is whole script is changing soon anyway. And I like the idea of future proofing against someone checking in a hard coded namespace.

israel-hdez (approved):
As a start looks OK.
I still have doubts about patching odh-model-controller. This should be done via the DSC once that part is ready.

:thinking: With the operator doing the deployment, several parts of this script may need to be rethinked.

Anyhoo:
/lgtm

PR #294, opened 2025-12-04
fix: model listing authorization

noyitz:
This PR addresses a security vulnerability where the /v1/models endpoint returned all models to all authenticated users regardless of their Model Auth policy permissions. The fix implements proper authorization filtering so users only see models they have access to based on their Kubernetes RBAC permissions.

Before: All authenticated users could see all models across all namespaces, creating a security issue where users could discover models they shouldn't have access to.

After: Users only see models in namespaces they have RBAC permissions for, following the principle of least privilege.

## Release Notes

* **New Features**
  * Token-based authorization for accessing and discovering available models
  * Automatic service account token exchange enabling seamless authorized access
  * Premium-tier simulator model with tier-based access control
  * Per-model authorization validation for security

* **Documentation**
  * Model discovery guide documenting token exchange and tiered access control
  * Deployment examples for standard and premium simulator variants

### DIFF
--- a/maas-api/internal/config/config.go
+++ b/maas-api/internal/config/config.go
@@ -62,11 +62,6 @@ type Config struct {
 	// DataPath is the path to the database file for disk mode.
 	// Default: /data/maas-api.db
 	DataPath string
-
-	// AuthCheckEndpoint is the endpoint path used for authorization checks.
-	// This will be appended to the base model URL for HEAD requests.
-	// Default: v1/models
-	AuthCheckEndpoint string
 }
 
 // Load loads configuration from environment variables.
@@ -75,16 +70,15 @@ func Load() *Config {
 	gatewayName := env.GetString("GATEWAY_NAME", constant.DefaultGatewayName)
 
 	c := &Config{
-		Name:              env.GetString("INSTANCE_NAME", gatewayName),
-		Namespace:         env.GetString("NAMESPACE", constant.DefaultNamespace),
-		GatewayName:       env.GetString("GATEWAY_NAME", gatewayName),
-		GatewayNamespace:  env.GetString("GATEWAY_NAMESPACE", constant.DefaultGatewayNamespace),
-		Port:              env.GetString("PORT", "8080"),
-		DebugMode:         debugMode,
-		StorageMode:       StorageModeInMemory,
-		DBConnectionURL:   env.GetString("DB_CONNECTION_URL", ""),
-		DataPath:          env.GetString("DATA_PATH", DefaultDataPath),
-		AuthCheckEndpoint: env.GetString("AUTH_CHECK_ENDPOINT", "/"),
+		Name:             env.GetString("INSTANCE_NAME", gatewayName),
+		Namespace:        env.GetString("NAMESPACE", constant.DefaultNamespace),
+		GatewayName:      env.GetString("GATEWAY_NAME", gatewayName),
+		GatewayNamespace: env.GetString("GATEWAY_NAMESPACE", constant.DefaultGatewayNamespace),
+		Port:             env.GetString("PORT", "8080"),
+		DebugMode:        debugMode,
+		StorageMode:      StorageModeInMemory,
+		DBConnectionURL:  env.GetString("DB_CONNECTION_URL", ""),
+		DataPath:         env.GetString("DATA_PATH", DefaultDataPath),
 	}
 
 	// Validate STORAGE_MODE env var through Set() to ensure consistent validation
@@ -109,5 +103,4 @@ func (c *Config) bindFlags(fs *flag.FlagSet) {
 	fs.Var(&c.StorageMode, "storage", "Storage mode: in-memory (default), disk, or external")
 	fs.StringVar(&c.DBConnectionURL, "db-connection-url", c.DBConnectionURL, "Database connection URL (required for --storage=external)")
 	fs.StringVar(&c.DataPath, "data-path", c.DataPath, "Path to database file (for --storage=disk)")
-	fs.StringVar(&c.AuthCheckEndpoint, "auth-check-endpoint", c.AuthCheckEndpoint, "Endpoint path for authorization checks (appended to model URL)")
 }
--- a/maas-api/internal/models/kserve_llmisvc.go
+++ b/maas-api/internal/models/kserve_llmisvc.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
-	"net/url"
 	"time"
 
 	kservev1alpha1 "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
@@ -69,12 +68,6 @@ func (m *Manager) userCanAccessModel(ctx context.Context, model Model, saToken s
 		return false
 	}
 
-	authCheckURL, err := url.JoinPath(model.URL.String(), m.authCheckEndpoint)
-	if err != nil {
-		m.logger.Debug("Failed to construct auth check URL", "modelID", model.ID, "error", err)
-		return false
-	}
-
 	backoff := wait.Backoff{
 		Steps:    4,
 		Duration: 100 * time.Millisecond,
@@ -84,7 +77,7 @@ func (m *Manager) userCanAccessModel(ctx context.Context, model Model, saToken s
 
 	lastResult := authDenied // fail-closed by default
 	if err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
-		lastResult = m.doAuthCheck(ctx, authCheckURL, saToken, model.ID)
+		lastResult = m.doAuthCheck(ctx, model.URL.String(), saToken, model.ID)
 		return lastResult != authRetry, nil
 	}); err != nil {
 		m.logger.Debug("Authorization check backoff failed", "modelID", model.ID, "error", err)
--- a/maas-api/internal/models/kserve_llmisvc_test.go
+++ b/maas-api/internal/models/kserve_llmisvc_test.go
@@ -258,7 +258,6 @@ func TestListAvailableLLMs_AlwaysAllowed(t *testing.T) {
 				fixtures.NewLLMInferenceServiceLister(fixtures.ToRuntimeObjects(tt.llmServices)...),
 				fixtures.NewHTTPRouteLister(fixtures.ToRuntimeObjects(tt.httpRoutes)...),
 				gateway,
-				"v1/models",
 			)
 			require.NoError(t, errMgr)
 
@@ -319,7 +318,6 @@ func TestListAvailableLLMs_Authorization(t *testing.T) {
 			fixtures.NewLLMInferenceServiceLister(fixtures.ToRuntimeObjects([]*kservev1alpha1.LLMInferenceService{llmService})...),
 			fixtures.NewHTTPRouteLister(),
 			gateway,
-			"v1/models",
 		)
 		require.NoError(t, errMgr)
 
@@ -337,7 +335,6 @@ func TestListAvailableLLMs_Authorization(t *testing.T) {
 			fixtures.NewLLMInferenceServiceLister(fixtures.ToRuntimeObjects([]*kservev1alpha1.LLMInferenceService{llmService})...),
 			fixtures.NewHTTPRouteLister(),
 			gateway,
-			"v1/models",
 		)
 		require.NoError(t, errMgr)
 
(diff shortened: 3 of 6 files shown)


maas-api/internal/models/kserve_llmisvc_test.go
```diff
@@ -293,6 +293,7 @@ func TestListAvailableLLMs(t *testing.T) { //nolint:maintidx // linter is compla
 				&kservefakev1beta1.FakeServingV1beta1{Fake: fakeKServe},
 				&kservefakev1alpha1.FakeServingV1alpha1{Fake: fakeKServe},
 				&gatewayfake.FakeGatewayV1{Fake: fakeGateway},
+				nil, // k8sClient not needed for this test
```
bartoszmajsak: I wonder if we want to provide some tests that actually demonstrate that the behaviour works as intended.

bartoszmajsak (commented):
I wonder if we can future-proof this verification, as we are currently locking it to the very specific implementation.

Related task: https://issues.redhat.com/browse/RHOAIENG-41343

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +38,56 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
...
+				Verb:      "get", // Check if user can access the model resource
+			},
+		},
+	}
+
+	result, err := m.k8sClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, review, metav1.CreateOptions{})
+	if err != nil {
+		log.Printf("Failed to check access for user %s to model %s/%s: %v", user.Username, namespace, modelName, err)
+		return false
+	}
```
pierDipi: don't we need to update the permissions for this call? https://github.com/opendatahub-io/maas-billing/blob/main/deployment/base/maas-api/rbac/clusterrole.yaml

```
- apiGroups:
  - "authorization.k8s.io"
  resources:
  - "subjectaccessreviews"
  verbs:
  - "create"
```

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +38,56 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
...
+				Group:     "serving.kserve.io",
+				Resource:  "llminferenceservices",
+				Name:      modelName,
+				Namespace: namespace,
+				Verb:      "get", // Check if user can access the model resource
+			},
+		},
+	}
+
+	result, err := m.k8sClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, review, metav1.CreateOptions{})
```
bartoszmajsak: We should consider resiliency here. This type of call can quickly become a bottleneck if not handled properly. It may be fine with a couple of models, but scaling to dozens (or more) will eventually push us into `kube-apiserver` rate limiting, and at that point we’ll start serving errors instead.

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +38,56 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
...
+	review := &authv1.SubjectAccessReview{
+		Spec: authv1.SubjectAccessReviewSpec{
+			User:   user.Username,
+			Groups: user.Groups,
+			ResourceAttributes: &authv1.ResourceAttributes{
+				Group:     "serving.kserve.io",
+				Resource:  "llminferenceservices",
+				Name:      modelName,
+				Namespace: namespace,
+				Verb:      "get", // Check if user can access the model resource
```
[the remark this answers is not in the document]
noyitz: done

maas-api/internal/models/kserve_llmisvc.go
noyitz: updated the ClusterRole permissions in deployment/base/maas-api/rbac/clusterrole.yaml to include the SubjectAccessReview permissions:

  # Subject access review for authorization
  - apiGroups: ["authorization.k8s.io"]
    resources: ["subjectaccessreviews"]
    verbs: ["create"]

maas-api/internal/models/kserve_llmisvc.go
bartoszmajsak: > plan to move to the caching in 299 once its merged.

I am not sure #299 will handle caching for `SubjectAccessReview` calls out of the box. Standard informers/caches don't apply here, meaning this code would result in expensive direct API calls (we still need direct API calls). That would mean also call for some other caching mechanism to make it robust.

**About the overall approach (and what I meant in [my general comment](https://github.com/opendatahub-io/maas-billing/pull/294#pullrequestreview-3544182446)):**
Instead of re-implementing our AuthPolicy logic here (and dealing with the caching complexity), could we test permission by looping back to the gateway directly? We can let the existing Auth(z) pipeline decide. We would just need to ensure we use a token with the [correct audience](https://github.com/opendatahub-io/maas-billing/blob/931ce3bc3e600f3a5c8e8b72d06b4c565ea492bf/deployment/base/policies/gateway-auth-policy.yaml#L30-L33) (likely the SA token rather than the user token).

jland-redhat (approved):

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -83,8 +213,9 @@ func llmInferenceServicesToModels(items []kservev1alpha1.LLMInferenceService) ([
 				OwnedBy: item.Namespace,
 				Created: item.CreationTimestamp.Unix(),
 			},
-			URL:   url,
-			Ready: checkLLMInferenceServiceReadiness(&item),
+			URL:          url,
+			Ready:        checkLLMInferenceServiceReadiness(&item),
+			ResourceName: item.Name, // Always use metadata.name for RBAC
```
bartoszmajsak: Is this still needed?

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +40,131 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
...
+	// This prevents test failures when using fake URLs that don't resolve
+	modelURLStr := model.URL.String()
+	if isTestURL(modelURLStr) {
+		log.Printf("Test URL detected for model %s, allowing access", model.ID)
+		return true
+	}
+
+	// Make a lightweight request to the model endpoint using the SA token
+	// The gateway's AuthPolicy will evaluate this request and determine if the user has access
+	healthURL := modelURLStr + "/health"
```
bartoszmajsak: Are we sure we are always applying the same auth rules for actual model endpoint?

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +40,131 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
...
+
+	req.Header.Set("Authorization", "Bearer "+saToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Printf("Failed to check access for user %s to model %s: %v", user.Username, model.ID, err)
+		return false
+	}
+	defer resp.Body.Close()
```
bartoszmajsak: I wonder if we can do a few retries here

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -35,6 +40,131 @@ func (m *Manager) ListAvailableLLMs(ctx context.Context) ([]Model, error) {
 	return llmInferenceServicesToModels(instanceLLMs)
 }
 
+// ListAvailableLLMsForUser lists LLMInferenceServices that the user has access to.
+// This method filters models based on user's permissions by delegating to the gateway's AuthPolicy.
+func (m *Manager) ListAvailableLLMsForUser(ctx context.Context, user *token.UserContext, saToken string) ([]Model, error) {
```
bartoszmajsak: `user` seems to be only used for logging, maybe we can let the caller decide what to log and remove this param here?

maas-api/internal/handlers/models.go
```diff
@@ -39,9 +40,37 @@ func (h *ModelsHandler) ListModels(c *gin.Context) {
...
+			}})
+		return
+	}
+
+	// Extract the SA token from the Authorization header
+	authHeader := c.GetHeader("Authorization")
+	saToken := ""
+	if authHeader != "" && len(authHeader) > 7 { // "Bearer " = 7 chars
+		saToken = authHeader[7:] // Remove "Bearer " prefix
+	}
```
bartoszmajsak: How this will work when I pass my user token which I think is generally the case for listing the model instead of SA token (or API key if we would be using those)?

Maybe we can also use `CutPrefix("Bearer")`  + `TrimSpace` to avoid magic numbers.

maas-api/internal/models/kserve_llmisvc.go
noyitz: Added retry logic with exponential backoff (3 attempts, 100ms/200ms/400ms delays)

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -34,6 +38,167 @@ func (m *Manager) ListAvailableLLMs() ([]Model, error) {
...
+			case <-ctx.Done():
+				return false
+			case <-time.After(retryDelays[attempt-1]):
+				// Continue with retry
+			}
+		}
+
+		// Create HTTP HEAD request for lightweight authorization check
+		// HEAD aligns with gateway policies while avoiding POST issues on inference endpoints
+		req, err := http.NewRequestWithContext(ctx, http.MethodHead, authCheckURL, nil)
```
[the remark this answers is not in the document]
jland-redhat: It is confusing but these are different types of permissions. Post was just a kubernetes permissions that was added to the vllm object in order to validate a user has access to it from a kubenetes standpoint.

But it does not have anything to do with http request. We are doing a HEAD request here because it is the most light weight method

maas-api/internal/config/config.go
```diff
@@ -70,15 +75,16 @@ func Load() *Config {
...
+		Name:              env.GetString("INSTANCE_NAME", gatewayName),
+		Namespace:         env.GetString("NAMESPACE", constant.DefaultNamespace),
+		GatewayName:       env.GetString("GATEWAY_NAME", gatewayName),
+		GatewayNamespace:  env.GetString("GATEWAY_NAMESPACE", constant.DefaultGatewayNamespace),
+		Port:              env.GetString("PORT", "8080"),
+		DebugMode:         debugMode,
+		StorageMode:       StorageModeInMemory,
+		DBConnectionURL:   env.GetString("DB_CONNECTION_URL", ""),
+		DataPath:          env.GetString("DATA_PATH", DefaultDataPath),
+		AuthCheckEndpoint: env.GetString("AUTH_CHECK_ENDPOINT", "/"),
```
ishitasequeira: The comment in the struct above seems to be defining default value of `AUTH_CHECK_ENDPOINT` to `v1/models`, but the code is defaulting it to `/`. Let's make it consistent?

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -31,7 +47,117 @@ func (m *Manager) ListAvailableLLMs() ([]Model, error) {
...
+		return authGranted
+
+	case resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden:
+		return authDenied
+
+	case resp.StatusCode == http.StatusNotFound:
+		// 404 is not a failure - endpoint might not exist but that's okay, allow access
+		// TODO: Follow-up with fail-close implementation https://issues.redhat.com/browse/RHOAIENG-45883
+		m.logger.Debug("Model endpoint returned 404, allowing access (not a denial)", "modelID", modelID)
+		return authGranted
```
ishitasequeira: I consider this a security concern. Should we start with returning authDenied instead of authGranted here itself?

maas-api/internal/config/config.go
ishitasequeira: Quick question: is the root path accessible or does it return anything? Right now if we only have `/v1/models` and `/completion`, `/v1/models` feels like the closest fit. For auth validation, though, it might be useful to have a dedicated endpoint that doesn’t return any data, perhaps something to consider for a future release.

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -31,7 +47,117 @@ func (m *Manager) ListAvailableLLMs() ([]Model, error) {
...
+		)
+		return authDenied
+
+	default:
+		// Retry on server errors (5xx) or other unexpected codes
+		m.logger.Debug("Unexpected status code, retrying",
+			"modelID", modelID,
+			"statusCode", resp.StatusCode,
+		)
+		return authRetry
```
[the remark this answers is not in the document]
ishitasequeira: You're right, I was just missing that part. The `lastResult := authDenied` default plus `return false` on backoff error covers this. LGTM 👍

PR #298, opened 2025-12-05
fix(rbac): adds missing HTTPRoute roles

bartoszmajsak:
As per title

## Summary by CodeRabbit

* **Chores**
  * Configured system access permissions to support Gateway API HTTP route management. The platform can now retrieve, list, and monitor routes, enabling enhanced traffic routing and filtering capabilities for improved network management and service visibility.


jland-redhat (approved):
How did this not cause an issue in the past?

Also did we recently add logic that requires we lookup this info?

PR #307, opened 2025-12-10
docs: document side effects when removing group membership during active usage

liangwen12year:
## Description
Document known issues and side effects that occur when removing a user from a group while they have active tokens or ongoing requests.

This addresses the Technical Preview requirement to identify and document undesired side effects that are tolerated but must be clearly documented for the community.

Side effects documented:
- Existing tokens remain valid until expiration
- Rate limiting continues at old tier until cache expires
- Service Account persists after group removal
- User downgrade creates new Service Account in lower tier
- Monitoring shows split metrics across tier namespaces

Includes recommended practices, workarounds, and related documentation links.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Added a new Known Issues page describing side effects and workarounds when removing group membership during active platform usage (token validity, rate limits, lingering service accounts, tier downgrade effects, and monitoring implications).
  * Updated Tier Configuration docs with a warning and best practices for safely removing group membership.
  * Added navigation entry so the new guidance appears in the Administrator Guide.

### DIFF
--- a/docs/content/configuration-and-management/group-membership-known-issues.md
+++ b/docs/content/configuration-and-management/group-membership-known-issues.md
@@ -94,10 +94,19 @@ When a user is removed from a group, their Service Account in the tier namespace
 
 **Workaround**:
 
-- Service Accounts can be manually cleaned up if needed:
+- Service Accounts can be manually cleaned up if needed.
+- The Service Account name is derived from the username: special characters are replaced with dashes, converted to lowercase, and an 8-character hash suffix is appended (e.g., `alice-example-com-a1b2c3d4` for `alice@example.com`).
+- To find the Service Account for a specific user, list and filter by the username pattern:
 
 ```bash
-kubectl delete serviceaccount <username-hash> -n maas-default-gateway-tier-<old-tier>
+# List all Service Accounts in the tier namespace
+kubectl get serviceaccount -n maas-default-gateway-tier-<old-tier>
+
+# Filter by username pattern (e.g., for user "alice@example.com")
+kubectl get serviceaccount -n maas-default-gateway-tier-<old-tier> | grep alice
+
+# Delete the identified Service Account
+kubectl delete serviceaccount <sa-name> -n maas-default-gateway-tier-<old-tier>
 ```
 
 #### 4. User Downgrade Creates New Service Account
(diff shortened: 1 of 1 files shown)


docs/content/configuration-and-management/group-membership-known-issues.md
```diff
@@ -0,0 +1,161 @@
...
+After:  Alice removed from "premium-users" (still in system:authenticated)
+        -> Old SA still exists in premium namespace
+        -> New token request creates SA in maas-default-gateway-tier-free
+        -> Alice now has SAs in both namespaces
+```
+
+**Workaround**:
+
+- Revoke tokens before changing group membership to ensure clean transition.
+- Periodically clean up orphaned Service Accounts in tier namespaces.
```
jland-redhat: This is a good suggestion but we don't have a great way of knowing what Service Accounts are orphaned.

Maybe change this to Deleting the service account manually when user changes group.

Lets also capture this in issues. This is something that I think we need to think through how we could handle in the operator or elsewhere.

docs/content/configuration-and-management/group-membership-known-issues.md
```diff
@@ -0,0 +1,161 @@
...
+- Old tokens continue to work until expiration.
+- This is a cleanup artifact, not a security issue (access is controlled by RBAC and rate limiting).
+
+**Workaround**:
+
+- Service Accounts can be manually cleaned up if needed:
+
+```bash
+kubectl delete serviceaccount <username-hash> -n maas-default-gateway-tier-<old-tier>
+```
```
bartoszmajsak: Hash might not be well-known, I wonder if we can add some labels to make this task easier

https://github.com/opendatahub-io/models-as-a-service/blob/39e62e98f382621e401211ae21cb9de0601b8266/maas-api/internal/token/labels.go#L13-L20

That would also open-up a possibility for deleting user tokens by admin

docs/content/configuration-and-management/group-membership-known-issues.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### 4. User Downgrade Creates New Service Account
+
+**Impact**: Low
+
+**Description**:
+
+When a user is moved to a lower tier (e.g., removed from `premium-users`, now only in `system:authenticated`):
```
bartoszmajsak: I think `system:authenticated` is our sample setup, not a hard rule. Maybe we can clarify this.

bartoszmajsak (commented):
These are very good finds!

@jland-redhat maybe worth capturing some things as tasks to improve UX

docs/content/configuration-and-management/group-membership-known-issues.md
liangwen12year: Agreed, that would be a nice enhancement. I've updated the docs to reflect the current approach for now. Will track this as a follow-up improvement.

PR #303, opened 2025-12-08
fix: copy of 302 pr just to be able to merge as we have blocking request there

dmytro-zaharnytskyi:
## Summary by CodeRabbit

* **Chores**
  * Updated build configuration settings to improve compilation behavior and system compatibility.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #308, opened 2025-12-10
chore(makefile): removes commented out section

bartoszmajsak:
Due to SQLite dependency we have to use `CGO_ENABLED=1` for our builds, otherwise the api_key endpoint is not functional.

The previously commented-out FIPS block no longer provides value and instead introduces confusion about when CGO should be enabled.

Leftover from #302 #303

## Summary by CodeRabbit

* **Chores**
  * Cleaned up build configuration by removing redundant code, with no impact to functionality or end-user experience.


jland-redhat (approved):
lgtm

PR #312, opened 2025-12-10
feat: auth flow update

jland-redhat:
Moved the authorization of which kube groups you are a part of to the AuthPolicy, and now we just get user context from the header. 

Note the upgrade to 1.3.1 was required as the way headers were handled has changed. In 1.3.0 they appended a new header for the envoy filter by default so I was able to override our header by just passing in one with the same key/value.

I have not validated this yet but one thing that should be nice about this is we can in theory hook up our AuthPolcies to other providers and it should work. We would just need to have the other providers group info in our config map.

## Summary by CodeRabbit

* **New Features**
  * Responses now include X-MaaS-Username and X-MaaS-Group headers.
  * Services accept header-based user identity (X-MaaS-Username, X-MaaS-Group) for request context.

* **Bug Fixes / Tests**
  * Tests converted to header-driven, table-driven suites validating header handling and error cases.

* **Chores**
  * Kuadrant operator references and installation docs updated to v1.3.1+.

### DIFF
--- a/deployment/base/maas-api/policies/auth-policy.yaml
+++ b/deployment/base/maas-api/policies/auth-policy.yaml
@@ -18,9 +18,9 @@ spec:
     response:
       success:
         headers:
-          X-Maas-Username:
+          X-MaaS-Username:
             plain:
               selector: auth.identity.user.username
-          X-Maas-Group:
+          X-MaaS-Group:
             plain:
               selector: auth.identity.user.groups.@tostr
--- a/maas-api/internal/token/handler.go
+++ b/maas-api/internal/token/handler.go
@@ -69,7 +69,8 @@ func (h *Handler) ExtractUserInfo() gin.HandlerFunc {
 			log.Printf("ERROR: Missing or empty %s header - auth policy configuration issue", constant.HeaderUsername)
 			c.JSON(http.StatusInternalServerError, gin.H{
 				"error":         "Exception thrown while generating token",
-				"exceptionCode": "ERR_7F3A",
+				"exceptionCode": "AUTH_FAILURE",
+				"refId":         "001",
 			})
 			c.Abort()
 			return
@@ -79,7 +80,8 @@ func (h *Handler) ExtractUserInfo() gin.HandlerFunc {
 			log.Printf("ERROR: Missing %s header for user: %s - auth policy configuration issue", constant.HeaderGroup, username)
 			c.JSON(http.StatusInternalServerError, gin.H{
 				"error":         "Exception thrown while generating token",
-				"exceptionCode": "ERR_B2C9",
+				"exceptionCode": "AUTH_FAILURE",
+				"refId":         "002",
 			})
 			c.Abort()
 			return
@@ -92,7 +94,8 @@ func (h *Handler) ExtractUserInfo() gin.HandlerFunc {
 			log.Printf("ERROR: Failed to parse %s header. Header value: %q, Error: %v - auth policy configuration issue", constant.HeaderGroup, groupHeader, err)
 			c.JSON(http.StatusInternalServerError, gin.H{
 				"error":         "Exception thrown while generating token",
-				"exceptionCode": "ERR_C4D1",
+				"exceptionCode": "AUTH_FAILURE",
+				"refId":         "003",
 			})
 			c.Abort()
 			return
--- a/maas-api/internal/token/handler_test.go
+++ b/maas-api/internal/token/handler_test.go
@@ -56,6 +56,7 @@ func TestAPIEndpoints(t *testing.T) {
 		expectedStatus  int
 		expectedError   string
 		expectedCode    string
+		expectedRefId   string
 		shouldHaveToken bool
 		description     string
 	}{
@@ -73,7 +74,8 @@ func TestAPIEndpoints(t *testing.T) {
 			group:          `["test-group"]`,
 			expectedStatus: http.StatusInternalServerError,
 			expectedError:  "Exception thrown while generating token",
-			expectedCode:   "ERR_7F3A",
+			expectedCode:   "AUTH_FAILURE",
+			expectedRefId:  "001",
 			description:    "Missing username header should return 500",
 		},
 		{
@@ -82,7 +84,8 @@ func TestAPIEndpoints(t *testing.T) {
 			group:          "",
 			expectedStatus: http.StatusInternalServerError,
 			expectedError:  "Exception thrown while generating token",
-			expectedCode:   "ERR_B2C9",
+			expectedCode:   "AUTH_FAILURE",
+			expectedRefId:  "002",
 			description:    "Missing group header should return 500",
 		},
 		{
@@ -91,7 +94,8 @@ func TestAPIEndpoints(t *testing.T) {
 			group:          "invalid-format",
 			expectedStatus: http.StatusInternalServerError,
 			expectedError:  "Exception thrown while generating token",
-			expectedCode:   "ERR_C4D1",
+			expectedCode:   "AUTH_FAILURE",
+			expectedRefId:  "003",
 			description:    "Invalid group header format should return 500",
 		},
 	}
@@ -147,6 +151,14 @@ func TestAPIEndpoints(t *testing.T) {
 						t.Errorf("expected exceptionCode to be a string, got %T", response["exceptionCode"])
 					} else if exceptionCode != tt.expectedCode {
 						t.Errorf("expected exceptionCode: '%s'; got: '%s'. Description: %s", tt.expectedCode, exceptionCode, tt.description)
+					}
+				}
+				if tt.expectedRefId != "" {
+					refId, ok := response["refId"].(string)
+					if !ok {
+						t.Errorf("expected refId to be a string, got %T", response["refId"])
+					} else if refId != tt.expectedRefId {
+						t.Errorf("expected refId: '%s'; got: '%s'. Description: %s", tt.expectedRefId, refId, tt.description)
 					}
 				}
 			}
--- a/maas-api/internal/constant/const.go
+++ b/maas-api/internal/constant/const.go
@@ -11,6 +11,6 @@ const (
 	DefaultResyncPeriod = 8 * time.Hour
 
 	// Header configuration constants.
-	HeaderUsername = "X-Maas-Username"
-	HeaderGroup    = "X-Maas-Group"
+	HeaderUsername = "X-MaaS-Username"
+	HeaderGroup    = "X-MaaS-Group"
 )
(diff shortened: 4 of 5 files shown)


maas-api/internal/token/handler.go
```diff
@@ -29,47 +31,91 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
+		// Create UserContext from headers
+		userContext := &UserContext{
+			Username:        username,
+			Groups:          groups,
+			IsAuthenticated: true, // Headers are set by auth policy, so user is authenticated
+			// UID and JTI are not available from headers, leave empty
+		}
+
+		log.Printf("DEBUG - Extracted user info from headers - Username: %s, Groups: %v, Source: %s",
+			username, groups, source)
```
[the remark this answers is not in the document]
jland-redhat: I plan on making this a debug log but that is a really good point

maas-api/internal/token/handler.go
```diff
@@ -29,40 +31,82 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
 			c.Abort()
 			return
 		}
 
+		// Create UserContext from headers
+		userContext := &UserContext{
+			Username:        username,
+			Groups:          groups,
+			IsAuthenticated: true, // Headers are set by auth policy, so user is authenticated
+			// UID and JTI are not available from headers, leave empty
```
bartoszmajsak: Maybe we can remove `UID` and `JTI` then?  What are we using tit for?

docs/samples/dashboards/README.md
```diff
@@ -91,7 +91,7 @@ To customize the dashboard:
 
 ### 📝 Notes
 
-- Dashboard is compatible with Kuadrant v1.3.0-rc2+
+- Dashboard is compatible with Kuadrant v1.3.1+
```
bartoszmajsak: Which dashboard do we mean here?

deployment/base/policies/rate-limit-policy.yaml
```diff
@@ -35,4 +35,4 @@ spec:
         - predicate: |
             auth.identity.tier == "enterprise"
       counters:
-        - expression: auth.identity.userid
+        - expression: auth.identity.userid
```
bartoszmajsak: Probably just revert?

deployment/base/maas-api/policies/auth-policy.yaml
```diff
@@ -15,3 +16,15 @@ spec:
...
             - maas-default-gateway-sa
+    response:
+      success:
+        headers:
+          X-MAAS-USERNAME:
+            plain:
+              selector: auth.identity.user.username
+          X-MAAS-GROUP:
+            plain:
+              selector: auth.identity.user.groups.@tostr
```
bartoszmajsak: I wonder if we can use the following headers:

- `X-Model-User`
- `X-Model-User-Groups`

Would it make them more self-explanatory? I am also on the fence between `X-MaaS` and `X-Model` (but maybe that's because of the time of the year? :))

maas-api/cmd/main.go
```diff
@@ -149,20 +149,14 @@ func registerHandlers(ctx context.Context, router *gin.Engine, cfg *config.Confi
...
-	reviewer := token.NewReviewer(clusterConfig.ClientSet, cfg.Name+"-sa")
-
 	// Model listing endpoint (v1Routes is grouped under /v1, so this creates /v1/models)
-	//nolint:contextcheck // Context is properly accessed via gin.Context in the returned handler
-	v1Routes.GET("/models", tokenHandler.ExtractUserInfo(reviewer), modelsHandler.ListLLMs)
+	v1Routes.GET("/models", tokenHandler.ExtractUserInfo(), modelsHandler.ListLLMs)
 
-	//nolint:contextcheck // Context is properly accessed via gin.Context in the returned handler
-	tokenRoutes := v1Routes.Group("/tokens", tokenHandler.ExtractUserInfo(reviewer))
+	tokenRoutes := v1Routes.Group("/tokens", tokenHandler.ExtractUserInfo())
```
bartoszmajsak: As `ExtractUserInfo` is currently used by two different endpoints, perhaps it should be promoted to its own package to avoid coupling? WDYT about moving it to top-level `middleware` package?

maas-api/internal/token/handler.go
```diff
@@ -29,40 +31,82 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
+			log.Printf("Missing X-MAAS-GROUP header for user: %s", username)
+			c.JSON(http.StatusUnauthorized, gin.H{"error": "X-MAAS-GROUP header required"})
 			c.Abort()
 			return
 		}
 
-		token := strings.TrimPrefix(authHeader, "Bearer ")
-
-		// Validate with Manager (K8s validation)
-		userContext, err := h.manager.ValidateToken(c.Request.Context(), token, reviewer)
```
bartoszmajsak: `ValidateToken` is dead code now, and also makes the whole `Reviewer` obsolete as this PR moves the logic of user extraction up to AuthPolicy level - I think we can remove both.

maas-api/internal/token/handler.go
```diff
@@ -29,40 +31,82 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
+}
+
+// ExtractUserInfo extracts user information from X-MAAS-* headers set by the auth policy.
+func (h *Handler) ExtractUserInfo() gin.HandlerFunc {
 	return func(c *gin.Context) {
-		authHeader := c.GetHeader("Authorization")
-		if authHeader == "" {
-			c.JSON(http.StatusUnauthorized, gin.H{"error": "Authorization header required"})
+		username := strings.TrimSpace(c.GetHeader("X-MAAS-USERNAME"))
+		groupHeader := c.GetHeader("X-MAAS-GROUP")
```
bartoszmajsak: Probably worth extracting to constants for consistency - we also use it in tests.

maas-api/internal/token/handler.go
```diff
@@ -29,40 +31,82 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
+			})
 			c.Abort()
 			return
 		}
 
+		// Create UserContext from headers
+		userContext := &UserContext{
+			Username:        username,
+			Groups:          groups,
+			IsAuthenticated: true, // Headers are set by auth policy, so user is authenticated
```
bartoszmajsak: It will always be `true` - probably not needed and we can remove it to simplify further.

maas-api/internal/token/handler_test.go
```diff
@@ -73,10 +68,88 @@ func TestAPIEndpoints(t *testing.T) {
...
+		req.Header.Set("X-Maas-Source", "kubernetes-local")
+
 		w := httptest.NewRecorder()
 		router.ServeHTTP(w, req)
 
 		assert.Equal(t, http.StatusCreated, w.Code)
 		mockManager.AssertExpectations(t)
 	})
+
+	t.Run("IssueToken - Missing Username Header", func(t *testing.T) {
```
bartoszmajsak: Following a little bit of DRY, maybe we can rework these tests to "data driven ones" like we have here https://github.com/opendatahub-io/maas-billing/blob/2a1a0bacc89a2cb8b0c2064a7f72d7773a2d755e/maas-api/internal/token/handler_duration_test.go#L29-L52

We would define the fixture (headers) and expectations, reusing repeated req-call-response interaction. Cursor should be fast with this :)

bartoszmajsak (commented):
Mostly minor comments.

docs/samples/dashboards/README.md
Jaland: Honestly not sure, I can probably remove this, the Kuadrant stuff was just a find/replace of all the 1.3.0

maas-api/internal/token/handler.go
```diff
@@ -29,40 +31,82 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
 
-		if !userContext.IsAuthenticated {
-			log.Printf("Token is not authenticated for user: %s", userContext.Username)
-			c.JSON(http.StatusUnauthorized, gin.H{"error": "not authenticated"})
+		// Parse groups from header - format: "[group1 group2 group3]"
+		groups, err := parseGroupsHeader(groupHeader)
+		if err != nil {
+			log.Printf("ERROR: Failed to parse X-MAAS-GROUP header. Header value: %q, Error: %v", groupHeader, err)
+			c.JSON(http.StatusBadRequest, gin.H{
+				"error": fmt.Sprintf("Invalid X-MAAS-GROUP header format: %v. Check auth policy configuration.", err),
```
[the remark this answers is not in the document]
bartoszmajsak: 500 is more defensive and stronger signal about infra issue. I think that's the best choice 👍🏻

docs/samples/dashboards/README.md
bartoszmajsak: I am sure that's global rename, and that's what reviews are for :) Maybe we just list Kuadrant to avoid confusing README

maas-api/internal/token/handler.go
jland-redhat: Done

deployment/base/maas-api/policies/auth-policy.yaml
```diff
@@ -15,3 +15,15 @@ spec:
...
```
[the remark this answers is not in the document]
jland-redhat: When testing we saw the exact opposite behavior. Where without this "[group1 group2]" was being passed in and when @tostr was passed in then we were getting "['group1', 'group2']" which was parsing correctly in the code

deployment/base/maas-api/policies/auth-policy.yaml
jland-redhat: So the thing is these are all internal anyways. So in theory no one sees them.

Also I would argue for MaaS here just because the header is not related to  the user of a specific model, but rather to the overall "MaaS User"

Sorry for the late address missed this on the front page convos

maas-api/cmd/main.go
jland-redhat: Yeah that is probably a good cleanup creating a JIRA for it here:
https://issues.redhat.com/browse/RHOAIENG-41999

maas-api/internal/token/handler.go
```diff
@@ -29,40 +32,81 @@ func NewHandler(name string, manager TokenManager) *Handler {
...
+		// Parse groups from header - format: "[group1 group2 group3]"
+		// Parsing errors also indicate configuration issues
+		groups, err := parseGroupsHeader(groupHeader)
 		if err != nil {
-			log.Printf("Token validation failed: %v", err)
-			c.JSON(http.StatusUnauthorized, gin.H{"error": "validation failed"})
+			log.Printf("ERROR: Failed to parse %s header. Header value: %q, Error: %v - auth policy configuration issue", constant.HeaderGroup, groupHeader, err)
+			c.JSON(http.StatusInternalServerError, gin.H{
+				"error":         "Exception thrown while generating token",
+				"exceptionCode": "ERR_C4D1",
```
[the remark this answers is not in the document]
jland-redhat: They were just sort of random that I made up. Just like the idea of having something the users can give support that we know where in the code something went wrong without having to dig into the logs (since often time we don't have access to those)

Did you have thoughts on a format

bartoszmajsak (approved):
/lgtm

PR #318, opened 2025-12-16
feat(models): expose LLMInferenceService annotations as model metadata

liangwen12year:
## Description
Add ModelDetails to the /v1/models endpoint response to surface model metadata from LLMInferenceService annotations:

- genaiUseCase: from opendatahub.io/genai-use-case annotation
- description: from openshift.io/description annotation
- displayName: from openshift.io/display-name annotation

This enables the Dashboard to display user-friendly model information (use case, description, display name) that was previously available with AAA Models but lost when MaaS became the only model source.

The modelDetails field is only included when at least one annotation is present, maintaining OpenAI API compatibility.

Ref: [RHOAIENG-41609](https://issues.redhat.com//browse/RHOAIENG-41609)

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Models now include optional enriched metadata (use case, description, display name) in API modelDetails when provided via annotations; absent/empty annotations omit modelDetails.

* **Tests**
  * Test fixtures can attach annotations to simulated models and specify per-scenario metadata assertions.
  * Added scenarios for full, partial, empty, and absent metadata; each runs baseline validations plus optional per-scenario checks of modelDetails.

### DIFF
--- a/maas-api/internal/handlers/models_test.go
+++ b/maas-api/internal/handlers/models_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"knative.dev/pkg/apis"
 
+	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/constant"
 	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/handlers"
 	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/logger"
 	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/models"
@@ -88,9 +89,9 @@ func TestListingModels(t *testing.T) {
 			GatewayName:      testGatewayName,
 			GatewayNamespace: testGatewayNamespace,
 			Annotations: map[string]string{
-				models.AnnotationGenAIUseCase: "General purpose LLM",
-				models.AnnotationDescription:  "A large language model for general AI tasks",
-				models.AnnotationDisplayName:  "Test Model Alpha",
+				constant.AnnotationGenAIUseCase: "General purpose LLM",
+				constant.AnnotationDescription:  "A large language model for general AI tasks",
+				constant.AnnotationDisplayName:  "Test Model Alpha",
 			},
 			AssertDetails: func(t *testing.T, model models.Model) {
 				t.Helper()
@@ -108,7 +109,7 @@ func TestListingModels(t *testing.T) {
 			GatewayName:      testGatewayName,
 			GatewayNamespace: testGatewayNamespace,
 			Annotations: map[string]string{
-				models.AnnotationDisplayName: "Test Model Beta",
+				constant.AnnotationDisplayName: "Test Model Beta",
 			},
 			AssertDetails: func(t *testing.T, model models.Model) {
 				t.Helper()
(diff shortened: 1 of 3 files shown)


maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -93,6 +94,35 @@ func findLLMInferenceServiceURL(llmIsvc *kservev1alpha1.LLMInferenceService) *ap
...
+	}
+
+	genaiUseCase := annotations[AnnotationGenAIUseCase]
+	description := annotations[AnnotationDescription]
+	displayName := annotations[AnnotationDisplayName]
+
+	// Only return ModelDetails if at least one field is populated
+	if genaiUseCase == "" && description == "" && displayName == "" {
+		return nil
+	}
```
bartoszmajsak: Is this check necessary?

maas-api/internal/handlers/models_test.go
```diff
@@ -164,6 +199,40 @@ func TestListingModels(t *testing.T) {
 			assert.Equal(t, tc.expectedModel.Ready, actualModel.Ready)
 		})
 	}
+
+	// Verify model with full metadata
+	t.Run("model-with-metadata has modelDetails", func(t *testing.T) {
```
bartoszmajsak: These tests are already executed in the main loop. 

We could extend `LlmTestScenario` to support per-scenario assertion hooks - functions defined on each test case that capture scenario-specific expectations as closures. We could conditionally call them in the main loop above - https://github.com/opendatahub-io/models-as-a-service/pull/318/changes#diff-017707c5a6d5f81952d731af8e6df1e0653a282bc74129aef05431b0104954fcR189-R201

maas-api/internal/models/kserve_llmisvc.go
```diff
@@ -93,6 +94,35 @@ func findLLMInferenceServiceURL(llmIsvc *kservev1alpha1.LLMInferenceService) *ap
 	return nil
 }
 
+// Annotation keys for model metadata.
+const (
```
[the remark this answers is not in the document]
liangwen12year: I moved it to `const.go` file, thanks!

liangwen12year:
> So this was not asked initially, but I am thinking would it make sense to just make this metadata piece environment variable driven?
> 
> I.E. read some like `LLM_INF_SVC_ANNT` that contains a list of the annotations we want to display and uses that instead of baking it into the code.
> 
> My thought is that if we decided we want to show more information later it will be easier to do this way. But it may be over engineering
> 
> Curios on your thoughts @bartoszmajsak

I think that it would be convenient to expand in the future and provide a lot of flexibility!

But on second thought,  I also worried that using an environment variable adds extra complexity here: the API response becomes less predictable for the Dashboard, we'd need extra code to parse and validate the config, and typos in annotation keys could cause hard-to-debug issues.  Since these three annotations are already agreed upon with the Dashboard team and won't change often, keeping them hardcoded is simpler, easier to test, and gives the Dashboard a reliable response format.

I honestly think that we can stick to the current method. If needed in the future, we can consider the environment var alternative method. Thanks!

jland-redhat (approved):

PR #319, opened 2025-12-17
fix: removed namespaces and cleaned the code

dmytro-zaharnytskyi:
Referencing these issues: 
https://issues.redhat.com/browse/RHOAIENG-41360
https://issues.redhat.com/browse/RHOAIENG-41359
https://issues.redhat.com/browse/RHOAIENG-41356

Tested on local cluster with this script scripts/verify-tokens-metadata-logic.sh and with go test

## Summary by CodeRabbit

* **Refactor**
  * API key flow simplified: keys and metadata are now scoped by username only (namespace removed); token handling unified to a concrete manager.

* **Chores**
  * Structured logger adopted across app and tests; storage schema, queries and tests updated to match username-only scope.
  * Logging API simplified (context-aware logger method removed).

* **Breaking Change**
  * Multiple public method signatures changed or removed — update any integrations and tests accordingly.

### DIFF
--- a/maas-api/test/fixtures/server_setup.go
+++ b/maas-api/test/fixtures/server_setup.go
@@ -222,6 +222,7 @@ func StubServiceAccountTokenCreation(clientset kubernetes.Interface) {
 		// Generate valid JWT
 		claims := jwt.MapClaims{
 			"jti": fmt.Sprintf("mock-jti-%d", time.Now().UnixNano()),
+			"iat": time.Now().Unix(),
 			"exp": time.Now().Add(time.Hour).Unix(),
 			"sub": "system:serviceaccount:test-namespace:test-sa",
 		}


maas-api/internal/api_keys/service.go
```diff
@@ -10,9 +10,7 @@ import (
 
 type TokenManager interface {
 	GenerateToken(ctx context.Context, user *token.UserContext, expiration time.Duration, name string) (*token.Token, error)
-	RevokeTokens(ctx context.Context, user *token.UserContext) (string, error)
-	// GetNamespaceForUser returns the namespace for a user based on their tier
-	GetNamespaceForUser(ctx context.Context, user *token.UserContext) (string, error)
+	RevokeTokens(ctx context.Context, user *token.UserContext) error
```
bartoszmajsak: I am still not sure why we also have this:

https://github.com/opendatahub-io/models-as-a-service/blob/39e62e98f382621e401211ae21cb9de0601b8266/maas-api/internal/token/handler.go#L19-L21

dmytro-zaharnytskyi: Created Jira for that - https://issues.redhat.com/browse/RHOAIENG-42341

maas-api/internal/token/manager.go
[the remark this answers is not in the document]
bartoszmajsak: Do we want to handle error or we assume `iat` is always set?

PR #338, opened 2026-01-05
fix(lint): resolve unused nolint directives for ireturn linter

bartoszmajsak:
## Description of the problem

The nolintlint linter was flagging unused `//nolint:ireturn` directives in test files. Removing explicit exclusion lead to exact opposite error reported - that functions return interfaces (`ireturn` errors)

Not sure what is the exact issue here, but I see this failing in seemingly unrelated PR (#227) where I think golangci-linter GitHub Action runs against a diff that include merge commits.

Tested with `act -j lint -W .github/workflows/maas-api-ci.yml`

## Proposed Fixes

Returns concrete type `*SQLStore` instead of `MetadataStore` interface in test helper where possible. For test fixtures that must return external interfaces (listers from client-go, kserve, gateway-api), uses `//nolint:nolintlint,ireturn` to handle both scenarios: when ireturn fires and when it doesn't.

Cleans up `.golangci.yml` by removing bloated ireturn allow list, keeping only built-in keywords. Edge cases are now handled locally with inline directives, making exceptions visible where they occur and avoiding config bloat.

## Summary by CodeRabbit

* **Tests**
  * Refactored test helpers and fixtures to improve test infrastructure reliability.

* **Chores**
  * Updated linting configuration and test directives to align with current code standards.

### DIFF
--- a/maas-api/.golangci.yml
+++ b/maas-api/.golangci.yml
@@ -64,17 +64,6 @@ linters:
         - empty
         - stdlib
         - generic
-        - EventHandler
-        - discovery.DiscoveryInterface
-        - dynamic.Interface
-        - predicate.Predicate
-        - client.Object
-        - client.Client
-        - types.AsyncAssertion
-        - kubernetes.Interface
-        - gomega.AsyncAssertion
-        - labels.Selector
-        - manager.Manager
     lll:
       line-length: 180
     nolintlint:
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
Spotted here https://github.com/opendatahub-io/models-as-a-service/actions/runs/20727503765/job/59507365355?pr=227#step:5:42

PR #343, opened 2026-01-09
fix(auth): prevent SA tokens from issuing new tokens

bartoszmajsak:
Closes a security loophole where MaaS-issued service account tokens could be used to create additional Service Accounts, enabling unlimited token proliferation.

Previously, both cluster identity tokens and MaaS-issued SA tokens were accepted for all maas-api endpoints, including token and API key creation. This allowed users to bootstrap unlimited tokens from a single initial token.

Token management endpoints now require Kubernetes identity tokens only. SA tokens remain valid for model inference, listing, and tier lookups, but cannot create or revoke tokens/API keys. This enforces a clean separation between the management plane (cluster identity) and the data plane (SA tokens for inference).

Reproducer: https://gist.github.com/bartoszmajsak/f14269598dd2fada7e3940b2268fe26a

Ref: [RHOAIENG-45886](https://issues.redhat.com/browse/RHOAIENG-45886)

### Improved Security Model

```
OpenShift Token (oc whoami -t)
  ├── /maas-api/v1/tokens/*     → Create, Delete tokens
  ├── /maas-api/v1/api-keys/*   → Create, List, View API keys
  └── /v1/models                → List models

SA Token / Keys (issued by MaaS)
  ├── /v1/models                → List models
  └── /llm/*                    → Model inference
```

❌ SA Token CANNOT:
- Create new tokens (prevents token-to-token)
- Revoke tokens
- Create/view API keys

## Release Notes

* **New Features**
  * Added a dedicated token issuance endpoint, enabling users to create and manage API tokens

* **Documentation**
  * Updated installation and validation guides with token endpoint setup instructions
  * Clarified authentication configuration requirements

* **Tests**
  * Enhanced authentication validation testing for token endpoints

### DIFF
--- a/test/e2e/tests/conftest.py
+++ b/test/e2e/tests/conftest.py
@@ -11,15 +11,35 @@ def maas_api_base_url() -> str:
 
 @pytest.fixture(scope="session")
 def token(maas_api_base_url: str) -> str:
-    # Expect smoke.sh to have minted TOKEN already, but allow fallback
+    """
+    MaaS token for API authentication.
+    
+    Token is typically provided via TOKEN env var (set by smoke.sh).
+    
+    Fallback minting via oc whoami -t works for:
+    - Real users (Kubernetes identity tokens)
+    - Service accounts from maas-ci-test namespace (CI bypass policy)
+    
+    Fallback will FAIL for other service accounts due to security policy
+    that prevents SA tokens from minting new tokens.
+    
+    See: deployment/base/maas-api/policies/auth-policy-token-issuance.yaml
+    See: test/e2e/scripts/ci-auth-policy-bypass.yaml (CI exception)
+    """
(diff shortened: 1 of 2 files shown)


deployment/base/maas-api/policies/auth-policy.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   rules:
     # Allow any authenticated user to access the API
     authentication:
-      openshift-identities:
+      default:
```
israel-hdez: Is this change to properly override in the new policy?

bartoszmajsak: No, everything is "default" if not in "override". It is just a label, to make ppl not freaking out that it is openshift-specific. We can think about better name.

PR #350, opened 2026-01-14
feat: update deploy script to work with new RHOAI Operator Flow

jland-redhat:
## Description

Updates the `deploy-rhoai-stable.sh` script and `validate-deployment.sh` to align with the new RHOAI/ODH operator-managed MaaS deployment model.

### Key Changes:

**deploy-rhoai-stable.sh:**
- Added support for both RHOAI and ODH operators via `OPERATOR_TYPE` environment variable (defaults to `rhoai`)
- Added `checkcsvexists()` function to check for operator installation using CSV prefix matching (e.g., `opendatahub-operator` matches `opendatahub-operator.v3.2.0`)
- Added mutual exclusivity checks to prevent installing both RHOAI and ODH on the same cluster
- Added `deploy_dscinitialization()` function to create DSCInitialization resource before DataScienceCluster
- Updated DataScienceCluster to include `modelsAsService` component managed by the operator
- The script now installs the Gateway and usage policies separately since core MaaS components (deployment, auth policy) are managed by the operator
- Added `CERT_NAME` environment variable for TLS certificate configuration (defaults to `data-science-gateway-service-tls`)
- Fixed `envsubst` by properly exporting `CERT_NAME` variable

**validate-deployment.sh:**
- Updated MaaS API pod checks to look in `opendatahub` namespace (where the operator deploys MaaS API) instead of `maas-api` namespace
- Updated HTTPRoute checks to look in `opendatahub` namespace
- Updated all error messages and suggestions to reference the correct namespace

**deployment/overlays/openshift/kustomization.yaml:**
- Removed reference to `../odh` overlay since operator now manages those components

## How Has This Been Tested?

- Deployed to a fresh OpenShift cluster using `./scripts/deploy-rhoai-stable.sh`
- Verified operator installation with CSV check
- Verified DSCInitialization and DataScienceCluster creation
- Verified Gateway and usage policies deployment
- Ran `./scripts/validate-deployment.sh` to confirm all checks pass
- Tested model inference through the MaaS Gateway
- Verified rate limiting and authentication are working

## Merge criteria:

## Note:

May need some alterations to the e2e test to make this work

## Summary by CodeRabbit

* **New Features**
  * Dual-operator deployment choice (RHOAI or OpenDataHub) with configurable catalog, image and channel
  * Models-as-a-Service enabled as a managed component

* **Chores**
  * Expanded deployment helpers, readiness checks and validation flow for more reliable installs
  * Updated CI smoke tests to exercise new flows, retries and diagnostics
  * Removed a default overlay resource to streamline overlays

### DIFF
--- a/scripts/deploy-rhoai-stable.sh
+++ b/scripts/deploy-rhoai-stable.sh
@@ -671,12 +671,37 @@ kubectl apply --server-side=true \
   -f <(kustomize build "https://github.com/opendatahub-io/models-as-a-service.git/deployment/base/policies/usage-policies?ref=${MAAS_REF}")
 
 # Fix audience for ROSA/non-standard clusters
+# =====================================================
+# Background:
+# - Hypershift/ROSA clusters use custom OIDC providers with non-standard audiences
+# - Default AuthPolicy expects audience: https://kubernetes.default.svc
+# - Service account tokens from these clusters have different audiences
+# - Without this patch, authentication fails with HTTP 401
+#
+# Problem:
+# - When modelsAsService.managementState: Managed, the operator creates AuthPolicy
+# - Operator's reconciliation loop may revert manual patches back to defaults
+# - This causes authentication to break after initial successful patching
+#
+# Solution:
+# - Annotate AuthPolicy with opendatahub.io/managed=false to prevent reconciliation
+# - Patch with cluster-specific audience
+# - Verify the patch persisted after giving operator time to reconcile
+# - Warn user if operator reverts the change
+# =====================================================
 if [[ -n "$AUD" && "$AUD" != "https://kubernetes.default.svc" ]]; then
   echo
   echo "## Configuring audience for non-standard cluster"
   echo "* Detected non-default audience: ${AUD}"
-  # Patch the correct AuthPolicy: maas-api-auth-policy in the openshift-ingress namespace
+
   if wait_for_resource "authpolicy" "maas-api-auth-policy" "$APPLICATIONS_NS" 300; then
+    # Step 1: Annotate to prevent operator reconciliation
+    # This tells the operator to not manage this resource, allowing our custom config to persist
+    kubectl annotate authpolicy maas-api-auth-policy -n "$APPLICATIONS_NS" \
+      opendatahub.io/managed="false" --overwrite 2>/dev/null || true
+
+    # Step 2: Patch AuthPolicy with cluster-specific audience
+    # The custom audience allows service account tokens from Hypershift/ROSA to be validated
     kubectl patch authpolicy maas-api-auth-policy -n "$APPLICATIONS_NS" --type=merge --patch-file <(echo "
 spec:
   rules:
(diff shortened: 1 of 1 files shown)


scripts/deploy-rhoai-stable.sh
```diff
@@ -283,56 +400,102 @@ spec:
...
+    for i in {1..60}; do
+      if kubectl get namespace opendatahub >/dev/null 2>&1; then
+        kubectl wait namespace/opendatahub --for=jsonpath='{.status.phase}'=Active --timeout=60s
+        return 0
+      fi
+      sleep 5
+    done
+    echo "  WARNING: opendatahub namespace was not created within timeout."
+  fi
+}
```
bartoszmajsak: Maybe we could move it to `deployment-helpers.sh` as parametrized function? Especially that we should wait for `redhat-ods-applications` if we use RHOAI

jland-redhat: I am starting to get the impression you want me to use deployment-helps.sh...

Agree that is a good idea let me offboard this logic there

scripts/deployment-helpers.sh
```diff
@@ -230,3 +343,86 @@ wait_for_csv() {
...
+    state=$(kubectl get catalogsource "$name" -n "$namespace" -o jsonpath='{.status.connectionState.lastObservedState}' 2>/dev/null || echo "")
+    
+    if [ "$state" = "READY" ]; then
+      echo "  * CatalogSource '$name' is ready"
+      return 0
+    fi
+
+    sleep $interval
+    elapsed=$((elapsed + interval))
+  done
```
ishitasequeira: Can we replace this with `kubectl wait` command? Also, looks like we are silently skipping the failure of CatalogSource becoming available? My suggestion is to update as something like below?

```suggestion
if ! kubectl wait catalogsource "$name" -n "$namespace" \
     --for=jsonpath='{.status.connectionState.lastObservedState}'=READY \
     --timeout="${timeout}s" 2>/dev/null; then
  local state=$(kubectl get catalogsource "$name" -n "$namespace" \
    -o jsonpath='{.status.connectionState.lastObservedState}' 2>/dev/null)
  echo "  WARNING: CatalogSource may not be fully ready yet (state: $state)"
  return 1  # ← Fix: should return error, not success
fi
echo "  * CatalogSource '$name' is ready"
return 0
```

test/e2e/scripts/prow_run_smoke_test.sh
```diff
@@ -88,16 +91,31 @@ check_prerequisites() {
 }
 
 deploy_maas_platform() {
-    echo "Deploying MaaS platform on OpenShift..."
-    if ! "$PROJECT_ROOT/scripts/deploy-openshift.sh"; then
+    echo "Deploying MaaS platform on OpenShift..." --operator-image quay.io/opendatahub/opendatahub-operator:pr-3063
```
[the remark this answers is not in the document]
liangwen12year: Other than the Nit above, the PR looks good to me!

deployment/overlays/openshift/kustomization.yaml
```diff
@@ -6,6 +6,5 @@ metadata:
 
 resources:
   - ../../base/networking/maas
-  - ../odh
```
bartoszmajsak: What's the reason for this change?

bartoszmajsak (changes requested):
Couple of observations, but the most important one is that it's not always `opendatahub` namespace we are dealing with. It seems to be hardcoded in a couple of places, but when we use RHOAI we should be targeting `red-hat-ods-application` namespace insteand.

### Default execution mode

Executing `./scripts/deploy-rhoai-stable.sh` without any parameters yields:

```
## Configuring DSCInitialization and DataScienceCluster
* A DSCInitialization already exists in the cluster. Skipping creation.
* Setting up DataScienceCluster with MaaS capability...
Error from server: failed to create typed patch object (/default-dsc; datasciencecluster.opendatahub.io/v2, Kind=DataScienceCluster): .spec.components.kserve.modelsAsService: field not declared in schema
```

Following the principle of least surprise we should change defaults to something more reasonable.

scripts/deploy-rhoai-stable.sh
```diff
@@ -269,9 +393,148 @@ spec:
...
+      sampleRatio: '0.1'
+      storage:
+        backend: pv
+        retention: 2160h
+    managementState: Managed
+  trustedCABundle:
+    customCABundle: ''
+    managementState: Managed
+EOF
+}
```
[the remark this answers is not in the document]
bartoszmajsak: Something to look at for consolidation?

PR #357, opened 2026-01-19
fix(config): defaults to HTTP mode to maintain backward compatibility

bartoszmajsak:
Currently running deployments that rely on `:latest` image might not have security-related flags nor settings in their deployment manifests. This breaks backward compatibility and can yield an error on startup:

```
{"level":"error","timestamp":"2026-01-19T08:12:53.474+0100","caller":"cmd/main.go:44","message":"Configuration
validation failed","error":"--secure requires either --tls-cert/--tls-key or --tls-self-signed"}
```

This change sets `SECURE` to `false` by default to ensure backward compatibility.

Existing kustomize manifests explicitly configure the protocol mode, so changing the default has no impact on production environments.

This preserves the original server behavior where HTTP was the default.

## Summary by CodeRabbit

* **Chores**
  * Default security mode changed to insecure (secure flag now defaults to false); help text updated to reflect the new default.

### DIFF
--- a/maas-api/internal/config/config.go
+++ b/maas-api/internal/config/config.go
@@ -80,7 +80,7 @@ type Config struct {
 func Load() *Config {
 	debugMode, _ := env.GetBool("DEBUG_MODE", false)
 	gatewayName := env.GetString("GATEWAY_NAME", constant.DefaultGatewayName)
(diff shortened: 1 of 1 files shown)


israel-hdez (approved):
/lgtm

PR #363, opened 2026-01-20
fix(ci): pipelinerun

MohammadiIram:
## Summary by CodeRabbit

* **Chores**
  * Updated internal build pipeline configurations and naming conventions.
  * Adjusted service account references for CI/CD infrastructure.

### DIFF
--- a/.tekton/odh-maas-api-pull-request.yaml
+++ b/.tekton/odh-maas-api-pull-request.yaml
@@ -13,9 +13,9 @@ metadata:
   creationTimestamp: null
   labels:
     appstudio.openshift.io/application: opendatahub-builds
-    appstudio.openshift.io/component: maas-api
+    appstudio.openshift.io/component: odh-maas-api-ci
     pipelines.appstudio.openshift.io/type: build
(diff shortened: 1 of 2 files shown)


.tekton/odh-maas-api-pull-request.yaml
```diff
@@ -44,7 +44,7 @@ spec:
     - name: pathInRepo
       value: pipeline/multi-arch-container-build.yaml
   taskRunTemplate:
-    serviceAccountName: build-pipeline-maas-api
+    serviceAccountName: build-pipeline-maas-api-poc
```
bartoszmajsak: Why do we need a `-poc` account?

MohammadiIram: We already have a service account for the POC, so we’re using the same one. This won’t affect how the build works.

bartoszmajsak: I see, it's just a bit confusing to see such prefix :) Maybe long term we can skip it to avoid confusion.

PR #344, opened 2026-01-12
feat(observability): add ServiceMonitors and monitoring configuration

tgitelman:
## Summary
Add core observability infrastructure for metrics collection and fix a timeout bug in deployment helpers.

### Observability (Commit 1)
- Add ServiceMonitors for Limitador, Authorino, and maas-api
- Add Istio Gateway Service and ServiceMonitor for Envoy metrics
- Add KServe LLM models ServiceMonitor for vLLM metrics
- Add cluster-monitoring-config to enable user-workload-monitoring
- Update base observability kustomization

### Bug Fix (Commit 2)
- Fix timeout handling in `wait_for_csv_with_min_version` function
- Now passes remaining time instead of full timeout to `wait_for_csv`
- Prevents potential 2x timeout duration

## Testing
All ServiceMonitors verified with Prometheus:
- Limitador: up=1 ✓
- Authorino: up=1 ✓
- KServe LLM: up=1 ✓
- Istio Gateway: up=1 ✓

## Summary by CodeRabbit

* **New Features**
  * Added Istio Gateway metrics Service and automatic ServiceMonitor for Prometheus scraping
  * Added ServiceMonitor sample for KServe LLM model workloads
  * Added Kustomize manifests to deploy observability monitors

* **Documentation**
  * Expanded observability guide with Grafana integration, dashboards, PromQL examples, and deploy instructions
  * Added sample cluster-monitoring ConfigMap and updated metric names: authorized_hits, authorized_calls, limited_calls

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #366, opened 2026-01-21
feat(observability): add per-user latency metrics

tgitelman:
Summary
This PR adds per-user latency tracking to Istio metrics and fixes a security vulnerability where the X-MaaS-Username header could be spoofed by clients.
Changes
Istio Telemetry resource (istio-telemetry.yaml) - adds a user label to REQUEST_DURATION metrics using correct Envoy attribute syntax (double quotes)
Security fix - Updated gateway-auth-policy to inject X-MaaS-Username header from authenticated identity, preventing client spoofing
Documentation update - Added per-user latency tracking section to observability docs with architecture flow and filtered query examples
Security Fix
Previously, the gateway-auth-policy (which protects model inference traffic) did NOT inject the X-MaaS-Username header - only maas-api-auth-policy did for /maas-api/* routes. This meant clients could spoof the header on model inference requests, enabling:
Unbounded cardinality attacks (millions of fake "users")
Attribution poisoning (requests attributed to other users)
The fix ensures the header is injected by Authorino from the trusted authenticated identity (auth.identity.userid), not from client input.
How it works
User authenticates via AuthPolicy
gateway-auth-policy injects X-MaaS-Username header from auth.identity.userid
Istio Telemetry reads the header and adds it as user label to REQUEST_DURATION metric
Prometheus scrapes metrics from the gateway
Example metric output
istio_request_duration_milliseconds_bucket{...,user="system:serviceaccount:maas-default-gateway-tier-free:kube-admin-81378af5",...} 2istio_request_duration_milliseconds_bucket{...,user="kube:admin",...} 3
Example query
# P99 latency per user (filter out unauthenticated requests)histogram_quantile(0.99, sum by (user, le) (rate(istio_request_duration_milliseconds_bucket{user!="",user!="unknown"}[5m])))
Testing
[x] Verified metrics contain user label with authenticated user identity
[x] Tested with oc whoami -t tokens → user="kube:admin"
[x] Tested with MaaS-issued tokens → user="system:serviceaccount:maas-default-gateway-tier-free:..."
[x] Verified unauthenticated/failed requests show user="unknown"
[x] Validated deployment via ./scripts/deploy-openshift.sh --insecure
Deployment
Automatically deployed when running ./scripts/deploy-openshift.sh as part of the observability stack.
Related
Complements existing Kuadrant TelemetryPolicy for Limitador metrics
Dashboard query updates for Grafana/Perses to be done in separate branches

### DIFF
--- a/scripts/validate-deployment.sh
+++ b/scripts/validate-deployment.sh
@@ -106,6 +106,10 @@ while [ $# -gt 0 ]; do
             shift 2
             ;;
         --namespace|-n)
+            if [ -z "$2" ] || [ "${2#-}" != "$2" ]; then
+                echo "Error: --namespace requires a value (e.g., --namespace maas-api)" >&2
+                exit 1
+            fi
             MAAS_API_NAMESPACE="$2"
             shift 2
             ;;


dmytro-zaharnytskyi:
I have re-run the test, but it continue failing. Error is not to informative, I made a small investigation and the only valid option that I have is that the fork name maas-billing doesn't match the upstream name models-as-a-service. So the solution would be to fork again and move this branch there OR reach out to someone who can help us with Konflux and re-configure it.

But it doesn't explain why we got https://github.com/opendatahub-io/models-as-a-service/pull/344 merged without issues. I think it could be some file change on the way.

But anyway, it is just an idea. maybe the reason is in something else

PR #368, opened 2026-01-21
fix(models): treat HTTP 405 as successful authorization in model access checks

bartoszmajsak:
The authorization check incorrectly treated HTTP `405 Method Not Allowed` as an authorization failure. This caused legitimate, authorized traffic to be denied when model servers didn't support the HTTP method used for probing.

A 405 response proves the request successfully traversed the gateway's AuthPolicies. If authorization had failed, the gateway would have returned 401, 403 or 404 before the backend could respond. The 405 simply indicates the HTTP method isn't enabled on that route, not an authorization denial.

"Variations" in CORS OPTIONS handling could affect vLLM deployments. Even though this is enabled by default, the specific behavior is an external decision that remains outside our control.

This PR also switches the probe endpoint from `OPTIONS /v1/chat/completions` to `GET (vllm)/v1/models` for future-proofing, as it's a standard OpenAI-compatible endpoint universally supported by inference servers.

> [!IMPORTANT]
> This is not the same call as `/v1/models` exposed by `maas-api`

Ref: [RHOAIENG-46770](https://issues.redhat.com/browse/RHOAIENG-46770)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced authorization verification mechanism for LLM services to more accurately interpret authorization responses across different scenarios
  * Updated endpoint and HTTP method parameters used for authorization checks to improve system reliability and compatibility
  * Refined error response handling to prevent false access denials in specific authorization scenarios
  * Improved overall consistency and robustness of the access control verification process

### DIFF
--- a/maas-api/internal/models/kserve_llmisvc.go
+++ b/maas-api/internal/models/kserve_llmisvc.go
@@ -76,7 +76,7 @@ func (m *Manager) userCanAccessModel(ctx context.Context, model Model, saToken s
 		Jitter:   0.1,
 	}
 
-	endpoint, errURL := url.JoinPath(model.URL.String(), "/v1/chat/completions")
+	endpoint, errURL := url.JoinPath(model.URL.String(), "/v1/models")
 	if errURL != nil {
 		m.logger.Error("Failed to create endpoint", "modelID", model.ID, "model.URL", model.URL.String(), "errURL", errURL)
 		return false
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
/retest

Hopefully infra hiccup, causing Kuadrant pods not being ready
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/opendatahub-io_models-as-a-service/368/pull-ci-opendatahub-io-models-as-a-service-main-e2e-test-maas-api/2014232935988727808#1:build-log.txt%3A205

PR #432, opened 2026-02-25
Maas subscription redesign fixes

jland-redhat:
## Summary

This PR adds the maas-controller layer to the kustomize overlays so it is picked up by the parent operator. It also removes the gateway-level auth policy (not used in the new subscription flow), and makes namespace and image configuration script-driven.

> [!NOTE]
> cert-manager and LWS operators are not installed by default in the kustomize flow. That was considered out of scope for this PR. For that flow, see: [install-cert-manager-and-lws.sh](https://gist.github.com/jland-redhat/b58968744b506e855b5c9db7bd490f46)

## Changes

- **Add maas-controller to kustomize overlays**: Include `maas-controller/config/default` in http-backend, tls-backend, and odh overlays so the parent operator deploys maas-controller with maas-api.
- **Remove gateway-auth-policy**: Delete `deployment/base/policies/auth-policies/` (gateway-auth-policy and its kustomization). This policy is not used in the new subscription flow; auth is handled per-route by maas-controller.
- **ODH overlay**: Replace auth-policies with maas-controller; ODH overlay now uses per-route auth/rate limits.
- **Remove install-maas-controller.sh**: maas-controller is installed via kustomize overlays.
- **Model samples**: Remove gateway-auth-policy references from simulator and simulator-premium model samples. Models cannot have that annotation until the webhook is removed from ODH-model-controller.
- **Namespace control in script**: Add `set_overlay_namespace` and `cleanup_overlay_namespace` so the deploy script sets the namespace before `kustomize build` instead of hardcoding it in overlays. maas-api and maas-controller deploy to `opendatahub` by default (matches operator mode).
- **`--maas-api-image` flag**: Add `--maas-api-image` to the deploy script for PR testing, consistent with `--operator-catalog` and `--operator-image`.
- **Kustomize overlay selection**: Use http-backend when TLS is disabled (instead of openshift) so maas-api is always deployed in kustomize mode.
- **Image handling**: Bake `MAAS_API_IMAGE` into the manifest before apply (via `set_maas_api_image`) to avoid server-side apply conflicts with post-apply patches.
- **Docs**: Update maas-controller README and related docs for the new flow.

## Testing

Performed a full install and confirmed that all components install and run correctly.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -500,30 +500,50 @@ deploy_via_kustomize() {
 #──────────────────────────────────────────────────────────────
 
 install_optional_operators() {
-  log_info "Installing optional operators (cert-manager and LWS in parallel)..."
+  log_info "Installing optional operators in parallel..."
 
   local data_dir="${SCRIPT_DIR}/data"
 
-  # Apply both subscriptions so OLM can process them in parallel
-  log_info "Applying cert-manager and LWS subscriptions..."
-  if ! kubectl apply -f "${data_dir}/cert-manager-subscription.yaml"; then
-    log_error "Failed to apply cert-manager subscription"
+  # Apply both subscriptions in parallel (they're independent)
+  log_info "Applying cert-manager and LeaderWorkerSet subscriptions..."
+  kubectl apply -f "${data_dir}/cert-manager-subscription.yaml" &
+  local cert_manager_pid=$!
+  kubectl apply -f "${data_dir}/lws-subscription.yaml" &
+  local lws_pid=$!
+
+  # Wait for both apply commands to complete and capture individual exit codes
+  local cert_manager_apply_rc=0
+  local lws_apply_rc=0
+  wait $cert_manager_pid || cert_manager_apply_rc=$?
+  wait $lws_pid || lws_apply_rc=$?
+
+  if [[ $cert_manager_apply_rc -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -363,32 +366,43 @@ main() {
...
+  log_info "MaaS Subscription Controller..."
+  local script_dir
+  script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  local project_root="$script_dir/.."
+  local controller_dir="$project_root/maas-controller"
+  local config_dir="$controller_dir/config/default"
+
+  if [[ ! -d "$controller_dir" ]]; then
+    log_warn "maas-controller directory not found at $controller_dir — skipping subscription controller"
+  else
```
ishitasequeira: Should we skip or error and exit? Does it make sense to go ahead if `subscription controller` is not available?

scripts/deploy.sh
```diff
@@ -302,7 +305,7 @@ validate_configuration() {
   if [[ "$DEPLOYMENT_MODE" == "kustomize" ]]; then
     # Kustomize mode: use provided namespace or default to maas-api
```
dmytro-zaharnytskyi: The comment here is outdated a bit :)) but I don't have anything else to comment here besides Ishita's comments, so that's it...

jland-redhat: Well it was factual pre this change which caused some issues ha ha.

But yeah should be updated

PR #455, opened 2026-02-27
fix: correcting audience/model endpoint to correct issues in test

jland-redhat:
## Merge criteria:

## Summary by CodeRabbit

* **Tests**
  * Subscription/auth tests now actively obtain cluster tokens (with fallback), log user/token info, remove skips, create service accounts when needed, poll for expected responses, assert correct status codes, and clean up.
* **Chores**
  * Deployment detects non-standard cluster audiences, patches the controller with a cluster-audience setting, centralizes handling, adds rollout checks, improved logging, and non-fatal warnings.
* **New Features**
  * Model discovery now surfaces backend-provided model IDs/names when access is granted.


test/e2e/tests/test_subscription.py
```diff
@@ -68,17 +72,41 @@ def _gateway_url():
...
+
 def _get_cluster_token():
-    # TODO: Consider always using SA for consistency; remove oc whoami -t fallback.
     sa_ns = os.environ.get("E2E_TEST_TOKEN_SA_NAMESPACE")
     sa_name = os.environ.get("E2E_TEST_TOKEN_SA_NAME")
-    return _create_sa_token(sa_name, namespace=sa_ns)
+    if sa_ns and sa_name:
+        token = _create_sa_token(sa_name, namespace=sa_ns)
+    else:
+        token_result = subprocess.run(["oc", "whoami", "-t"], capture_output=True, text=True)
```
dmytro-zaharnytskyi: I think it is worth adding timeout to these commands. As we have lots of issues with test, it could happen that once in a while it could just stuck. So either either timeout and rerun or just timeout and error in case of it.

But anyway, not the major issue right now, just a thought I caught myself on while looking through the code...

PR #488, opened 2026-03-06
fix: maas-controller performs duplicate recons

mynhardtburger:
<!--- Provide a general summary of your changes in the Title above -->

## Description
Fix for https://issues.redhat.com/browse/RHOAIENG-51607

This PR addresses three distinct performance/correctness issues across all three controllers, plus adds test infrastructure to prevent regressions. 

### 1. LastTransitionTime updated on every reconcile                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                
Problem: All three controllers had hand-rolled condition logic that set LastTransitionTime: metav1.Now() on every reconcile, regardless of whether the condition actually changed. This made the timestamp meaningless for debugging and caused every status write to produce a "change" — defeating any future diffing.                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                  
Fix: Replaced the hand-rolled logic with apimeta.SetStatusCondition (from k8s.io/apimachinery/pkg/api/meta), which only updates LastTransitionTime when the condition's Status field transitions (e.g., True → False).

Files: maasmodelref_controller.go, maassubscription_controller.go, maasauthpolicy_controller.go — the updateStatus / updateStatusWithReason methods.

### 2. Redundant writes to generated Kuadrant resources (O(N²) amplification)

Problem: When N subscriptions (or auth policies) reference the same model, each reconciliation aggregates all N into a single TokenRateLimitPolicy (or AuthPolicy) and unconditionally calls Update. Since the aggregated content is identical regardless of which subscription  triggered the reconcile, this produces N redundant writes. Each write triggers a watch event on the generated resource, which maps back to a subscription for re-reconciliation — cascading into O(N²) total reconciliations.

Fix: equality.Semantic.DeepEqual comparison done to skip the update if nothing changed.

Files: maassubscription_controller.go (TokenRateLimitPolicy update path), maasauthpolicy_controller.go (AuthPolicy update path).

 ### 3. Self-triggered re-reconciliation from status writes

Problem: All three controllers unconditionally call Status().Update() at the end of every reconcile. Status is a subresource that doesn't increment .metadata.generation, but without filtering, the For() watch treats it as a change and enqueues another reconciliation. This creates an unnecessary feedback loop on every reconcile.

Fix (two layers):
  - GenerationChangedPredicate on the For() watch of all three controllers — filters out status-only update events before they reach the work queue, preventing the reconcile from being triggered at all.
  - equality.Semantic.DeepEqual comparison in MaaSModelRef's updateStatusWithReason — skips the Status().Update() call if nothing changed. This is defense-in-depth for reconciliations triggered by other watch sources  (HTTPRoute, LLMInferenceService).

Files: maasmodelref_controller.go, maassubscription_controller.go, maasauthpolicy_controller.go — SetupWithManager methods and updateStatusWithReason.

## How Has This Been Tested?
<!--- Please describe in detail how you tested your changes. -->
<!--- Include details of your testing environment, and the tests you ran to -->
<!--- see how your change affects other areas of the code, etc. -->

## Merge criteria:
<!--- This PR will be merged by any repository approver when it meets all the points in the checklist -->
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Skip redundant resource and status writes by detecting no-op changes, reducing unnecessary updates and improving controller efficiency.
  * Reconciliations now trigger primarily on meaningful generation, deletion, or readiness transitions; logging better distinguishes unchanged vs updated resources.

* **Tests / Refactor**
  * Added duplicate-reconciliation tests to verify idempotency across controllers.
  * Tests refactored to use shared helper constructors and updated scaffolding to align with namespace/scoping changes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
+++ b/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
@@ -586,7 +586,10 @@ func (r *MaaSAuthPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	generatedAuthPolicy.SetGroupVersionKind(schema.GroupVersionKind{Group: "kuadrant.io", Version: "v1", Kind: "AuthPolicy"})
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&maasv1alpha1.MaaSAuthPolicy{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&maasv1alpha1.MaaSAuthPolicy{}, builder.WithPredicates(predicate.Or(
+			predicate.GenerationChangedPredicate{},
+			predicate.Funcs{UpdateFunc: deletionTimestampSet},
+		))).
 		// Watch HTTPRoutes so we re-reconcile when KServe creates/updates a route
 		// (fixes race condition where MaaSAuthPolicy is created before HTTPRoute exists).
 		Watches(&gatewayapiv1.HTTPRoute{}, handler.EnqueueRequestsFromMapFunc(
--- a/maas-controller/pkg/controller/maas/helpers.go
+++ b/maas-controller/pkg/controller/maas/helpers.go
@@ -7,7 +7,17 @@ import (
 
 	maasv1alpha1 "github.com/opendatahub-io/models-as-a-service/maas-controller/api/maas/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 )
+
+// deletionTimestampSet returns true when an object's DeletionTimestamp transitions
+// from nil to non-nil, indicating the object is being deleted. Use with
+// predicate.Funcs{UpdateFunc: deletionTimestampSet} alongside
+// GenerationChangedPredicate so that finalizer-based deletion handlers run.
+func deletionTimestampSet(e event.UpdateEvent) bool {
+	return e.ObjectOld.GetDeletionTimestamp().IsZero() &&
+		!e.ObjectNew.GetDeletionTimestamp().IsZero()
+}
 
 // validateCELValue checks that a string is safe to interpolate into a CEL expression.
 // Rejects values containing characters that could break or inject into CEL string literals.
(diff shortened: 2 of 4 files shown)


maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -539,24 +557,21 @@ func getAuthPolicyConditionState(ap *unstructured.Unstructured) (accepted, enfor
 
 func (r *MaaSAuthPolicyReconciler) updateStatus(ctx context.Context, policy *maasv1alpha1.MaaSAuthPolicy, phase, message string) {
 	policy.Status.Phase = phase
-	condition := metav1.Condition{
-		Type: "Ready", Status: metav1.ConditionTrue, Reason: "Reconciled", Message: message, LastTransitionTime: metav1.Now(),
```
mynhardtburger: Updating `LastTransitionTime` with each status update call is incorrect. `LastTransitionTime` must only be updated if the condition transitioned from one status to another. Status update logic for all 3 controllers were updated to make use of the standard `apimeta.SetStatusCondition` which only performs an update when the condition changed.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -569,7 +584,7 @@ func (r *MaaSAuthPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	generatedAuthPolicy.SetGroupVersionKind(schema.GroupVersionKind{Group: "kuadrant.io", Version: "v1", Kind: "AuthPolicy"})
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&maasv1alpha1.MaaSAuthPolicy{}).
+		For(&maasv1alpha1.MaaSAuthPolicy{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
mynhardtburger: Add predicate to filter out events which don't represent updates to the `spec`. This avoids self triggering a reconciliation when the previous run updated the status of the `MaasAuthPolicy`.

PR #584, opened 2026-03-23
feat: remove reliance on subscription header in inference and models endpoint

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-54596

### API key minting
- A column to API key storage and an optional field to API key creation request for subscription are added.
- If the field is omitted, the subscription with the highest priority is selected (with limit and name being the tie breaker).
### AuthPolicy flow
- Subscription name is now grabbed from `apiKeyValidation` result instead of the header `X-MaaS-Subscription`.
- `subscription-info` is still needed for validation check, as well as the header. It is injected after auth success.
### Controller feedback
- MaaSSubscriptions have a status condition `SpecPriorityDuplicate` to indicate if they share the same priority with other peers.

## How Has This Been Tested?

- Created API keys with different subscriptions, and see them in the database.
- Verified the auto-select rule.
- Verified that each API key can only access the model ref in associated subscription.
- Verified the status condition.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * API keys are now bound to a specific subscription at creation time. Optionally specify a `subscription` parameter when creating keys; if omitted, the highest-priority accessible subscription is automatically selected.
  * Subscriptions can now include a `priority` field to control default selection when multiple are available.
  * The `X-MaaS-Subscription` header is no longer required for API key-based inference requests; subscription is determined by the key's binding.

* **Documentation**
  * Updated guides to reflect subscription binding at key creation and removed header-requirement documentation.

### DIFF
--- a/test/e2e/tests/test_models_endpoint.py
+++ b/test/e2e/tests/test_models_endpoint.py
@@ -661,11 +661,11 @@ class TestModelsEndpoint:
                 check=True,
             )
 
-            # Create API key
+            # Create API key bound to our test subscription
             api_key_response = requests.post(
                 f"{_maas_api_url()}/v1/api-keys",
                 headers={"Authorization": f"Bearer {sa_token}", "Content-Type": "application/json"},
-                json={"name": "e2e-dedup-test-key"},
+                json={"name": "e2e-dedup-test-key", "subscription": subscription_name},
                 timeout=TIMEOUT,
                 verify=TLS_VERIFY,
             )
@@ -826,11 +826,11 @@ class TestModelsEndpoint:
                 check=True,
             )
 
-            # Create API key
+            # Create API key bound to our test subscription
             api_key_response = requests.post(
                 f"{_maas_api_url()}/v1/api-keys",
                 headers={"Authorization": f"Bearer {sa_token}", "Content-Type": "application/json"},
-                json={"name": "e2e-diff-refs-test-key"},
+                json={"name": "e2e-diff-refs-test-key", "subscription": subscription_name},
                 timeout=TIMEOUT,
                 verify=TLS_VERIFY,
             )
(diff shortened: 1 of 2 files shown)


maas-api/db/schema/0003_add_subscription_column.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Schema for API Key Management: 0003_add_subscription_column.up.sql
+-- Description: Add subscription column — binds each API key to a MaaSSubscription name at mint time
+
+-- Add subscription column (idempotent). Value is MaaSSubscription metadata.name, resolved when the key is created.
+ALTER TABLE api_keys ADD COLUMN IF NOT EXISTS subscription TEXT NOT NULL DEFAULT '';
```
ishitasequeira: Can we set the default as `null` instead?

maas-controller/pkg/controller/maas/maassubscription_controller.go
```diff
@@ -345,6 +360,108 @@ func (r *MaaSSubscriptionReconciler) updateStatus(ctx context.Context, subscript
 	}
 }
 
+// scanForDuplicatePriority lists live MaaSSubscriptions and sets SpecPriorityDuplicate
+// on each. Triggered on create, delete, or when spec.priority changes (see SetupWithManager).
+func (r *MaaSSubscriptionReconciler) scanForDuplicatePriority(ctx context.Context) {
```
ishitasequeira: Since `spec.priority` defaults to `0` via kubebuilder, any subscriptions that don't explicitly set priority will all be flagged as duplicates of each other. This might cause noise for operators who haven't configured priorities yet. Consider either:
- Documenting this behavior prominently
- Or treating `0` as "unset" and excluding from duplicate detection

maas-api/internal/handlers/models_test.go
```diff
@@ -500,12 +501,13 @@ func TestListingModelsWithSubscriptionHeader(t *testing.T) {
 		err = json.Unmarshal(w.Body.Bytes(), &response)
 		require.NoError(t, err, "Failed to unmarshal response body")
 
-		// User only has access to "free" subscription, so it auto-selects and returns free model
-		require.Len(t, response.Data, 1, "Expected one model with auto-selected free subscription")
+		// User token (no X-MaaS-Subscription header) returns all accessible models
```
jland-redhat: Reference to X-MaaS-Subscription header

PR #616, opened 2026-03-25
fix: remove reference to removed apiKeyValidation.subscription field

### LINKED ISSUE #615: bug: auth policy references removed apiKeyValidation.subscription field
## Bug

PR #584 removed the `subscription` field from the maas-api `/internal/v1/api-keys/validate` response, but the MaaS controller's auth policy generator still references `auth.metadata.apiKeyValidation.subscription` in three places:

1. `subscription-info` metadata body expression (`requestedSubscription` field)
2. `subscription-info` cache key selector
3. `X-MaaS-Subscription` response header expression

This causes Authorino's CEL evaluation to fail with `"no such key: subscription"`, which means the `subscription-info` metadata fetch fails, the `subscription-valid` OPA rule denies the request, and **all inference requests get 403**.

## Impact

All inference requests return 403 after deploying with PR #584's changes.

## Fix

Replace `auth.metadata.apiKeyValidation.subscription` with the request header fallback (`"x-maas-subscription" in request.headers ? request.headers["x-maas-subscription"] : ""`) in all three locations in `maasauthpolicy_controller.go`.

noyitz:
## Summary

PR #584 removed the `subscription` field from the maas-api `/internal/v1/api-keys/validate` response, but the auth policy controller still references `auth.metadata.apiKeyValidation.subscription` in CEL expressions. This breaks all inference requests with 403.

## Changes

Replaced `auth.metadata.apiKeyValidation.subscription` with the request header fallback in 3 places in `maasauthpolicy_controller.go`:

1. `subscription-info` metadata body expression
2. `subscription-info` cache key selector  
3. `X-MaaS-Subscription` response header expression

## Test plan

- [x] Verified on RHOAI cluster: all 5 models return 200 after the fix
- [x] Invalid API key returns 401
- [x] Auth policies regenerated correctly

Fixes #615

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription selection now consistently uses the incoming subscription request header instead of previously used validation data.
  * Subscription cache keys are derived from the request header to ensure consistent lookup behavior.
  * Response headers now reflect the resolved subscription selection metadata rather than prior validation-specific values.

### DIFF
--- a/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
+++ b/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
@@ -215,7 +215,7 @@ func (r *MaaSAuthPolicyReconciler) reconcileModelAuthPolicies(ctx context.Contex
 							"expression": fmt.Sprintf(`{
   "groups": auth.metadata.apiKeyValidation.valid == true ? auth.metadata.apiKeyValidation.groups : auth.identity.user.groups,
   "username": auth.metadata.apiKeyValidation.valid == true ? auth.metadata.apiKeyValidation.username : auth.identity.user.username,
-  "requestedSubscription": auth.metadata.apiKeyValidation.valid == true ? auth.metadata.apiKeyValidation.subscription : ("x-maas-subscription" in request.headers ? request.headers["x-maas-subscription"] : ""),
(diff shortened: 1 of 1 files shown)


jland-redhat:
@ryancham715 / @jrhyness please take a look at this and make sure there is not something lost in translation here.

PR #692, opened 2026-04-07
feat: [RHOAIRFE-496] add publicPaths to MaaSAuthPolicy for unauthenticated access to docs endpoints

Gajanan1995:
## Summary

Adds a `publicPaths` field to `MaaSAuthPolicySpec` that allows administrators to specify URL path suffixes (e.g., `/docs`, `/openapi.json`) that should be accessible **without authentication**. This enables users to browse model Swagger/OpenAPI documentation directly in a browser without needing a token, while keeping inference and all other endpoints fully authenticated.

**Motivation:** vLLM and other serving runtimes expose a `/docs` (Swagger UI) and `/openapi.json` endpoint, but the current MaaS authentication setup blocks unauthenticated access to all paths. There is no way to selectively expose documentation endpoints without bypassing auth entirely.

## Changes

- **CRD (`maasauthpolicy_types.go`):** Add `PublicPaths []string` field to `MaaSAuthPolicySpec` with `+optional`, max 10 items validation
- **Deepcopy (`zz_generated.deepcopy.go`):** Updated to handle the new `PublicPaths` slice
- **CRD manifest (`maas.opendatahub.io_maasauthpolicies.yaml`):** Regenerated via `controller-gen`
- **Reconciler (`maasauthpolicy_controller.go`):**
  - Added `buildPublicPathsRegex()` — generates a `request.url_path` regex pattern for Authorino
  - Added `buildPublicPathsCELPredicate()` — generates CEL `when` predicates to skip auth/metadata for public paths
  - Injects an `anonymous-public-paths` authentication rule into the generated AuthPolicy, scoped to matching paths
  - Adds `when` predicates to `apiKeyValidation`, `subscription-info` metadata evaluators and `auth-valid`, `subscription-valid`, `require-group-membership` authorization rules to skip them for public paths
- **Gateway default-deny TRLP (`gateway-default-deny.yaml`):** Updated predicate to exclude `/docs` and `/openapi.json` from the zero-rate-limit deny rule
- **Unit tests (`maasauthpolicy_controller_test.go`):** Added comprehensive tests for helper functions and reconciler behavior with and without `publicPaths`

## Usage

```yaml
apiVersion: maas.opendatahub.io/v1alpha1
kind: MaaSAuthPolicy
metadata:
  name: my-model-policy
spec:
  modelRefs:
    - name: my-model
      namespace: my-namespace
  subjects:
    groups:
      - system:authenticated
  publicPaths:
    - "/docs"
    - "/openapi.json"
```

## What changes, what stays the same

- `/docs` and `/openapi.json` → accessible without auth (anonymous)
- `/v1/completions`, `/v1/chat/completions`, etc. → still require full authentication (API key or token)
- Models without `publicPaths` → no behavior change at all (backward compatible)

## Test plan

- [x] Unit tests for `buildPublicPathsRegex` and `buildPublicPathsCELPredicate` helpers
- [x] Unit test: reconciler generates `anonymous-public-paths` auth rule when `publicPaths` is set
- [x] Unit test: reconciler adds `when` predicates to metadata/authorization evaluators
- [x] Unit test: reconciler produces unchanged AuthPolicy when `publicPaths` is empty
- [x] `make -C maas-controller test` — all tests pass
- [x] `make -C maas-controller verify-codegen` — generated files in sync
- [x] `./scripts/ci/validate-manifests.sh` — all kustomize manifests validate
- [ ] E2E: deploy with custom controller image, apply MaaSAuthPolicy with `publicPaths`, verify `/docs` returns 200 without auth and inference endpoints still require auth

## Release Notes

* **New Features**
  * Added optional public paths configuration for authentication policies. Specified paths now bypass authentication requirements while maintaining security on inference endpoints. Supports up to 10 public paths per policy.

### DIFF
--- a/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
+++ b/maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -341,6 +342,60 @@ func subscriptionGatewayCacheKeySelector() string {
 	)
 }
 
+// applyPublicPathsCELToRules appends a CEL "when" predicate to every evaluator in the
+// given rule map so that public-path requests skip those evaluators entirely.
+func applyPublicPathsCELToRules(rules map[string]any, celPredicate string) {
+	for name, evaluator := range rules {
+		eval, ok := evaluator.(map[string]any)
+		if !ok {
+			continue
+		}
+		when, _ := eval["when"].([]any)
+		eval["when"] = append(when, map[string]any{
+			"predicate": celPredicate,
+		})
+		rules[name] = eval
+	}
+}
+
+// buildPublicPathsRegex constructs a regex that matches exact public-path URLs across all
+// models in the allowlist. Each model's route prefix is /{namespace}/{name} per the
+// HTTPRoute convention, so /docs on model default/llm becomes ^/default/llm/docs$.
+func buildPublicPathsRegex(allowlists map[string]modelSubjectAllowlist) string {
+	var parts []string
+	for modelKey, entry := range allowlists {
+		if len(entry.PublicPaths) == 0 {
+			continue
+		}
+		for _, p := range entry.PublicPaths {
+			full := "/" + modelKey + p
+			parts = append(parts, fmt.Sprintf("^%s$", regexp.QuoteMeta(full)))
+		}
+	}
+	sort.Strings(parts)
+	return strings.Join(parts, "|")
+}
+
+// buildPublicPathsCELPredicate constructs a CEL predicate that evaluates to true for
+// requests that are NOT targeting public paths. Attach as a "when" predicate on metadata
+// and authorization evaluators so they are skipped for unauthenticated doc endpoints.
+func buildPublicPathsCELPredicate(allowlists map[string]modelSubjectAllowlist) string {
+	var parts []string
+	for modelKey, entry := range allowlists {
+		if len(entry.PublicPaths) == 0 {
+			continue
+		}
+		for _, p := range entry.PublicPaths {
+			full := "/" + modelKey + p
+			escaped := strings.ReplaceAll(full, `\`, `\\`)
+			escaped = strings.ReplaceAll(escaped, `"`, `\"`)
+			parts = append(parts, fmt.Sprintf(`request.path != "%s"`, escaped))
+		}
+	}
+	sort.Strings(parts)
+	return strings.Join(parts, " && ")
+}
+
 //+kubebuilder:rbac:groups=maas.opendatahub.io,resources=maasauthpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=maas.opendatahub.io,resources=maasauthpolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=maas.opendatahub.io,resources=maasauthpolicies/finalizers,verbs=update
@@ -447,7 +502,7 @@ func (r *MaaSAuthPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	if err := r.reconcileGatewayAuthPolicy(ctx, log, string(modelAllowlistsJSON), oidc, tenantID, gatewayNs, gatewayName); err != nil {
+	if err := r.reconcileGatewayAuthPolicy(ctx, log, string(modelAllowlistsJSON), oidc, tenantID, gatewayNs, gatewayName, modelAllowlists); err != nil {
 		log.Error(err, "failed to reconcile gateway AuthPolicy")
 		r.updateStatus(ctx, policy, maasv1alpha1.PhaseFailed, fmt.Sprintf("Failed to reconcile gateway AuthPolicy: %v", err), statusSnapshot)
 		return ctrl.Result{}, err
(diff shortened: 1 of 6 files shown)


jland-redhat:
I like the idea but just late in the game for 3.4 marking as a fast follow

Gajanan1995:
@jland-redhat Thanks for the review and for marking this as a fast follow! Happy to rebase once the target branch for 3.5 is ready, or I can rebase against `main` now.

liangwen12year (commented):
Nice work on this feature — the per-model AuthPolicy anonymous rule is well-scoped and the CodeRabbit iterations look solid. One concern on the gateway-level changes:

**Gateway default-deny/TRLP hardcodes `/docs` and `/openapi.json` exclusions globally**

The changes to `gateway-default-deny.yaml` and `gateway-default-auth.yaml` unconditionally exclude `/docs` and `/openapi.json` from rate limiting and auth for **all** models on the gateway, regardless of whether any model has `publicPaths` configured.

```yaml
# gateway-default-deny.yaml
- predicate: '!request.path.startsWith("/maas-api") && !request.path.startsWith("/v1/models") && !request.path.endsWith("/docs") && !request.path.endsWith("/openapi.json")'
```

```yaml
# gateway-default-auth.yaml
when:
  - predicate: '!request.path.endsWith("/docs") && !request.path.endsWith("/openapi.json")'
```

This means a model that does **not** set `publicPaths` still has its `/docs` and `/openapi.json` bypass the gateway-level deny and rate limiting. The per-route AuthPolicy (generated by the reconciler) would still block unauthenticated access for models without `publicPaths`, but the defense-in-depth layer from the TRLP is gone for those paths on all models.

Should these gateway-level exclusions be removed so the feature is purely opt-in per model via the `publicPaths` field? The per-route anonymous auth rule already handles the bypass for models that explicitly configure it.

Minor items:
- `buildPublicPathsCELPredicate` is called twice with the same `publicPaths` input (~line 414 and ~558) — could hoist the variable to compute once.
- The test `contains()` helper reimplements `strings.Contains` — could just use the stdlib.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -227,6 +228,37 @@ const (
 		`("x-maas-subscription" in request.headers ? request.headers["x-maas-subscription"] : "")`
 )
 
+// buildPublicPathsRegex constructs a regex pattern that matches any of the given path suffixes.
+// Each path is fully escaped with regexp.QuoteMeta and anchored with .* prefix and $ suffix.
+// Example: ["/docs", "/openapi.json"] -> ".*/docs$|.*/openapi\.json$"
+func buildPublicPathsRegex(paths []string) string {
```
EgorLu: `endsWith("/docs")` and `.*\/docs$` match any path whose last segment is `/docs`, not just the model's documentation root.
A request to `/llm/my-model/v1/completions/docs` would match and bypass auth. Consider switching to an exact-path match against the model's route prefix + suffix.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -477,11 +509,23 @@ func (r *MaaSAuthPolicyReconciler) reconcileModelAuthPolicies(ctx context.Contex
 			}
 		}
 
+		// Aggregate publicPaths from ALL auth policies for this model
+		var publicPaths []string
+		for _, ap := range allPolicies {
+			for _, p := range ap.Spec.PublicPaths {
+				if err := validatePublicPath(p); err != nil {
```
EgorLu: This opens a possibility for human error of the admin.
For example, an admin could mistakenly set `publicPaths: ["/v1/completions"]` and disable authentication for a model's inference.

I see two possibilities here. One is restricting `publicPaths` to a curated allowlist of safe suffixes, such as `/docs`. Or, on the contrary, having a denylist of known inference path segments, such as `/completions`, `/chat`, etc.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
Gajanan1995: @EgorLu Good catch you're right that `endsWith ("/docs")` would also match crafted paths like `/v1/completions/docs`. I'll switch to exact-path matching using the model's route prefix so the match is scoped to `{routePrefix}{suffix}` rather than a loose suffix. Will update the PR.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
Gajanan1995: @EgorLu Valid concern. I think an allowlist of known-safe documentation paths is the cleaner approach as it's simpler, less fragile than maintaining a denylist, and there's a small well-known set `/docs, /openapi.json`. A denylist would need to keep up with new inference paths added over time.

Would a hardcoded allowlist like `["/docs", "/openapi.json"]` with CRD-level enum validation work, or do you think we should keep it as free-form strings but validate against the allowlist in the controller?

EgorLu: I think that your idea with a CRD-level enum is great for this usecase. We won't often need to update that list, so a rare update would justify a CRD update on demand :)

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
Gajanan1995: Changed from `endsWith("/docs")` / `.*\/docs$` (matches any path ending in `/docs`) to exact-path matching using the model's route prefix: `request.path != "/default/llm/docs"` / `^/default/llm/docs$`
A crafted path like `/v1/completions/docs` will no longer bypass auth

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
Gajanan1995: Changed `PublicPaths` from free-form `[]string` to a `[]PublicPath` enum type restricted to `/docs` and `/openapi.json` at the CRD level

EgorLu:
@Gajanan1995 The `lint` check is failing with 6 issues:

**1. `errcheck` (4 issues)** — unchecked type assertions need the `, ok` two-value form:
- `maasauthpolicy_controller.go:686` — `rule["authentication"].(map[string]any)`
- `maasauthpolicy_controller.go:701` — `rule["metadata"].(map[string]any)`
- `maasauthpolicy_controller.go:703` — `evaluator.(map[string]any)` (metadata loop)
- `maasauthpolicy_controller.go:858` — `evaluator.(map[string]any)` (authorization loop)

**2. `gci` (1 issue)** — import grouping in the test file (`maasauthpolicy_controller_test.go:1235`). Likely needs the standard blank-line separation between stdlib / external / internal groups.

**3. `gocyclo` (1 issue)** — `reconcileModelAuthPolicies` hit cyclomatic complexity 41 (limit 40). The new `publicPaths` logic pushed it just over. Extracting the public-paths block (anonymous auth rule + metadata/authorization `when` injection) into a helper like `applyPublicPaths(rule, authRules, publicPaths, routePrefix)` should bring it back under the threshold.

Also, the PR title doesn't pass the "Validate PR Title" check — the conventional commit type needs to come first:
- **Current:** `[RHOAIRFE-496]: feat: add publicPaths to MaaSAuthPolicy for unauthenticated access to docs endpoints`
- **Fix:** `feat: [RHOAIRFE-496] add publicPaths to MaaSAuthPolicy for unauthenticated access to docs endpoints`

EgorLu:
We'll have to wait, as code freeze is in effect, but I ran the tests to make sure that the pr is ready

EgorLu:
Tested on a cluster with RHOAI 3.5.0-ea.2. Built and deployed a controller image from this branch.

**With `publicPaths: ["/docs", "/openapi.json"]`:**
- Reconciles to `phase: Active`
- Gateway AuthPolicy contains `anonymous-public-paths` rule with correct path regex
- CEL skip predicates added to all metadata and authorization evaluators (`apiKeyValidation`, `subscription-info`, `auth-valid`, `require-group-membership`, `subscription-valid`, `tenant-gateway-isolation`)

**Without `publicPaths`:**
- No `anonymous-public-paths` rule, no skip predicates
- Behavior unchanged from main

EgorLu (approved):
LGTM

jland-redhat (approved):

PR #731, opened 2026-04-11
fix: patch params.env for custom image injection in kustomize mode

chaitanya1731:
## Summary

Fix `MAAS_API_IMAGE` and `MAAS_CONTROLLER_IMAGE` env vars being silently ignored during kustomize-mode deployments, causing CI to always deploy `latest` instead of PR images.

### Problem

When deploying MaaS via kustomize mode (used by Konflux CI and `prow_run_smoke_test.sh`), the `MAAS_API_IMAGE` environment variable was silently overridden. The deploy script logged the correct PR image, but the pod always ended up running `latest`:

```
# Logs showed correct image:
Using custom MaaS API image: quay.io/opendatahub/maas-api:odh-pr-721

# But the pod had:
image: quay.io/opendatahub/maas-api:latest
```

### Root Cause

The `shared-patches` kustomize component uses `replacements:` to set container images from `params.env`. But `set_maas_api_image()` and `set_maas_controller_image()` were only patching the base `images:` transformer in `kustomization.yaml`.

Kustomize processes `images:` transformers **before** `replacements:`, so `params.env` (hardcoded to `latest`) always overwrote the custom image. This regression was introduced when the `shared-patches` component was added to centralize overlay configuration.

The maas-controller was not visibly affected because `deploy.sh` had a post-apply `kubectl set image` workaround that corrected the image after kustomize had already applied it with the wrong tag. maas-api had no such workaround.

### Fix

- Add `_patch_params_env` helper to patch `params.env` with custom image values before `kustomize build`, so replacements pick up the correct image.
- Call `_patch_params_env` from both `set_maas_api_image` and `set_maas_controller_image` after the existing base kustomization patching.
- Add `_cleanup_params_env` to restore `params.env` from backup after build.
- Remove the post-apply `kubectl set image` workaround for maas-controller in `deploy.sh` since `params.env` now carries the correct image through the kustomize build pipeline.
- Log deployed maas-api and maas-controller images at end of deployment for easier verification.

## How it was tested

- Verified locally that `kustomize build` with the old approach (patching base `images:` transformer) still produces `latest` — confirming the bug.
- Verified locally that `kustomize build` with the fix (patching `params.env`) produces the correct custom image.
- Tested both `tls-backend` and `http-backend` overlays — both produce correct images.
- Verified operator mode is unaffected (base `images:` transformer still works for direct base builds).
- Verified default behavior (no env var set) still produces `latest`.
- Verified `params.env` is restored from backup after deployment (no leftover `.backup` file).
- Deployed on a live cluster with `MAAS_API_IMAGE` and `MAAS_CONTROLLER_IMAGE` set — both pods running correct PR images.

## Summary by CodeRabbit

* **Improvements**
  * Deployment now always logs the live container images for key services to help verification and troubleshooting, with safe fallbacks when data is missing.
  * Image update operations also persistently update deployment configuration so chosen images remain synchronized across tools and are cleanly restored during cleanup.
  * Final completion message changed to “Models-as-a-Service Deployment completed successfully!” to reflect branding.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -555,44 +555,6 @@ main() {
     log_info "  Subscription controller ready."
     log_info "  Create MaaSModelRef, MaaSAuthPolicy, and MaaSSubscription to enable per-model auth and rate limiting."
 
-    # When using a custom controller image, annotate deployment to prevent operator reconciliation
-    # and patch the deployment with the custom image
-    if [[ -n "${MAAS_CONTROLLER_IMAGE:-}" ]]; then
-      # Log the current image before patching
-      local actual_image
-      actual_image=$(kubectl get deployment/maas-controller -n "$NAMESPACE" -o jsonpath='{.spec.template.spec.containers[0].image}' 2>/dev/null || echo "")
-      log_info "  Controller image before patch: $actual_image"
-      log_info "  Expected image: $MAAS_CONTROLLER_IMAGE"
-      
-      # Step 1: Annotate to prevent operator from reverting our changes
-      log_info "  Annotating maas-controller deployment to prevent operator reconciliation..."
-      kubectl annotate deployment/maas-controller -n "$NAMESPACE" \
-        opendatahub.io/managed="false" --overwrite 2>/dev/null || true
-      
-      # Step 2: Patch the deployment with the custom image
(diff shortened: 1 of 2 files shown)


jrhyness (approved):
/lgtm

PR #733, opened 2026-04-11
test: fix `test_subscription_status_transitions_on_model_deletion()`

yu-teo:
## Description
My assumption is that `validateModelRefs()` needs to rely on `deletionTimestamp` just like `findHTTPRouteForModel()` does. 
So, the failing scenario must be the following:
`MaaSModelRef` has a finalizer, so when you delete it:
  1. Kubernetes sets `deletionTimestamp` but the object continues to exist in the cache until the finalizer is removed
  2. `validateModelRefs()` calls r.Get() which succeeds (object still exists). Hence, it sets `ready=true` and `reason=Valid`
  3. `checkTokenRateLimitHealth()` has `findHTTPRouteForModel()` that calls r.Get() which succeeds, but then checks `deletionTimestamp` returns `ErrModelNotFound`
  4. `deriveFinalPhase()` correctly detects the inconsistency via TRLP health and sets `phase=Failed`
  5. But `updateStatus()` persists with `phase=Failed` while `modelRefStatuses=[{ready: true, reason: Valid}]` since `modelRefStatuses` was already set with `ready=true` and never corrected

The model's finalizer cleanup (deleting AuthPolicies, TRLPs, backend resources) can take time, so the model remains in "deleting" state for the duration. During this window, every reconciliation produces the same stale **ready=true** in `modelRefStatuses`.

**After the change:**
  1. Test deletes model, so `deletionTimestamp` is set (finalizer might still be completing its task hence the object remain alive)
  2. Subscription reconciles:
    - `validateModelRefs()` sets `ready=true`
    - `checkTokenRateLimitHealth()` sets `BackendNotReady`
    - fix: `modelRefStatuses[0]` from `ready=true` to `ready=false` and `reason=NotFound`
    - `deriveFinalPhase()` sets `phase=Failed`
    - `updateStatus()` saves `phase=Failed` and `modelRefStatuses=[ready=false]`
  3. Test's _wait_for_subscription_phase("Failed") succeeds
  4. Test's poll for `modelRefStatuses[0].ready` is `False` succeeds immediately because the status was corrected in the same reconciliation

## How Has This Been Tested?
Tests pass

## Merge criteria:

## Release Notes

* **Bug Fixes**
  * Fixed an issue where models marked for deletion would incorrectly remain in the "ready" state. The system now properly identifies models undergoing deletion and corrects their status to "not found" to ensure accurate health reporting and phase information.

* **Tests**
  * Added test coverage for model deletion scenarios to verify proper status correction during reconciliation.


jrhyness (approved):
/lgtm

PR #735, opened 2026-04-13
feat(maas-controller): maas`Tenant` CR and reconciler

ishitasequeira:
## Description

Adds the `Tenant` CR (`maas.opendatahub.io/v1alpha1`) and a platform reconciliation pipeline to `maas-controller` so it can render and apply MaaS platform workloads (maas-api, gateway config, auth policies, telemetry).
`Tenant` replaces the previous `ModelsAsService` (`components.platform.opendatahub.io/v1alpha1`) as the persisted CR for the MaaS component. ODH still uses the `ModelsAsService` component name internally for enablement checks, labels, and DSC status aggregation, but the object on the cluster is now `Tenant`. This gives `maas-controller` full ownership of platform workload lifecycle while ODH retains control of the component lifecycle (install, enable/disable, cleanup).

### What's included
- `Tenant` CRD with API key, external OIDC, gateway, and telemetry configuration
- `TenantReconciler`: prerequisites → dependencies → kustomize render → post-render → SSA apply → deployment readiness
- Post-render: gateway AuthPolicy/TokenRateLimitPolicy/DestinationRule targeting, external OIDC patching, TelemetryPolicy + IstioTelemetry injection, config-hash rollout annotation
- Finalizer with cross-namespace cleanup via tracking labels
- Management state support (Managed/Unmanaged/Removed)
- Unit tests for reconcile, finalization, singleton enforcement, management states

### Design decisions (based on review feedback)
- **Namespace-scoped**: lives in `models-as-a-service` alongside `MaaSSubscription`/`MaaSAuthPolicy`. First release with no deployed CRDs — avoids a CRD scope migration later (Kubernetes does not allow changing scope on an existing CRD)
- **Self-bootstrap**: `maas-controller` creates the default Tenant on startup; ODH operator's `NewCRObject` is a no-op
- **No DSCI dependency**: app namespace derived from `tenant.Namespace` — no cross-operator API calls or extra RBAC
- **Cross-namespace ownership**: tracking labels for cluster-scoped/cross-namespace children; `ownerReferences` for same-namespace only
- **Singleton via CEL**: `self.metadata.name == 'default-tenant'` — removing the rule later enables multi-tenancy without CRD migration
- **Gateway policy alignment**: `gateway-default-auth` (AuthPolicy) and `gateway-default-deny` (TokenRateLimitPolicy) names match actual manifests

_Related ODH PR:_ https://github.com/opendatahub-io/opendatahub-operator/pull/3412

## How Has This Been Tested?

- Unit tests for the reconcile entry-point (`maastenant_reconcile_test.go`)
- Manual End-to-end testing on ROSA cluster with custom ODH operator + maas-controller images:
  - Verified self-creation of `default-tenant` in `models-as-a-service` namespace
  - Platform workloads applied via SSA
  - Toggled MaaS off/on in DSC to verify cleanup and re-provisioning
  - CRD namespace scope and CEL singleton enforcement confirmed

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Adds a Tenant custom resource with validation, status/phase, and CLI printer columns.
  * Ships a controller that ensures a singleton default Tenant, reconciles rendered manifests, monitors readiness, and performs safe teardown.

* **New Features (rendering)**
  * Rendering/post-processing injects OIDC, telemetry policies, gateway defaults, params, and a deterministic config-hash on deployments.

* **Chores**
  * Expanded RBAC, dependency and linter updates, deployment script improvements, and added reconciliation tests.


deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maastenants.yaml
```diff
@@ -0,0 +1,207 @@
...
+    controller-gen.kubebuilder.io/version: v0.16.4
+  name: maastenants.maas.opendatahub.io
+spec:
+  group: maas.opendatahub.io
+  names:
+    kind: MaaSTenant
+    listKind: MaaSTenantList
+    plural: maastenants
+    singular: maastenant
+  scope: Cluster
```
lphiri: We expect this to be a tenant managed object in the future, so  I suggest we make it namespaced, so we can better limit access.

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maastenants.yaml
```diff
@@ -0,0 +1,207 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.4
+  name: maastenants.maas.opendatahub.io
+spec:
+  group: maas.opendatahub.io
+  names:
+    kind: MaaSTenant
```
lphiri: Naming nit,  I would drop "maas" from the object names, because its redundant and already part of the group name - how about 
simply "Tenant":
tenants.maas.opendatahub.io,
Tenant, TenantList  etc?

deployment/base/maas-controller/rbac/clusterrole.yaml
```diff
@@ -17,22 +17,89 @@ rules:
...
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - delete
+  - get
+  - list
+  - watch
```
lphiri, line 37: Do we need all these permissions?

deployment/base/maas-controller/rbac/clusterrole.yaml
```diff
@@ -17,22 +17,89 @@ rules:
...
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterrolebindings
+  - clusterroles
+  verbs:
+  - delete
+  - get
+  - list
+  - watch
```
lphiri, line 74: Do we need all these permissions?

maas-controller/api/maas/v1alpha1/maastenant_types.go
```diff
@@ -0,0 +1,162 @@
...
+type MaaSTenant struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   MaaSTenantSpec   `json:"spec,omitempty"`
+	Status MaaSTenantStatus `json:"status,omitempty"`
+}
+
+// MaaSTenantSpec defines the desired state of MaaSTenant (spec parity with ModelsAsService).
+type MaaSTenantSpec struct {
```
lphiri: See earlier comment, I think we want this to be namespace scoped, so need namespace here?

maas-controller/api/maas/v1alpha1/maastenant_types.go
```diff
@@ -0,0 +1,162 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// MaaSTenantKind is the API kind for the cluster MaaS tenant / platform singleton.
```
lphiri: I  think the singleton needs a bit more discussion. 
I dont know if we want to manage this at all,
or ask the admin to create a default instance as part of the set up and just have our controller materialize it. 
This is the approach taken by ModelRegistry and MLFlow integration, or even Kuadrant.

maas-controller/api/maas/v1alpha1/tenant_types.go
```diff
@@ -0,0 +1,162 @@
...
+	MaaSTenantKind = "MaaSTenant"
+	// MaaSTenantInstanceName is the singleton resource name enforced by the API.
+	MaaSTenantInstanceName = "default-maastenant"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == 'default-maastenant'",message="MaaSTenant name must be default-maastenant"
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`,description="Ready"
```
lphiri, line 34: See comment above about not making this cluster scoped.
We could still have a rule that says there can only be one instance named "default-tenant"

lphiri (commented):
@ishitasequeira 

Lets continue review after further discussion - I realize my suggestions here will have significant cascading effects to the rest of your changes.

maas-controller/api/maas/v1alpha1/tenant_types.go
ugiordan: The CEL rule `self.metadata.name == 'default-maastenant'` baked into the CRD means changing the singleton pattern later requires a CRD schema migration, since CEL rules are part of the structural schema. If multi-tenancy is planned for a future release, this will be painful to evolve. A webhook-based approach or a namespaced resource (as @lphiri suggested) would be easier to relax later.

maas-controller/pkg/controller/maas/maastenant_finalize.go
```diff
@@ -0,0 +1,322 @@
...
+
+func (r *MaaSTenantReconciler) operatorNamespace() string {
+	if r.OperatorNamespace != "" {
+		return r.OperatorNamespace
+	}
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+		return ns
+	}
+	return os.Getenv("WATCH_NAMESPACE")
+}
```
ugiordan: This function fallback chain (field -> POD_NAMESPACE -> WATCH_NAMESPACE) silently returns an empty string if all are unset. When `watchNamespaces` is empty, the finalizer loop skips cleanup entirely and removes the finalizer anyway. This orphans all tenant resources (Deployments, Services, NetworkPolicies, RBAC) with no error logged.

Also, the finalizer is removed during `Removed` and `Unmanaged` management state transitions (not just deletion). If an admin sets the state to Removed and later deletes the CR, the finalizer is already gone, and no cleanup runs.

We should return an error and requeue if namespaces resolve to empty, rather than silently dropping the finalizer.

maas-controller/pkg/controller/maas/tenant_controller.go
```diff
@@ -0,0 +1,151 @@
...
+		Watches(
+			&corev1.ConfigMap{},
+			handler.EnqueueRequestsFromMapFunc(enqueueDefaultMaaSTenant),
+			builder.WithPredicates(deletedConfigMapOnly()),
+		).
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(enqueueDefaultMaaSTenant),
+			builder.WithPredicates(secretNamedMaaSDB()),
+		).
```
ugiordan, line 177: This watches for ConfigMap deletes without a namespace predicate. Every ConfigMap deletion across the entire cluster triggers a reconcile event. On a busy cluster, this is a significant performance concern. The same applies to Secret watches. These should have namespace-scoped predicates or use `cache.Options.ByObject` field selectors.

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -0,0 +1,280 @@
...
+			return fmt.Errorf("gateway %s/%s not found: the specified Gateway must exist before enabling MaaS platform reconcile", namespace, name)
+		}
+		return fmt.Errorf("failed to look up gateway %s/%s: %w", namespace, name, err)
+	}
+	return nil
+}
+
+func (r *MaaSTenantReconciler) patchStatus(ctx context.Context, tenant *maasv1alpha1.MaaSTenant, status metav1.ConditionStatus, reason, message string) error {
+	apimeta.SetStatusCondition(&tenant.Status.Conditions, metav1.Condition{
+		Type:               tenantreconcile.ReadyConditionType,
```
ugiordan, line 283: `patchStatus` only sets the `Ready` condition, but several error paths in the reconcile loop call `Status().Update()` directly and set `Phase` without touching conditions. This creates a window where `.status.phase` says "Failed" but `.status.conditions[Ready]` still says "True" from a previous reconcile. Pick one pattern (preferably `patchStatus` everywhere) and make it set both phase and conditions atomically.

maas-controller/pkg/platform/tenantreconcile/apply.go
```diff
@@ -0,0 +1,130 @@
...
+func ApplyRendered(ctx context.Context, c client.Client, scheme *runtime.Scheme, tenant *maasv1alpha1.MaaSTenant, objs []unstructured.Unstructured) error {
+	for i := range objs {
+		u := objs[i].DeepCopy()
+		if err := controllerutil.SetControllerReference(tenant, u, scheme); err != nil {
+			return fmt.Errorf("set controller reference on %s %s/%s: %w", u.GetKind(), u.GetNamespace(), u.GetName(), err)
+		}
+		unstructured.RemoveNestedField(u.Object, "metadata", "managedFields")
+		unstructured.RemoveNestedField(u.Object, "metadata", "resourceVersion")
+		unstructured.RemoveNestedField(u.Object, "status")
+		if err := c.Patch(ctx, u, client.Apply, client.FieldOwner(ssaFieldOwner), client.ForceOwnership); err != nil {
```
ugiordan, line 135: `ForceOwnership: true` in SSA means this controller will silently steal fields managed by other controllers (the ODH operator, for example) without conflict errors. If both controllers attempt to manage the same resource, the last writer wins with no warning. 
This is particularly risky during the migration period, where both `ModelsAsService` and `MaaSTenant` paths might be active. Consider using `ForceOwnership: false` and handling conflicts explicitly, at least during the transition.

maas-controller/pkg/controller/maas/maastenant_reconcile_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	res, err := r.Reconcile(context.Background(), ctrl.Request{NamespacedName: types.NamespacedName{Name: tenant.Name}})
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(res.RequeueAfter).To(Equal(finalizeRequeueInterval))
+
+	var updated maasv1alpha1.MaaSTenant
+	g.Expect(cl.Get(context.Background(), client.ObjectKey{Name: tenant.Name}, &updated)).To(Succeed())
+	g.Expect(updated.Finalizers).To(ContainElement(maasTenantFinalizer))
+}
```
ugiordan: I see only 2 tests here, both covering the deletion/finalizer path. There are no tests for happy-path reconciliation, status transitions, management state changes, gateway configuration, or error handling. We should cover, at a minimum, the happy path (create -> reconcile -> Ready) and management state transitions (Managed -> Removed -> cleanup).

maas-controller/pkg/platform/tenantreconcile/prerequisites.go
```diff
@@ -0,0 +1,227 @@
...
+func ApplicationNamespace(ctx context.Context, c client.Client) (string, error) {
+	if ns := os.Getenv("RHAI_APPLICATIONS_NAMESPACE"); ns != "" {
+		return ns, nil
+	}
+
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(dsciListGVK)
+	if err := c.List(ctx, list); err != nil {
+		return "", fmt.Errorf("list DSCInitialization: %w", err)
+	}
```
ugiordan: `ApplicationNamespace` reads the DSCInitialization singleton directly from the API server to figure out which namespace to deploy workloads into. This creates a coupling, I think we should avoid:

1. **Extra RBAC**: the maas-controller needs `list` on `dscinitializations.dscinitialization.opendatahub.io` cluster-wide, just to read one field from a CR it doesn't own.
2. **Fragile at runtime**: if the DSCI CR doesn't exist yet (bootstrap, someone deletes it), the maas-controller can't reconcile at all. It's blocked on a resource it doesn't control.
3. **Silent breakage on API changes**: you're using unstructured access to `spec.applicationsNamespace`. If ODH renames that field or bumps the API version, this silently returns empty, and workloads go to the wrong namespace (or fail).

The ODH operator already knows `applicationsNamespace` from its own DSCI when it creates the MaaSTenant CR (in the [operator-side PR](https://github.com/opendatahub-io/opendatahub-operator/pull/3412)). It could just write it into `MaaSTenantSpec` at creation time, so the maas-controller reads it from its own CR. No cross-operator API calls, no extra RBAC, one-directional data flow: ODH operator -> MaaSTenant CR -> maas-controller.

Alternatively, pass it as a controller flag or env var, since ODH is also deploying the maas-controller itself.

deployment/base/maas-controller/rbac/clusterrole.yaml
ishitasequeira: removed the `delete` permission, but rest are needed for adding a watch on `maas-db-config` secret.

deployment/base/maas-controller/rbac/clusterrole.yaml
ishitasequeira: The `maas-controller`'s TenantReconciler now manages the `maas-api` Deployment as part of the platform stack. We need these permissions for managing `maas-api` deployment.

maas-controller/api/maas/v1alpha1/tenant_types.go
ishitasequeira: We moved to a namespace scoped Tenant CR. We are still keeping the singleton pattern to be managed by maas-controller for now.

maas-controller/pkg/controller/maas/tenant_controller.go
ishitasequeira: Added a restriction on tenant namespace.

maas-controller/pkg/platform/tenantreconcile/apply.go
ishitasequeira: maas-controller is now the sole owner of Tenant resource.

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -0,0 +1,279 @@
...
+			return ctrl.Result{}, err2
+		}
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
+	if r.ManifestPath == "" {
+		if err := r.patchStatus(ctx, &tenant, "Failed", metav1.ConditionFalse, "ManifestPathUnset",
+			"MAAS_PLATFORM_MANIFESTS is not set and no default kustomize path resolved; cannot apply platform manifests"); err != nil {
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
ishitasequeira: `main.go` always sets a non-empty default, and `RenderKustomize` fails clearly with `"kustomize build ... no such file"` if the path is bad. Adding an `os.Stat` check here is redundant.

maas-controller/pkg/platform/tenantreconcile/prerequisites.go
```diff
@@ -0,0 +1,234 @@
...
+	log := log.FromContext(ctx)
+	var rep PrerequisiteReport
+
+	if msg := checkAuthorinoTLS(ctx, c); msg != "" {
+		rep.Warnings = append(rep.Warnings, msg)
+		log.V(1).Info("MaaS prerequisite warning", "check", "authorino-tls", "message", msg)
+	}
+	if msg := checkDatabaseSecret(ctx, c, appNamespace); msg != "" {
+		rep.Blocking = append(rep.Blocking, msg)
+		log.Error(nil, "MaaS prerequisite error", "check", "database-secret", "message", msg)
```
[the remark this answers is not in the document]
ishitasequeira: DB secret is required for starting `maas-api`. Logging it as error helps to highlight the issue.

maas-controller/pkg/platform/tenantreconcile/kustomize.go
```diff
@@ -0,0 +1,114 @@
...
+	return fs.Exists(p)
+}
+
+// DefaultManifestPath returns MAAS_PLATFORM_MANIFESTS or a dev default relative to cwd (models-as-a-service repo layout).
+func DefaultManifestPath() string {
+	if v := os.Getenv("MAAS_PLATFORM_MANIFESTS"); v != "" {
+		return v
+	}
+	return "../maas-api/deploy/overlays/odh"
+}
```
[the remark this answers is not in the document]
ishitasequeira: in-cluster deployment always sets the env var/flag; dev default is intentional for local development.

maas-controller/pkg/platform/tenantreconcile/prerequisites.go
ishitasequeira: Good points!! I fully agree. Two changes address all three concerns:

* `Tenant` is now namespace-scoped and co-located with `MaaSSubscription/MaaSAuthPolicy` in `models-as-a-service`. Since the Tenant CR's namespace is the application namespace, we now use `tenant.Namespace` directly.

* Removed ApplicationNamespace entirely — the function, the `dsciListGVK` variable, and the DSCI RBAC marker are all gone. The controller no longer reads DSCInitialization at all, eliminating the extra RBAC, runtime fragility, and unstructured field access concerns you raised.

maas-controller/api/maas/v1alpha1/tenant_types.go
ugiordan: Since the plan is to drop the singleton pattern later,  I'd suggest tracking that as a follow-up so it doesn't get buried. The CEL name constraint in the CRD schema is the part that's hard to evolve, so the sooner it's removed, the fewer upgrade migrations you'll need to deal with.

deployment/base/maas-controller/rbac/clusterrole.yaml
ugiordan: The justification makes sense at a high level, but it would help to see a per-resource breakdown. For example, does the controller actually need `delete` on `clusterroles` and `clusterrolebindings`, or just `get/list/watch`? Same question for `networkpolicies`. Broad RBAC is a common source of CVE surface in operators, so we should be explicit about what's create/update/delete vs read-only.

jland-redhat:
Required commit:
https://github.com/openshift/release/pull/77925

jland-redhat (approved):
lgtm

deployment/base/maas-controller/rbac/clusterrole.yaml
lphiri: The DB secret has a specific name we know ahead of time.
I suggest creating a role with a resourceName constraint to grant access to only that secret.

deployment/base/maas-controller/rbac/clusterrole.yaml
```diff
@@ -135,6 +174,29 @@ rules:
...
+  resources:
+  - clusterrolebindings
+  - clusterroles
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - watch
```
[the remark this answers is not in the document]
lphiri: Lets revisit the need for this to see if there is a way to scope down the permissions.

PR #748, opened 2026-04-15
docs: improve validation guide

EgorLu:
## Summary

I've tried rewriting the doc in such a way, that a completely new user, who might be confused, will understand what they are supposed to do.

- Restructure validation page: lead with `validate-deployment.sh` ("Quick Validation"), then step-by-step
- Add intro explaining what to verify and why (infrastructure, model inference, security)
- Explain the script's 4 check phases (components, gateway, policies, endpoints) with sample output
- Add expected output examples per step in collapsible blocks
- Add per-step error guidance (what 401/403/500/503 means at that step)
- Add OC token vs API key callout
- Switch inference examples to `/v1/chat/completions`

## Depends on

Builds on changes from #687 (endpoint format, optional steps). May need a rebase after that PR merges.

## Test plan

- [x] Verify MkDocs renders correctly (`mkdocs serve` in docs/)
- [x] Verify all anchors still resolve (linked from troubleshooting.md, tls-configuration.md)
- [x] Verify collapsible blocks expand/collapse correctly
- [x] Walk through the manual steps on a deployed cluster

## Summary by CodeRabbit

* **Documentation**
  * Replaced the single manual walkthrough with a consolidated step-by-step validation flow and expanded automated validation (per-model invocation).
  * Clarified TLS verification, credential distinctions (admin token vs model API key), gateway discovery fallback, and non-admin/OpenShift constraints.
  * Stopped displaying truncated API keys, added more failure-case guidance, normalized examples to chat-based inference with a fallback, and updated rate-limit test output expectations.


docs/content/install/validation.md
```diff
@@ -1,13 +1,82 @@
...
+
+1. That the infrastructure is healthy: pods are running, the gateway is accepting traffic, and auth/rate-limit policies are enforced.
+2. That you can deploy a model and talk to it through the gateway: create an API key, list available models, and send an inference request.
+3. That security is active: requests without credentials are rejected and rate limits kick in after a few calls.
 
 !!! note "Prerequisite"
     At least one model must be deployed to validate the installation. See [Model Setup (On Cluster)](model-setup.md) to deploy sample models.
 
-## Manual Validation (Recommended)
+## Quick Validation
```
jland-redhat: So "Quick Validation" is covering the same info as "Automated Validation".

Also I would prefer if this was not the first thing on the page because I do really want people to run through the manual validation first as it does a better job of helping them understand issues they are running into.

jland-redhat (changes requested):
There is actually one request other request I have if you are changing this page.

Can we update the validation steps so that it is a little easier to extract the curl commands (like maybe something that puts the curl command in an env variable or something)

Right now if anything fails you just get a JQ error which is not super helpful

docs/content/install/validation.md
EgorLu: Oh ok, I assume that automated validation is being preferred now. Good point.

docs/content/install/validation.md
```diff
@@ -17,39 +24,68 @@ HOST="https://maas.${CLUSTER_DOMAIN}" && \
 echo "Gateway endpoint: $HOST"
 ```
 
+Extract the cluster CA certificate so that `curl` can verify TLS without the `-k` flag:
```
jland-redhat: I feel like this adds complexity to this page that is not really needed.

docs/content/install/validation.md
```diff
@@ -63,66 +99,214 @@ echo "API key obtained: ${API_KEY:0:20}..."
...
 ```
 
-The script automates the manual validation steps above and provides detailed feedback with specific suggestions for fixing any issues found. This is useful when you need to quickly verify deployment status, but understanding the manual steps above helps with troubleshooting.
+To validate a specific model:
+
+```bash
+./scripts/validate-deployment.sh <model-name>
+```
+
+!!! note "Non-admin users"
```
jland-redhat: This is a good note but it comes so much later in the page after we run through the validation I am not sure folks would see it

docs/content/install/validation.md
```diff
@@ -63,66 +99,214 @@ echo "API key obtained: ${API_KEY:0:20}..."
...
+### What the script checks
+
+The script runs four groups of checks:
+
+1. **Component status** -- MaaS API pods, policy engine pods, RHOAI/KServe pods, and deployed models
+2. **Gateway status** -- gateway is accepted and programmed, HTTPRoute is configured, hostname is reachable
+3. **Policy status** -- AuthPolicy is enforced, TokenRateLimitPolicy is configured
+4. **API endpoint tests** -- authentication token works, API key creation works, models endpoint responds, model inference returns a result, rate limiting kicks in, unauthorized requests are rejected
+
+### Reading the output
```
jland-redhat, line 273: This is another one where I am not seeing the value in writing what the script outputs. I feel like it is pretty self explanatory

jland-redhat (changes requested):
There is some stuff I like here but I feel like it adds too much complexity to this page overall.

docs/content/install/validation.md
```diff
@@ -18,38 +25,59 @@ echo "Gateway endpoint: $HOST"
...
     ```bash
-    HOST="https://gateway.${CLUSTER_DOMAIN}" && \
-    echo "Using fallback gateway endpoint: $HOST"
+    HOST="https://maas.apps.your-cluster.example.com"
+    ```
+
+??? success "Expected output"
+    ```text
+    Gateway endpoint: https://maas.apps.your-cluster.example.com
     ```
```
ishitasequeira: Have we verified how does the `??? success "Expected output:` render? The rendering seems off in github docs.

docs/content/install/validation.md
```diff
@@ -18,38 +25,59 @@ echo "Gateway endpoint: $HOST"
...
 
+**If this fails:**
+
+- Empty `CLUSTER_DOMAIN`: you don't have permission to read the ingress config. Set `HOST` manually as shown above.
+
```
ishitasequeira: Looks like we removed `CLUSTER_DOMAIN` from the example above. Adding this note is not required in that case.

docs/content/install/validation.md
EgorLu: Good point. I used the new mkdocs (was testing material theme compatibility) and didn't test here on Github.
Replaced all `???` success collapsibles with the plain text.

docs/content/install/validation.md
EgorLu: `CLUSTER_DOMAIN` is still in the command above, but the "If this fails" is redundant with the `!!! note` right above it. Removed it

docs/content/install/validation.md
EgorLu: Agreed and removed

docs/content/install/validation.md
EgorLu: Good point, moved upwards

docs/content/install/validation.md
EgorLu: Yeah, honestly, I also thought that it was redundant, but was ensure if this is what we wanted. Removed.

jland-redhat:
We will need to do some validation script updates in general. This is a good reference point I think but going to close this since we require more changes than just general clean up.

PR #753, opened 2026-04-15
fix: use targetModel in HTTPRoute header match

### LINKED ISSUE #745: fix: ExternalModel reconciler fails with dots in name and lacks namespace isolation in header match
## Problem

Two issues in the ExternalModel reconciler:

### 1. Dots in model names break Service creation
When an ExternalModel name contains dots (e.g., `openai.gpt-oss-20b`), the reconciler creates a Service with the same name. Services require DNS-1035 names (no dots), so the creation fails:
```
Service "openai.gpt-oss-20b" is invalid: a DNS-1035 label must consist of lower case alphanumeric characters or '-'
```

### 2. Same-name models in different namespaces collide on header-based routing
The HTTPRoute header match uses just the model name (`X-Gateway-Model-Name: openai.gpt-oss-20b`). When two namespaces have models with the same name, both HTTPRoutes match the same header value. After BBR ClearRouteCache, Envoy picks the first match — breaking namespace isolation.

## Fix

1. Sanitize dots to dashes in Service names (`sanitizeDNSName`)
2. Use `namespace/name` in the header match value (e.g., `llm/openai.gpt-oss-20b`)

## Tested

On RHOAI cluster with two models named `openai.gpt-oss-20b`:
- `llm/openai.gpt-oss-20b` → real AWS Bedrock (200)
- `prod-models/openai.gpt-oss-20b` → simulator (200)

Both route correctly to different backends.

noyitz:
## Problem

The HTTPRoute header-based rule matches `X-Gateway-Model-Name` against
the ExternalModel `metadata.name`. This breaks when `targetModel`
differs from the name (e.g., Bedrock models: `name=my-bedrock`,
`targetModel=openai.gpt-oss-20b`).

The user sends `targetModel` in the request body. BBR's
`body-field-to-header` plugin extracts it as `X-Gateway-Model-Name`.
After ClearRouteCache, the header doesn't match → `route_not_found`.

## Fix

Pass `targetModel` to `buildHTTPRoute` and use it in the header match
value instead of `name`.

## Changes

- `reconciler.go`: pass `extModel.Spec.TargetModel` to `buildHTTPRoute`
- `resources.go`: accept `targetModel` param, use in header match
- `resources_test.go`: update existing test, add test case where
  targetModel differs from name

## Tested

On RHOAI cluster:
- `llm/my-bedrock` (targetModel: `openai.gpt-oss-20b`) → Bedrock 200 ✓
- Header match correctly uses `openai.gpt-oss-20b` not `my-bedrock`

Fixes #745

```release-note
NONE
```

## Release Notes

* **Improvements**
  * Enhanced HTTP routing logic for external models to separately use target model identifiers in request matching, enabling more precise routing when the model name differs from its target model designation.

### DIFF
--- a/maas-controller/pkg/reconciler/externalmodel/reconciler.go
+++ b/maas-controller/pkg/reconciler/externalmodel/reconciler.go
@@ -182,7 +182,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	// 4. HTTPRoute (routes requests to external provider via gateway)
-	hr := buildHTTPRoute(extModel.Spec.Endpoint, name, ns, port, gwName, gwNamespace, labels)
+	hr := buildHTTPRoute(extModel.Spec.Endpoint, name, extModel.Spec.TargetModel, ns, port, gwName, gwNamespace, labels)
 	if err := controllerutil.SetControllerReference(extModel, hr, r.Scheme); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to set owner on HTTPRoute: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


nirrozenbaum (approved):
/lgtm
/approve

great catch!

PR #771, opened 2026-04-19
fix: minor updates for external model

nirrozenbaum:
the CRD initially was very restrictive, with enum specifying the providers. 
the PR make the field less strict, to avoid some issues around it in payload processing.
generally speaking, this CRD would need to update, and current change is a temp solution.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Extended provider support to accept custom provider values beyond the previously hardcoded set, enabling greater flexibility in model configuration.

* **Bug Fixes**
  * Refined resource access permissions to improve security posture by limiting scope to essential resources.


EgorLu (approved):
LGTM (but I would definitely add a simple regex to limit provider names to latin lowercase alphanumerics and underscore)

PR #805, opened 2026-04-27
feat(observability): deploy Perses dashboards via maas-controller

ishitasequeira:
## Description
The ODH operator previously managed Perses dashboard and datasource resources for MaaS (opendatahub-io/opendatahub-operator#3381 + opendatahub-io/models-as-a-service#686). This commit replicates that capability in maas-controller as part of the lifecycle migration.
https://redhat.atlassian.net/browse/RHOAIENG-59675
## How Has This Been Tested?
Manually tested on a cluster:
* With COO absent: MaaS deploys successfully; Perses resources are skipped (logged as skipping resource: optional CRD not yet registered).
* With COO present: after installing COO, the CRD watch triggers a reconcile and Perses resources (PersesDashboard, PersesDatasource) are applied automatically.
* Dashboard is visible and loads data correctly in the Perses UI.

## Merge criteria:

### DIFF
--- a/deployment/overlays/odh/kustomization.yaml
+++ b/deployment/overlays/odh/kustomization.yaml
@@ -43,7 +43,6 @@ resources:
   - ../../base/maas-api/overlays/tls  # maas-api with TLS (includes DestinationRule + NetworkPolicy)
   - ../../base/maas-controller/default
   - ../../base/payload-processing/default  # BBR ext_proc for external model payload processing
-  - ../../components/observability/observability/dashboards/
 
 # Include shared-patches component for common configuration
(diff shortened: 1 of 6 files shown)


ahadas (commented):
@ishitasequeira this looks good
another thing we lost in the process of moving this part from the odh-operator to Maas-controller is setting the MaaS CR as an owner of the dashboard - so that the dashboard would be removed when removing MaaS. not that common case to happen so I'm ok with doing it separately, but if you're already on filling the gaps that were created in that process, it would be nice to tackle this as well

jland-redhat (approved):
lgtm

PR #809, opened 2026-04-27
fix: scale down kserve-controller-manager to prevent crash-loop on Kind

yossiovadia:
## Summary
- The opendatahub fork `kserve-controller` image watches `Route.route.openshift.io` for `InferenceGraph`, which crash-loops on non-OpenShift clusters (Kind) even with stub CRDs installed.
- Scale `kserve-controller-manager` to 0 replicas instead of swapping its image to the ODH fork. The vanilla KServe install still provides CRDs, RBAC, and webhook configs.
- `LLMInferenceService` reconciliation is handled by the separate `llmisvc-controller-manager`, which runs fine on Kind.

## Test plan
- [ ] Run `local-deploy.sh` on a fresh Kind cluster and verify no pods are in CrashLoopBackOff
- [ ] Verify `llmisvc-controller-manager` is 1/1 Ready and handles LLMInferenceService resources
- [ ] Verify smoke test passes end-to-end

## Summary by CodeRabbit

* **Chores**
  * Updated test deployment script to simplify KServe controller initialization. The deployment process no longer uses a custom controller image and instead delegates inference service handling to a separate manager component.

### DIFF
--- a/test/e2e/scripts/local-deploy.sh
+++ b/test/e2e/scripts/local-deploy.sh
@@ -908,11 +908,11 @@ else
   # Install vanilla KServe (provides base controller, CRDs, webhooks)
   "$PROJECT_ROOT/scripts/installers/install-kserve.sh"
 
-  # Swap to opendatahub fork image (adds LLMInferenceService support)
-  kubectl set image deployment/kserve-controller-manager \
-    manager="$KSERVE_ODH_IMAGE" -n kserve
-  kubectl rollout status deployment/kserve-controller-manager -n kserve --timeout=120s
-  ok "KServe controller installed (opendatahub fork)"
+  # Scale down kserve-controller-manager — the opendatahub fork image watches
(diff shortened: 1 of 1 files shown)


noyitz (approved):
/lgtm

Tested locally on a fresh Kind cluster (Apple Silicon / arm64). Full deploy completed successfully — all 18 `--validate` checks passed:

- **No CrashLoopBackOff pods** — kserve-controller-manager scaled to 0/0, no crash loop
- **llmisvc-controller-manager 1/1 Ready** — handles LLMInferenceService correctly
- **External model inference**: llm-katan-echo via ExternalModel ✓
- **Internal model inference**: facebook/opt-125m via llm-d simulator ✓
- **Auth**: 401 without key, 403 with invalid key ✓
- **TLS**: cert-manager CA chain, SIMPLE mode on DestinationRule ✓

Clean fix — scaling to 0 avoids the Route.route.openshift.io crash-loop while keeping the CRDs, RBAC, and webhook configs from vanilla KServe intact.

PR #816, opened 2026-04-28
fix: mandate spec for MaaS CRs

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-59640

## Description

This PR:
1. mandates `spec` in MaaS CRs (MaaSAuthPolicy, MaaSSubscription, MaaSModelRef, Tenant); and
2. for the first three, considers backward compat and reports `Invalid` when `spec` doesn't exist.

## How Has This Been Tested?

1. Create no-spec CRs. Their phases remain empty.
2. Deploy the PR image.
3. Observe that no-spec CRs have `Invalid` phases.
4. Try creating new no-spec CRs and get blocked.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * API validation now requires the top-level spec for MaaS resources; resources missing spec are rejected and show clear error messages.
  * Controllers now short-circuit on missing/empty spec: resources are marked Invalid (Ready=False, reason InvalidSpec), skip finalizer installation and further reconciliation.

* **Tests**
  * Added reconciliation tests for legacy/empty-spec resources to verify deterministic Invalid status and no finalizers.

### DIFF
--- a/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_externalmodels.yaml
+++ b/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_externalmodels.yaml
@@ -169,6 +169,8 @@ spec:
                 - Failed
                 type: string
             type: object
+        required:
+        - spec
         type: object
     served: true
     storage: true
--- a/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasauthpolicies.yaml
+++ b/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasauthpolicies.yaml
@@ -250,6 +250,8 @@ spec:
                 - Failed
                 type: string
             type: object
+        required:
+        - spec
         type: object
     served: true
     storage: true
--- a/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasauthpolicies.yaml
+++ b/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasauthpolicies.yaml
@@ -173,6 +173,7 @@ spec:
                       - NotEnforced
                       - BackendNotReady
                       - ConditionsNotFound
+                      - InvalidSpec
                       - Unknown
                       type: string
                   required:
@@ -248,6 +249,7 @@ spec:
                 - Active
                 - Degraded
                 - Failed
+                - Invalid
                 type: string
             type: object
         required:
--- a/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasmodelrefs.yaml
+++ b/deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maasmodelrefs.yaml
@@ -181,6 +181,7 @@ spec:
                 - Ready
                 - Unhealthy
                 - Failed
+                - Invalid
                 type: string
             type: object
         required:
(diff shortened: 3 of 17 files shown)


jrhyness:
A few things:

- Validation in controller/reconciler is missing for Tenant and ExternalModel
- Tests for same are also missing
- I'm not sure a spec should even be required for a Tenant, though.    There are defaults, I believe?  We need to be careful not to screw up the tenant.  
- This is going in post-GA .  We need to be  concerned about backward compatibility and migration.   Can you find a way to migrate and/or document what to look for/how to recover?

jrhyness (changes requested):
Left comment

ryancham715:
> * Validation in controller/reconciler is missing for Tenant and ExternalModel

These two CRs are different in some ways ATM:
- Tenant is singleton. Users cannot create new Tenants. The worst thing they can do is to delete `spec` from `default-tenant` (compared to creating a bunch of no-spec Tenants).
- Tenant doesn't have mandatory fields. No-spec Tenant works just fine. Mandating `spec` in Tenants is more for the future multi-tenant vision. I don't think disturbing users before that is necessary.
- ExternalModel does not report status (and is not GA yet?). It feels more appropriate to implement this later as we develop this WIP.
> * This is going in post-GA .  We need to be  concerned about backward compatibility and migration.   Can you find a way to migrate and/or document what to look for/how to recover?

There's no force migration. I believe the dedicated reason `InvalidSpec` and message `spec is required` is self-explanatory enough for backward compat.

ishitasequeira:
> > * Validation in controller/reconciler is missing for Tenant and ExternalModel
> 
> These two CRs are different in some ways ATM:
> 
> * Tenant is singleton. Users cannot create new Tenants. The worst thing they can do is to delete `spec` from `default-tenant` (compared to creating a bunch of no-spec Tenants).
> * Tenant doesn't have mandatory fields. No-spec Tenant works just fine. Mandating `spec` in Tenants is more for the future multi-tenant vision. I don't think disturbing users before that is necessary.

Looking at the diff and the discussion here, I think `Tenant` and `ExternalModel` should be excluded from the `required: [spec]` and `omitempty` removal changes.

@ryancham715, you mentioned that "Tenant doesn't have mandatory fields. No-spec Tenant works just fine" and that mandating spec is more for the future multi-tenant vision. I agree with that, but the code in this PR still adds the CRD constraint and drops `omitempty` for Tenant. Feels like the intent and the code are out of sync.

Since every field in `TenantSpec` is optional, the CRD `required: [spec]` constraint only ensures the key is present, `spec: {}` still passes validation. The controller already handles empty spec fine via `applyGatewayDefaults`. So the constraint doesn't really buy us anything today. I'd say let's land it when multi-tenancy actually introduces required fields in TenantSpec.

> ExternalModel does not report status (and is not GA yet?). It feels more appropriate to implement this later as we develop this WIP.

Same story for `ExternalModel`, the CRD constraint adds no practical enforcement as of now.

The `MaaSSubscription`, `MaaSAuthPolicy`, and `MaaSModelRef` changes look good to me. The empty spec is genuinely invalid for those, and the `PhaseInvalid` + `reflect.DeepEqual` approach gives a clean upgrade path for legacy objects.

ishitasequeira (approved):
/lgtm

jrhyness (approved):
lgtm

PR #827, opened 2026-04-30
docs: update and deduplicate contents in tls-configuration.md

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-59748

## Description

1. `quickstart.md` is basically the use of `deploy.sh`, which includes ODH/RHOAI, RHCL installation and TLS config. Deleting them from that page.
2. Deleting unnecessary details such as volume mounts in `tls-configuration.md`.
3. Removing redundant contents and keeping only one copy of the following topics:
    - Running standalone script `setup-authorino-tls.sh`;
    - Annotations in Gateway to trust Authorino certificates; and
    - TLS config verification.

## Merge criteria:

## Summary by CodeRabbit

* **Documentation**
  * Consolidated and simplified TLS setup into a streamlined run/setup flow and renamed the manual Gateway→Authorino listener-TLS procedure
  * Clarified Gateway annotation requirement for Authorino TLS communication and noted it as temporary
  * Made verification steps namespace-agnostic and now point to an external validation guide
  * Simplified maas-api TLS docs by removing example env values, volume-mount details, and pruning kustomize overlays

### DIFF
--- a/docs/content/quickstart.md
+++ b/docs/content/quickstart.md
@@ -9,12 +9,6 @@ This guide provides quickstart instructions for deploying the MaaS Platform infr
 
 - **OpenShift cluster** (4.19.9+) with kubectl/oc access
       - **Recommended** 16 vCPUs, 32GB RAM, 100GB storage
-- **ODH/RHOAI requirements**:
-      - RHOAI 3.0 +
-      - ODH 3.0 +
-- **RHCL requirements** (Note: This can be installed automatically by the script below):
-      - RHCL 1.2 +
-- **Authorino TLS**: Listener TLS must be enabled on Authorino (see [Configure Authorino TLS](#configure-authorino-tls))
 - **Cluster admin** or equivalent permissions
 - **Required tools**:
       - `oc` (OpenShift CLI)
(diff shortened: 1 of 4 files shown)


docs/content/quickstart.md
```diff
@@ -23,17 +17,6 @@ This guide provides quickstart instructions for deploying the MaaS Platform infr
       - `kustomize` (v5.7.0+)
       - `gsed` (GNU sed) - **macOS only**: `brew install gnu-sed`
 
-## Configure Authorino TLS
```
jland-redhat, line 26: So actually can we leave this section? Just because forgetting to set up TLS is the one major issue that keeps popping up in our example dpeloyments

jland-redhat (changes requested):

docs/content/quickstart.md
ryancham715: We definitely retain TLS in the detailed deployment `maas-setup.md`. However in `quickstart.md`, we basically tell the reader to run `deploy.sh`, in which TLS is defaulted. That's why I think this is noise here (but certainly not in `maas-setup.md`).

jland-redhat: That is a good point

jland-redhat (approved):
lgtm

PR #828, opened 2026-04-30
fix: route model access probes through gateway's internal service

mynhardtburger:
## Summary
- On disconnected clusters, `FilterModelsByAccess` probes fail because they connect to the gateway via its external URL, which resolves to public ELB IPs unreachable from workload pods (blocked by `AdminNetworkPolicy` or VPC routing). Every probe hangs for 15 seconds (the `accessCheckTimeout`) and returns an empty model list.
- At startup, the maas-api now resolves the gateway's cluster-internal service address by looking up Services labeled with `gateway.networking.k8s.io/gateway-name` in the configured gateway namespace.
- A custom `DialContext` on the HTTP Transport routes all probe TCP connections to this internal address, while the original URL is preserved unchanged — keeping TLS SNI and the Host header correct so Envoy routing and Authorino ext-authz work identically.

## Details
- **Custom DialContext** — rather than rewriting probe URLs (which breaks TLS SNI matching on Envoy's filter chain), the Transport intercepts TCP connections and redirects them to the internal service. The probe URL, TLS ServerName, and Host header all remain the original external hostname.
- **Service resolution** is bounded by the configured `ACCESS_CHECK_TIMEOUT_SECONDS` to prevent stalling startup.
- **Service validation** requires the Service to have an `ownerReference` with `kind: Gateway` matching the configured gateway name, and exactly one matching candidate. This prevents a rogue Service with the same label from receiving probe traffic containing auth headers.
- HTTPS is maintained end-to-end; `InsecureSkipVerify: true` (already present) handles the cert mismatch between the external hostname and the internal service.

## Test plan
- [x] `go build ./...` and `go vet ./...` pass
- [x] Existing unit tests pass (they use httptest.Server URLs with empty `gatewayInternalHost`, so `DialContext` is not installed)
- [x] Deploy updated maas-api on a disconnected cluster and verify `GET /v1/models` returns non-empty model list
- [x] Verify Authorino ext-authz still enforces auth on probes (unauthorized users see empty list)
- [x] Re-run maas-tests Jenkins job to confirm the 22 disconnected-cluster failures are resolved

## Release Notes

* **New Features**
  * Added gateway internal host resolution during startup to enable access probes to route through the gateway's internal network, improving internal communication pathways.

* **Refactor**
  * Updated system initialization and underlying logic to support gateway internal host configuration and routing.

* **Tests**
  * Updated test cases to accommodate gateway internal host parameter changes.

### DIFF
--- a/maas-api/cmd/main.go
+++ b/maas-api/cmd/main.go
@@ -175,7 +175,15 @@ func registerHandlers(ctx context.Context, log *logger.Logger, router *gin.Engin
 
 	subscriptionSelector := subscription.NewSelector(log, cluster.MaaSSubscriptionLister)
 
-	modelManager, err := models.NewManager(log, cfg.AccessCheckTimeoutSeconds)
+	resolveCtx, resolveCancel := context.WithTimeout(ctx, time.Duration(cfg.AccessCheckTimeoutSeconds)*time.Second)
+	gatewayInternalHost, err := config.ResolveGatewayInternalHost(resolveCtx, cluster.ClientSet, cfg.GatewayName, cfg.GatewayNamespace)
+	resolveCancel()
+	if err != nil {
+		return fmt.Errorf("failed to resolve gateway internal address: %w", err)
+	}
+	log.Info("Resolved gateway internal host for access probes", "host", gatewayInternalHost)
+
+	modelManager, err := models.NewManager(log, cfg.AccessCheckTimeoutSeconds, gatewayInternalHost)
 	if err != nil {
 		log.Fatal("Failed to create model manager", "error", err)
 	}
--- a/maas-api/internal/config/cluster_config.go
+++ b/maas-api/internal/config/cluster_config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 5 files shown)


maas-api/cmd/main.go
```diff
@@ -150,7 +150,15 @@ func registerHandlers(ctx context.Context, log *logger.Logger, router *gin.Engin
 
 	subscriptionSelector := subscription.NewSelector(log, cluster.MaaSSubscriptionLister)
 
-	modelManager, err := models.NewManager(log, cfg.AccessCheckTimeoutSeconds)
+	resolveCtx, resolveCancel := context.WithTimeout(ctx, 5*time.Second)
```
jland-redhat: Lets respect respect the `AccessCheckTimeoutSeconds` here

jland-redhat (changes requested):
Two follow-up questions here:

1. Do we want to make this optional? So that we do external by default but a model can specify it wants to use the internal url?

2. Does this discovery have any effect on the model discovery? Because I believe that is following the same path

maas-api/cmd/main.go
mynhardtburger: done.

maas-api/cmd/main.go
```diff
@@ -150,7 +150,15 @@ func registerHandlers(ctx context.Context, log *logger.Logger, router *gin.Engin
 
 	subscriptionSelector := subscription.NewSelector(log, cluster.MaaSSubscriptionLister)
 
-	modelManager, err := models.NewManager(log, cfg.AccessCheckTimeoutSeconds)
+	resolveCtx, resolveCancel := context.WithTimeout(ctx, time.Duration(cfg.AccessCheckTimeoutSeconds)*time.Second)
+	gatewayInternalHost, err := config.ResolveGatewayInternalHost(resolveCtx, cluster.ClientSet, cfg.GatewayName, cfg.GatewayNamespace)
+	resolveCancel()
```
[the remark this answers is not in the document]
mynhardtburger: @coderabbitai: The Config.Validate() already checks that AccessCheckTimeoutSeconds is positive.

https://github.com/opendatahub-io/models-as-a-service/blob/10a153843aa3ebd155736fedf5b054bd3ffe0288/maas-api/internal/config/config.go#L158-L160

mynhardtburger:
@jland-redhat 
> Two follow-up questions here:
> 
>     1. Do we want to make this optional? So that we do external by default but a model can specify it wants to use the internal url?

To me it makes sense that there is a preference for maas-api to use internal URLs for inter-pod communication. What advantage is there to using an external url other than potentially supporting mutli-cluster environments where the model is located on an other external cluster. I don't think that maas claims to support multi-cluster environments?
 
>     2. Does this discovery have any effect on the model discovery? Because I believe that is following the same path
- maas-controller watches the kube resources
- maas-api's model discovery uses a `NewDynamicSharedInformerFactory` which populates and maintains a cache of MaasModelRef CRs in the cluster via kube API. Meaning that it doesn't make an external call. Only FilterModelsByAccess is impacted by this PR.

jland-redhat (dismissed):
lgtm

jland-redhat (approved):
lgtm

PR #830, opened 2026-04-30
feat(observability): add Perses resources to Tenant finalizer cleanup

ishitasequeira:
## Description

Add `PersesDashboard` and `PersesDatasource` to the Tenant finalizer's cleanup path so that Perses observability resources are removed when the MaaS Tenant CR is deleted.

Perses resources are rendered into the `opendatahub` namespace while the Tenant CR lives in `models-as-a-service`. Because Kubernetes forbids cross-namespace ownerReferences, `ApplyRendered` applies tracking labels (`maas.opendatahub.io/tenant-name`, `maas.opendatahub.io/tenant-namespace`) instead. The finalizer already uses these labels to clean up other cross-namespace extension resources (Kuadrant policies, Istio DestinationRules, etc.) — this PR extends that list to include the Perses GVKs.

Context: @ahadas flagged this gap in https://github.com/opendatahub-io/models-as-a-service/pull/805#pullrequestreview-4181758593.

https://redhat.atlassian.net/browse/RHOAIENG-60373

## How Has This Been Tested?

**Unit tests:**
- New test `TestTenantReconcile_DeletionCleansUpPersesResourcesByTrackingLabel` creates `PersesDashboard` and `PersesDatasource` unstructured objects with tracking labels in the `AppNamespace` (cross-namespace from the Tenant CR), triggers Tenant deletion, and asserts both resources are removed by the finalizer.
- All existing tests continue to pass (`go test ./... -count=1`).

**Manual cluster test (ROSA):**
1. Installed MaaS CRDs, stub Perses CRDs, and RBAC on a fresh cluster.
2. Created a Tenant CR in `models-as-a-service` and Perses resources with tracking labels in `opendatahub`.
3. Ran the controller locally, deleted the Tenant CR.
4. Verified both `PersesDashboard` and `PersesDatasource` were cleaned up (`No resources found in opendatahub namespace.`).

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Tenant deletion now includes proper cleanup of Perses Dashboard and Datasource resources, ensuring complete removal of tenant-owned instances across the system.

* **Tests**
  * Added test coverage to verify that Perses Dashboard and Datasource resources are properly cleaned up when tenants are deleted and finalization completes.

### DIFF
--- a/maas-controller/pkg/controller/maas/tenant_finalize.go
+++ b/maas-controller/pkg/controller/maas/tenant_finalize.go
@@ -52,6 +52,8 @@ var optionalPlatformGVKs = []schema.GroupVersionKind{
 	{Group: "networking.istio.io", Version: "v1", Kind: "DestinationRule"},
 	{Group: "networking.istio.io", Version: "v1alpha3", Kind: "EnvoyFilter"},
 	{Group: "telemetry.istio.io", Version: "v1", Kind: "Telemetry"},
+	tenantreconcile.GVKPersesDashboard,
+	tenantreconcile.GVKPersesDatasource,
 }
 
 func (r *TenantReconciler) operatorNamespace() string {
--- a/maas-controller/pkg/controller/maas/tenant_reconcile_test.go
+++ b/maas-controller/pkg/controller/maas/tenant_reconcile_test.go
@@ -10,6 +10,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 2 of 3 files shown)


yu-teo:
@ishitasequeira, took me some time to go over the JIRA issue and all of its requirements, but:
As per jira ticket, it's required to: ``` each resource's metadata.ownerReferences includes the MaaS Tenant CR (with controller: true and appropriate blockOwnerDeletion)```
The PR does not add `ownerReferences`. I see, it adds `PersesGVK`'s to optionalPlatformGVKs for finalizer-based manual deletion. The existing `ApplyRendered` logic sets `ownerReferences` only when `childNs == tenant.Namespace`. You touched on it, but if Perses resources live in `opendatahub` while Tenant lives in `models-as-a-service`, will it not fall into the `else` branch and get `tracking` labels, not `ownerReferences`?

Also, you chose to rely on finalizers because that is how we are used to doing the cleanup everywhere else, right? I am bringing it up because in cases when the controller is down and `Tenant` is deleted, the cleanup would stall until the controller restarts. Is that okay or will it come back as another bug to us?

ishitasequeira:
@yu-teo Great questions!! Let me respond for each of those individually.

> @ishitasequeira, took me some time to go over the JIRA issue and all of its requirements, but: As per jira ticket, it's required to: ` each resource's metadata.ownerReferences includes the MaaS Tenant CR (with controller: true and appropriate blockOwnerDeletion)` The PR does not add `ownerReferences`. I see, it adds `PersesGVK`'s to optionalPlatformGVKs for finalizer-based manual deletion. The existing `ApplyRendered` logic sets `ownerReferences` only when `childNs == tenant.Namespace`. You touched on it, but if Perses resources live in `opendatahub` while Tenant lives in `models-as-a-service`, will it not fall into the `else` branch and get `tracking` labels, not `ownerReferences`?

You're right!! the `Perses` resources live in `opendatahub` namespace while the `Tenant CR` lives in `models-as-a-service`, so ApplyRendered falls into the else branch and sets tracking labels, not ownerReferences. We are correctly using the tracking labels in the PR because of the cross-namespace restriction.

That being said, good catch on the JIRA mismatch. The acceptance criteria currently call for ownerReferences, which was created based on the older flow of ODH operator. I'll update the ticket to reflect the `tracking-label + finalizer` approach so it matches what's actually implementable and right.


> Also, you chose to rely on finalizers because that is how we are used to doing the cleanup everywhere else, right? I am bringing it up because in cases when the controller is down and `Tenant` is deleted, the cleanup would stall until the controller restarts. Is that okay or will it come back as another bug to us?

Yes, that's expected behavior. If the `maas-controller` is down, there's no one present to manage these resources. The Tenant delete will be blocked until the controller comes back and same logic would apply for every resource the Tenant owns.

yu-teo (approved):
lgtm

PR #859, opened 2026-05-05
fix: grant maas-controller RBAC permission for PodMonitor resources

EgorLu:
## Summary

- The maas-controller service account is missing RBAC permission to create/patch `podmonitors` in the `monitoring.coreos.com` API group
- This causes the ODH operator to fail when applying the `maas-api-metrics` PodMonitor added in #813
- Adds `podmonitors` to the maas-controller ClusterRole with the same verbs as other managed resources (create, delete, get, list, patch, watch)

## Test plan

- [ ] Verify ODH operator e2e no longer fails with `podmonitors is forbidden` error
- [ ] Verify `maas-api-metrics` PodMonitor is created successfully in the target namespace

## Summary by CodeRabbit

* **Chores**
  * Simplified monitoring configuration by streamlining endpoint selection.
  * Extended controller permissions to manage monitoring resources, enabling create, delete, retrieve, list, update, and watch operations.

### DIFF
--- a/deployment/base/maas-api/monitoring/podmonitor.yaml
+++ b/deployment/base/maas-api/monitoring/podmonitor.yaml
@@ -8,9 +8,6 @@ metadata:
     app.kubernetes.io/component: monitoring
     monitoring.opendatahub.io/scrape: "true"
 spec:
-  namespaceSelector:
-    matchLabels:
-      opendatahub.io/dashboard: "true"
   selector:
     matchLabels:
       app.kubernetes.io/name: maas-api
--- a/deployment/base/maas-controller/rbac/clusterrole.yaml
+++ b/deployment/base/maas-controller/rbac/clusterrole.yaml
@@ -242,6 +242,17 @@ rules:
   - update
   - watch
 - apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - podmonitors
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
   - networking.k8s.io
   resources:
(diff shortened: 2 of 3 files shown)


EgorLu:
@davidebianchi 
As I understand, after merging this PR to `main`, we would need to cherrypick both my previous PR with the `PodMonitor` (#813) and this PR to `stable`. Correct?
And then update the commit sha in `get_all_manifests.sh`.

davidebianchi:
Yes, the commit sha in the `get_all_manifests.sh` will be automatically updated by an operator github action once e2e run

jland-redhat (changes requested):
Can we identify the code that allows our E2E test to pass without this change and remove it?

EgorLu:
> Can we identify the code that allows our E2E test to pass without this change and remove it?

@jland-redhat The e2e tests filter out `PodMonitors` and `ServiceMonitors` before applying manifests, because the Prometheus Operator CRDs aren't installed in the e2e environment. Since the PodMonitor never got created, we never tested for missing RBAC.

I suggest that we fix that in a separate PR. WDYT?

mynhardtburger:
The RBAC markers need to be recorded within the `.go` files. That is what drives `make verify-codegen`.
See https://book.kubebuilder.io/reference/controller-gen.html

jland-redhat (approved):
lgtm

carlkyrillos:
@EgorLu @jland-redhat Any update on when this change will be synced to the `stable` branch? We need this change in `stable` to update `get_all_manifests.sh` and unblock the opendatahub-operator E2E tests.

PR #864, opened 2026-05-05
refactor: move Tenant reconciler away from params.env and maas-parameters

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-59864

## Description

**Major**

This PR consolidates the ownership of `params.env` and ConfigMap `maas-parameters`. After the change, only ODH reads from `params.env` and writes to `maas-parameters`, MaaS doesn't any more. Instead, Tenant reconciler gathers required params from MaaS controller args/env vars, Tenant CR and auto-detection, and patches them to the manifests with code instead of kustomize.

**Minor**

This PR also includes some other chores:
- starts to move `params.env` from outdated `/overlays/odh` to `base/maas-controller/default`, and includes only the necessary params there (the image ones).
- cleans up useless `shared-patches` and MaaS-side `params.env`.
- updates a whole bunch of stale or dead docs, scripts and comments.

## How Has This Been Tested?

Deployed on a cluster and verified all resources work as before and pass all E2E tests.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Platform values (images, gateway, namespace, audience, API key expiry) are now injected during post-render, eliminating previous file-based parameterization.
  * Image defaults for maas-api, maas-controller, payload-processing, and maas-api-key-cleanup are explicitly configurable.
  * Deployment/installer now manages a ConfigMap for image overrides instead of env-var patching.
  * Reduced unnecessary reconciliations previously triggered by deleted ConfigMaps.

* **Improvements**
  * Better handling and readiness processing for additional resource types (HTTPRoute, CronJob).


scripts/deploy.sh
```diff
@@ -545,17 +545,31 @@ main() {
...
+  local cm_payload_processing_image="quay.io/opendatahub/odh-ai-gateway-payload-processing:odh-stable"
+  local cm_cleanup_image="registry.redhat.io/ubi9/ubi-minimal:9.7"
+
+  log_info "  Ensuring maas-parameters ConfigMap..."
+  kubectl create configmap maas-parameters -n "$NAMESPACE" \
+    --from-literal="maas-api-image=${cm_maas_api_image}" \
+    --from-literal="maas-controller-image=${cm_maas_controller_image}" \
+    --from-literal="payload-processing-image=${cm_payload_processing_image}" \
+    --from-literal="maas-api-key-cleanup-image=${cm_cleanup_image}" \
+    --dry-run=client -o yaml | kubectl apply -f - || {
```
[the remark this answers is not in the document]
ryancham715: This is not a shell injection issue. In Bash, expanding a variable inside double quotes does not re-evaluate its contents as command substitution, so a value like MAAS_API_IMAGE='$(malicious command)' is passed to kubectl as a literal string, not executed. Command execution would only happen if the caller's shell already evaluated $(...) before invoking the script, which is not a vulnerability in this code path. There is no eval, bash -c, or other secondary interpretation here. If we want extra hardening, the relevant improvement would be image-reference validation, not shell-injection mitigation.

jrhyness:
You could add some brief documentation about the new ownership model.

ryancham715:
> You could add some brief documentation about the new ownership model.

Let me wait for @jland-redhat or @ishitasequeira's opinion on this. Honestly I think this is an internal process and probably shouldn't concern the users.

maas-controller/pkg/platform/tenantreconcile/params.go
```diff
@@ -0,0 +1,418 @@
...
+			r.SetNamespace(params.GatewayNamespace)
+		case gvk.Kind == "ServiceAccount" && name == "payload-processing":
+			r.SetNamespace(params.GatewayNamespace)
+		case gvk.Kind == "ConfigMap" && name == "payload-processing-plugins":
+			r.SetNamespace(params.GatewayNamespace)
+		case gvk.Kind == "ClusterRoleBinding" && name == "payload-processing-reader":
+			if err := patchClusterRoleBindingSubjectNS(r, params.GatewayNamespace); err != nil {
+				return err
+			}
+		}
```
ishitasequeira: Can we move this hard-coded strings to constants?

maas-controller/pkg/platform/tenantreconcile/params.go
```diff
@@ -0,0 +1,418 @@
...
+		return errors.New("HTTPRoute parentRefs[0] is not an object")
+	}
+	ref["namespace"] = params.GatewayNamespace
+	ref["name"] = params.GatewayName
+	parentRefs[0] = ref
+	if err := unstructured.SetNestedSlice(r.Object, parentRefs, "spec", "parentRefs"); err != nil {
+		return fmt.Errorf("write HTTPRoute parentRefs: %w", err)
+	}
+	return nil
+}
```
ishitasequeira, line 168: Every other patch function either skips on missing data or logs a warning. Only this one and `patchPayloadProcessingEnvoyFilter` hard-error, which means a malformed or variant HTTPRoute breaks the entire Tenant reconcile. Is that intentional?

scripts/deploy.sh
```diff
@@ -527,60 +510,62 @@ main() {
...
+    log_error "Failed to create/update maas-parameters ConfigMap"
+    return 1
+  }
```
ishitasequeira: The controller reads `RELATED_IMAGE_ODH_MAAS_API_IMAGE` from `configMapKeyRef` at pod startup. If you re-run `deploy.sh` with only `MAAS_API_IMAGE` changed, the ConfigMap is updated but the controller pod is not restarted, so the running pod still has the old image value in its env. A `kubectl rollout restart` is needed if the ConfigMap data actually changed.

scripts/deploy.sh
```diff
@@ -527,60 +510,62 @@ main() {
...
+  # Custom images from CLI flags override the defaults; other images keep defaults.
+  local cm_maas_api_image="${MAAS_API_IMAGE:-quay.io/opendatahub/maas-api:latest}"
+  local cm_maas_controller_image="${MAAS_CONTROLLER_IMAGE:-quay.io/opendatahub/maas-controller:latest}"
+  local cm_payload_processing_image="quay.io/opendatahub/odh-ai-gateway-payload-processing:odh-stable"
+  local cm_cleanup_image="registry.redhat.io/ubi9/ubi-minimal:9.7"
+
+  log_info "  Ensuring maas-parameters ConfigMap..."
+  kubectl create configmap maas-parameters -n "$NAMESPACE" \
+    --from-literal="maas-api-image=${cm_maas_api_image}" \
+    --from-literal="maas-controller-image=${cm_maas_controller_image}" \
```
ishitasequeira: Do we need the controller image to be set in configmap anymore as we are removing it from `manager.yaml`?

scripts/deploy.sh
```diff
@@ -527,60 +510,62 @@ main() {
...
+    if [[ "$cm_maas_controller_image" != "quay.io/opendatahub/maas-controller:latest" ]]; then
+      log_info "  Installing maas-controller with image: $cm_maas_controller_image"
+    fi
     if [[ "$NAMESPACE" != "opendatahub" ]]; then
       (cd "$project_root" && kustomize build deployment/base/maas-controller/default | \
-        sed "s/namespace: opendatahub/namespace: $NAMESPACE/g") | kubectl apply -f - || {
+        sed -e "s/namespace: opendatahub/namespace: $NAMESPACE/g" \
+            -e "s|image: quay.io/opendatahub/maas-controller:latest|image: $cm_maas_controller_image|g") | kubectl apply -f - || {
         log_error "Failed to apply maas-controller manifests"
         return 1
```
ishitasequeira: If the kustomize base image tag ever changes (e.g. `latest` → a version tag), the sed silently becomes a no-op and the wrong image is deployed with no error. Use `kustomize edit set` image in a temp overlay instead.

maas-controller/pkg/platform/tenantreconcile/params.go
ishitasequeira, line 1: Can we add some tests for this file?

scripts/deploy.sh
ryancham715: Technically no, we don't need it, but the operator is providing that, so I simulated the same behavior here. Ideally the operator can exclude controller image too. I just don't think it's worth the effort changing that.

maas-controller/pkg/platform/tenantreconcile/params.go
ryancham715: I'm not sure I fully understand you here:
- If you're referring to missing params, I now centralize all defaulting behavior in `BuildPlatformParams` for better readability.
- If you're referring to missing fields in manifests, my idea is to hard-error whenever manifest structures are not as expected. We should ensure the code matches the manifests, since we control both side.

maas-controller/pkg/platform/tenantreconcile/params.go
ryancham715: Sure, do you think it's okay to use the real manifests for testing?

maas-controller/pkg/platform/tenantreconcile/params.go
ishitasequeira: I was curious on why fail on missing data in this function, but not other places. But I am good if we have validated the changes correctly.

maas-controller/pkg/platform/tenantreconcile/params.go
ishitasequeira: We can use the examples from our e2e tests as needed.

maas-controller/pkg/platform/tenantreconcile/params.go
ryancham715: I think it's because we patch based on the assumption of a specific manifest structure. If the manifest is not in that structure (although it's not likely to happen), the behavior can be unexpected. All in all it's AI's defensive fail-fast programming, which looks fine to me in this case.

ishitasequeira (approved):
lgtm

PR #865, opened 2026-05-05
fix: make MaaSModelRef ExternalModel handler API-group agnostic

### LINKED ISSUE #856: feat: MaaSAuthPolicy/MaaSSubscription support for inference.opendatahub.io ExternalModels
## Context

The new `inference.opendatahub.io/v1alpha1` ExternalModel CRDs (being added in [ai-gateway-payload-processing#183](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/183) / [#184](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/184)) create HTTPRoutes via the BBR ExternalModel controller. However, the MaaS controller currently only creates AuthPolicies and TokenRateLimitPolicies for `maas.opendatahub.io` ExternalModels.

This means models deployed via the new CRDs have no Kuadrant auth protection — the gateway-level default-deny blocks them entirely, and there's no per-route AuthPolicy to enable authenticated access.

See: [ai-gateway-payload-processing#235](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/235) for the original discussion. Nir confirmed this is MaaS scope.

## What needs to change

The `MaaSAuthPolicy` and `MaaSSubscription` reconcilers need to support `inference.opendatahub.io` ExternalModels in addition to `maas.opendatahub.io` ones. This likely means:

1. `MaaSModelRef` should be able to reference ExternalModels from both API groups
2. The reconcilers that generate Kuadrant AuthPolicy and TokenRateLimitPolicy should work for HTTPRoutes created by either the MaaS ExternalModel reconciler or the BBR ExternalModel controller

The API group change should be straightforward — the generated Kuadrant resources target HTTPRoutes by name, and the HTTPRoute name matches the ExternalModel name regardless of which reconciler created it.

## Acceptance criteria

- Models deployed via `inference.opendatahub.io` ExternalModel CRDs can be protected by MaaSAuthPolicy
- Models deployed via `inference.opendatahub.io` ExternalModel CRDs can have rate limits via MaaSSubscription
- Existing `maas.opendatahub.io` ExternalModel auth continues to work unchanged

cc @nirrozenbaum @jland-redhat

yossiovadia:
## Summary

- Remove the `maas.opendatahub.io` ExternalModel CR read from `externalModelHandler.ReconcileRoute` — it was only used for a validation check and a log line
- The handler now only checks for the HTTPRoute (which is the actual resource that AuthPolicy/TRLP target)
- This makes MaaSModelRef work with ExternalModels from any API group, as long as an HTTPRoute with the matching name exists

## Why

The new `inference.opendatahub.io/v1alpha1` ExternalProvider/ExternalModel CRDs ([ai-gateway-payload-processing#183](https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/183), [#184](https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/184)) create HTTPRoutes via the BBR ExternalModel controller. Creating a `MaaSModelRef` pointing at these models failed with:

```
ExternalModel new-katan-openai not found in namespace llm
```

The handler tried to read the `maas.opendatahub.io` ExternalModel CR, which doesn't exist — the model is in a different API group. But the HTTPRoute (created by the BBR reconciler) does exist with the correct name, and that's all the auth/rate-limit chain needs.

## What changed

**`providers_external.go`**: Removed the `h.r.Get(ctx, externalModelKey, externalModel)` call and the log line referencing `externalModel.Name`/`externalModel.Spec.Provider`. Removed unused `types` import.

**`providers_external_test.go`**: Updated `TestExternalModel_ReconcileRoute_MissingExternalModel` — now expects success (waiting state) instead of error when ExternalModel CR is missing, since only the HTTPRoute matters.

## Test plan

- [x] All existing unit tests pass (`make -C maas-controller test`)
- [x] Verified on Kind cluster (maas-experimental):
  - Created `MaaSModelRef` pointing at `inference.opendatahub.io` ExternalModel `new-katan-openai` → **Phase: Ready**
  - Created `MaaSAuthPolicy` + `MaaSSubscription` → Kuadrant AuthPolicy + TRLP created
  - Applied gateway-level default-deny AuthPolicy
  - No auth → **401**, invalid key → **401**, valid API key → **200**
  - All existing `maas.opendatahub.io` MaaSModelRefs still **Ready**

Closes #856

cc @jland-redhat @nirrozenbaum

## Summary by CodeRabbit

* **Refactor**
  * Simplified route validation logic to reduce external dependencies and improve performance.
  * Enhanced system resilience by gracefully handling missing references during route reconciliation.
  * Updated related tests to reflect new validation behavior.

### DIFF
--- a/maas-controller/pkg/controller/maas/providers_external.go
+++ b/maas-controller/pkg/controller/maas/providers_external.go
@@ -23,7 +23,6 @@ import (
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
@@ -41,22 +40,12 @@ type externalModelHandler struct {
 }
 
 // ReconcileRoute validates the HTTPRoute for an external model and populates status.
-// The ExternalModel reconciler creates the "maas-model-<model.Name>" HTTPRoute in the
-// model's namespace. This method validates that it exists and is accepted by the gateway.
+// The ExternalModel reconciler creates the HTTPRoute in the model's namespace.
+// This method validates that it exists and is accepted by the gateway.
+// Note: we don't read the ExternalModel CR itself — only the HTTPRoute matters.
+// This keeps MaaSModelRef agnostic to the ExternalModel API group
+// (maas.opendatahub.io or inference.opendatahub.io).
 func (h *externalModelHandler) ReconcileRoute(ctx context.Context, log logr.Logger, model *maasv1alpha1.MaaSModelRef) error {
-	// Fetch the referenced ExternalModel CR to get provider configuration
-	externalModel := &maasv1alpha1.ExternalModel{}
-	externalModelKey := types.NamespacedName{
-		Name:      model.Spec.ModelRef.Name,
-		Namespace: model.Namespace,
-	}
-	if err := h.r.Get(ctx, externalModelKey, externalModel); err != nil {
-		if apierrors.IsNotFound(err) {
-			return fmt.Errorf("ExternalModel %s not found in namespace %s", model.Spec.ModelRef.Name, model.Namespace)
-		}
-		return fmt.Errorf("failed to get ExternalModel %s: %w", model.Spec.ModelRef.Name, err)
-	}
-
 	routeName := model.Spec.ModelRef.Name
 	routeNS := model.Namespace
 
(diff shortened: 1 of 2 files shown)


yossiovadia:
@coderabbitai review

jland-redhat (changes requested):
So I am not sure we want to make this agnostic for a few reasons.

But the biggest is that if we need to modify the behavior of the way we handle External Models in some way it could cause us issues.

I would rather us explicitly detect and handle each of these for now (even if "handling" means reusing existing fuctions.)

yossiovadia:
Hey Jamie, no problems. indeed explicit handling per API group is safer for future divergence. Happy to leave it to you to implement the way that fits best with the MaaS controller patterns. The approach works either way — the key insight is that the handler only needs the HTTPRoute (by name), not the ExternalModel CR itself. Feel free to take this in whatever direction you prefer.

PR #866, opened 2026-05-06
fix: add gateway-level default-deny AuthPolicy to local-deploy

yossiovadia:
## Summary

- Add `gateway-default-auth` AuthPolicy to `local-deploy.sh` — blocks all model inference routes that don't have a per-route AuthPolicy

## Why

The production MaaS deployment already has this policy (`deployment/base/maas-controller/policies/gateway-default-auth.yaml`), but local-deploy was missing it. Without it, models deployed via the new `inference.opendatahub.io` CRDs are completely unprotected — requests bypass Authorino entirely, and any request (no auth, invalid auth) returns HTTP 200.

Found during testing of ExternalProvider/ExternalModel CRDs ([ai-gateway-payload-processing#183](https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/183)). The default-deny is the same pattern used in production — routes with `MaaSAuthPolicy` override the defaults and allow authenticated access.

## What changed

`test/e2e/scripts/local-deploy.sh` — added a block after subscription/auth policy deployment that applies the gateway-default-auth AuthPolicy (idempotent, skips if already exists).

## Test plan

- [x] Verified on Kind cluster (maas-experimental):
  - Without default-deny: new CRD routes return HTTP 200 with no auth
  - With default-deny: new CRD routes return HTTP 401/403 (blocked)
  - Old MaaS routes (with per-route AuthPolicy) continue to work normally
  - Valid API key on new CRD routes returns HTTP 200 (when MaaSModelRef + AuthPolicy are configured)

## Summary by CodeRabbit

* **Tests**
  * Added a default-deny authorization policy for model inference routes in the test deployment.
  * Added a wait/check to ensure the policy is accepted and enforced, with timeout handling.
  * Updated smoke tests to validate responses allowing 401/403 or 200 depending on per-route authorization.


yossiovadia:
@coderabbitai review

jland-redhat:
The main auth policy has been moved to the gateway level which is captured by local so going to go ahead and close this PR

PR #868, opened 2026-05-06
fix: maasmodelref endpoint selection in setups with multiple gateways available

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-52336

## Description
In `getEndpointFromLLMISvc`, when multiple gateways exist, `llmisvc.Status.Addresses` contains gateway-external entries from **ALL** gateways. The method picks the first HTTPS available, which may belong to the wrong gateway (e.g., `openshift-ai-inference` instead of `maas-default-gateway`).

The proposed fix:
- In `providers_llmisvc.go` — `getEndpointFromLLMISvc` now accepts `expectedHostnames`. When non-empty, it filters gateway-external addresses to only those whose hostname matches, and skips fallback to unfiltered addresses (which may belong to the wrong gateway, how we do it now, essentially). When empty, old behavior is preserved for single-gateway deployments.
- Additionally, added 7 test cases covering `providers_llmisvc.go`. Testing the following scenarios: 
   1) correct hostname selected from multiple gateways
   2) no match returns empty
   3) legacy behavior without hostnames
   4) single gateway with hostnames
   5) fallback to first address with old logic 
   6) no fallback when filtering 
   7) HTTPS preference

## How Has This Been Tested?
Added unit tests to make the changes work as intended.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Improved LLM Inference Service endpoint selection to prefer gateway addresses matching expected hostnames and better handle multi-gateway setups.

* **Bug Fixes**
  * More reliable endpoint fallback behavior when no matching gateway address is found, reducing incorrect endpoint resolution.

* **Tests**
  * Added unit tests covering hostname selection, TLS preference, case-insensitive matching, and legacy fallback scenarios.


yu-teo:
@coderabbitai 
Do you agree with the following fixes addressign security and code quality fixes?
1. CWE-129 (slice indexing): Already safe — gatewayExternalURLs[0] is guarded by len(...) > 0, and  lmisvc.Status.Addresses[0].URL is guarded by len(...) > 0 && ... != nil. No change needed.
2. CWE-248 (URL parse error handling): Added host == "" check after parsed.Hostname() — addresses with no extractable hostname are now skipped instead of silently matching nothing and falling through. New test TestGetEndpointFromLLMISvc_EmptyHostnameSkipped covers this.
3. CWE-200 (case-sensitive hostname): Both the hostSet keys and the parsed hostname are now lowercased via strings.ToLower() per RFC 4343. New test TestGetEndpointFromLLMISvc_CaseInsensitiveHostname covers this.
4. CWE-476 (empty endpoint contract): The doc comment now explicitly documents that "" is returned when no match is found and that Status() falls through to GetModelEndpoint. The caller already validates this at line 127.

liangwen12year (approved):
lgtm.

somya-bhatnagar:
Looks Good. Just two small comments :

1. Please correct two typos in PR description:
“odl” → “old”, “Tetsing” → “Testing” 

2. Adding a small test for Name == nil / URL == nil on Addressable would lock in the existing guards if someone refactors the loop. ( CodeRabbit Nit, I think looks valid here)

somya-bhatnagar (approved):
lgtm

PR #892, opened 2026-05-11
fix: add timeout to cleanup finalizer to prevent stuck namespace deletion

dmytro-zaharnytskyi:
The `LifecycleReconciler` adds `maas.opendatahub.io/cleanup` finalizer to the maas-controller Deployment and removes it after cleaning up Tenants and cluster-scoped resources. When a **namespace** is deleted (e.g., during ODH e2e test teardown), the controller pod is terminated before cleanup completes, leaving the finalizer stuck. The Deployment can't be garbage-collected, and the namespace stays in `Terminating` indefinitely.

This is currently blocking all PRs from merging to `opendatahub-operator` — the RHOAI e2e test fails on every run with:
```
NamespaceFinalizersRemaining: Some content in the namespace has finalizers
remaining: maas.opendatahub.io/cleanup in 1 resource instances
```

**Fix:** Add a 2-minute timeout (measured from `DeletionTimestamp`) to `runCleanup`. If Tenants are still pending after the timeout, force-remove the finalizer so the Deployment and namespace can be deleted. Cleanup errors are logged but no longer block finalizer removal.

**File:** `maas-controller/pkg/controller/maas/self_deployment_controller.go`

## Release Notes

* **Bug Fixes**
  * Introduced a 2-minute timeout for deployment cleanup to prevent indefinite waiting during teardown operations.
  * Enhanced error resilience during cleanup, allowing the process to continue gracefully despite failures instead of halting entirely.

### DIFF
--- a/maas-controller/pkg/controller/maas/self_deployment_controller.go
+++ b/maas-controller/pkg/controller/maas/self_deployment_controller.go
@@ -109,23 +109,35 @@ func (r *LifecycleReconciler) ensureFinalizer(ctx context.Context, log logr.Logg
 	return nil
 }
 
+// cleanupTimeout bounds the total time we wait for Tenants to terminate.
+// If exceeded, we force-remove the finalizer so the Deployment (and its
+// namespace) can be deleted. Without this, a namespace delete kills the
+// controller pod before cleanup finishes, leaving the finalizer stuck.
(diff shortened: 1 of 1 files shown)


jland-redhat:
This should me addressed by this pr:
https://github.com/opendatahub-io/opendatahub-operator/pull/3516

PR #894, opened 2026-05-12
fix: fixing cleanup issue where not all objects were removed

jland-redhat:
### Description

This change introduces a **cluster-scoped `Config`** resource (`maas.opendatahub.io/v1alpha1`, singleton name `default`) as the **anchor for Models-as-a-Service platform lifecycle and garbage collection**. Workloads and related objects applied by the **Tenant** reconciler (server-side apply) receive **`metadata.ownerReferences`** to that `Config` where appropriate, so deleting the anchor drives cleanup of the rendered platform graph instead of relying on legacy Tenant finalizers alone.

**Highlights**

- **New CRD + default instance**: `Config` CRD and `deployment/base/maas-controller/default/config-default.yaml` for `default`.
- **Bootstrap** (`cmd/manager`): Ensures `Config/default` exists, then ensures `default-tenant` exists with a **controller** owner reference from `Config` → `Tenant` on create; patches an **additional owner reference** on existing Tenants when needed (upgrade path). Uses the **API reader** after `Config` create to avoid cache-stale reads.
- **Tenant reconcile** (`tenant_reconcile.go`): Gates platform work on a live, non-terminating `Config` with UID; surfaces **WaitingForConfigUID** when needed. **Deletion path** deletes the `Config` anchor when present, then strips the legacy finalizer. **Proactive** removal of **`maas.opendatahub.io/tenant-finalizer`** on non-deleting reconciles so upgraded clusters converge without manual patches.
- **Platform apply** (`tenantreconcile/apply.go`): `SetControllerReference` from `Config` for rendered objects; **extensible skip list** (`configOwnerRefSkips`) for resources that must not be controller-owned by `Config` (e.g. **`maas-controller` Deployment**, **`maas-parameters` ConfigMap** so it can remain aligned with operator ownership). Handles **`AlreadyOwnedError`** without failing the whole apply.
- **Lifecycle reconciler** (`self_deployment_controller.go`): Non-controller **owner reference** from `maas-controller` **Deployment** → `Config`, strips legacy cleanup finalizer on the Deployment; aligns with moving teardown off the Tenant object.
- **RBAC**: Split so **`maas-controller-role`** keeps **read-only** `configs` rules; **`maas-controller-cluster-config-role`** (+ binding) holds **mutating** `configs` rules scoped to **`resourceNames: ["default"]`** where RBAC allows (separate rule for `create`).
- **Install / scripts**: `deploy.sh` + `deployment-helpers` apply **CRDs first** and wait for **Established** before applying the full controller bundle (avoids races when not using the ODH operator’s ordering). **`cleanup-odh.sh`**: clear **finalizers** on cluster anchor CRs (`Config`, legacy `ClusterTenant`) before delete.
- **Container build**: `container.mk` builds with **repo root** context and `-f maas-controller/Dockerfile` so `COPY maas-controller/...` and embedded manifests resolve correctly.

**Coordination with Open Data Hub operator**

Validated together with **[opendatahub-io/opendatahub-operator#3535](https://github.com/opendatahub-io/opendatahub-operator/pull/3535)**:

- Operator side moves MaaS disable/cleanup to **`maas-controller` Deployment** lifecycle (instead of deleting the **Tenant** CR), provisions/watches the **cluster-scoped MaaS config** surface, and adds **RBAC for `configs`** so the operator manager can list/watch the new API.
- MaaS side avoids fighting the operator on **`maas-parameters`** by **not** setting a Config **controller** owner ref on that ConfigMap (skip list), while still applying labels and SSA.

Merge **both** PRs in a coordinated release (or ensure operator RBAC lands before / with the controller build that assumes operator watches `Config`).

### How Has This Been Tested?

- `make -C maas-controller test` (unit tests, including Tenant / lifecycle / apply paths).
- `go build` for `maas-controller` (including `cmd/manager`).
- CI: **Konflux / group-test** run referenced on the PR (see bot artifact links on the PR conversation).
- Manual / cluster (recommended before merge): install controller bundle after CRD wait; confirm `config default` exists; confirm `tenant default-tenant` owner refs; delete `config default` and confirm platform resources garbage-collect; confirm **`maas-parameters`** has no Config **controller** owner ref if the operator still owns it; disable MaaS via operator with **#3535** and confirm **`maas-controller` Deployment** deletion path and cleanup behavior.

### Risk analysis

- **Risk rating**: **4**
- **Why**: Touches **CRD install order**, **RBAC**, **bootstrap**, **GC semantics**, and **cross-repo operator** behavior. E2E smoke in CI may not cover every disable/cleanup and upgrade permutation; wrong ordering or missing operator RBAC can cause **forbidden** list/watch on `Config` or stuck anchors. Mitigations: two-phase CRD apply in `deploy.sh`, dedicated Config ClusterRole with `resourceNames`, skip list for `maas-parameters`, API reader after create, and explicit operator PR for `configs` RBAC.

## Summary by CodeRabbit

* **New Features**
  * Added a cluster-scoped Config "default" to anchor controller behavior.
  * Controller startup now waits for CRDs and the Config anchor before proceeding.

* **Documentation**
  * Installation and deployment docs updated with two-phase install guidance and CRD readiness notes.

* **Refactor**
  * Simplified lifecycle and teardown behavior; removed legacy finalizer-driven cleanup flows.
  * RBAC rules reorganized to separate config permissions from other controller privileges.

* **Ops**
  * Shortened controller pod termination grace period for faster shutdowns.

https://github.com/user-attachments/assets/a1388507-56a1-4748-9e59-79d42d59faaa


maas-controller/cmd/manager/main.go
```diff
@@ -325,47 +326,100 @@ func getClusterServiceAccountIssuer(c client.Reader) (string, error) {
...
+						Kind:       maasv1alpha1.ConfigKind,
+					},
+					ObjectMeta: metav1.ObjectMeta{Name: maasv1alpha1.ConfigInstanceName},
+				}
+				if err := c.Create(ctx, toCreate); err != nil && !errors.IsAlreadyExists(err) {
+					log.Error(err, "failed to create Config")
+					return
+				}
+				if err := c.Get(ctx, ctKey, &ct); err != nil {
+					log.Error(err, "re-fetch Config after create")
```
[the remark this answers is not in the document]
jland-redhat: Makes sense

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -70,40 +73,52 @@ func (r *TenantReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			}
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, nil
 	}
 
 	ms := managementState(tenant.Annotations)
 	if ms == managementStateRemoved || ms == managementStateUnmanaged {
 		return r.handleIdleManagementState(ctx, &tenant, ms)
 	}
```
ishitasequeira: With the Config GC cascade coming in from ODH PR [#3535](https://github.com/opendatahub-io/opendatahub-operator/pull/3535), I think the `Removed` case here can be removed. When MaaS is disabled, ODH deletes `ModelsAsService` CR → GC cascades to Config → GC cascades to Tenant, and `readyConfigOrWait` already gates on Config's `DeletionTimestamp` correctly. So `TenantReconciler` doesn't need to act on Removed at all. It will idle naturally once Config starts terminating.

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -225,33 +248,121 @@ func (r *TenantReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-				return ctrl.Result{}, err
-			}
-			if pending {
-				return ctrl.Result{RequeueAfter: finalizeRequeueInterval}, nil
-			}
+	if ms == managementStateRemoved {
+		requeue, err := r.ensureConfigDeletedForRemoval(ctx)
+		if err != nil {
+			return ctrl.Result{}, err
 		}
```
ishitasequeira: This is the part I'm a bit worried about. `TenantReconciler` deleting `Config` here means a child is deleting its own owner, and it races with the GC cascade that's already doing the same thing. If both paths fire around the same time it can cause confusing intermediate states. I think `ensureConfigDeletedForRemoval` (lines 307–327) can be removed, and `handleIdleManagementState` simplified to only handle Unmanaged.

What do you think?

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -225,33 +248,121 @@ func (r *TenantReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 			return ctrl.Result{}, err
 		}
+		if requeue != nil {
+			return *requeue, nil
+		}
+		return ctrl.Result{}, nil
+	}
+	if err := r.stripLegacyTenantFinalizerIfPresent(ctx, tenant); err != nil {
+		return ctrl.Result{}, err
 	}
```
ishitasequeira: Given that we have just added the finalizer post 3.4 release, do we want this backward compatibility yet?

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -225,33 +248,121 @@ func (r *TenantReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+
+// SkipConfigBootstrap returns true when the default Tenant is in Removed management state.
+// The manager startup runnable must not recreate Config or patch owner refs in that state,
+// because Removed reconciliation deletes Config to drive platform GC.
+func SkipConfigBootstrap(tenant *maasv1alpha1.Tenant) bool {
+	if tenant == nil || tenant.Annotations == nil {
+		return false
+	}
+	return tenant.Annotations[managementStateAnnotation] == managementStateRemoved
+}
```
ishitasequeira: This function exists to prevent the bootstrap runnable from recreating Config during a `Removed` teardown cycle. But if we remove the `Removed` handling from `TenantReconciler` as suggested in my other comment, and instead let `LifecycleReconciler` check `Deployment.DeletionTimestamp` before creating Config, this guard becomes unnecessary. The Deployment being terminating is a much more reliable signal than an annotation on a Tenant that may not exist yet. I think this function can be removed alongside the other `Removed` cleanup.

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -81,135 +64,77 @@ func (r *LifecycleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	if err := r.List(ctx, crdList, componentLabel); err != nil {
-		return fmt.Errorf("list maas-controller CRDs: %w", err)
+	if !cfg.DeletionTimestamp.IsZero() {
+		return nil
 	}
-	if len(crdList.Items) == 0 {
-		log.Info("no CRDs matched component label; skipping deletion (operator may not have stamped the label)")
+	if cfg.UID == "" {
+		return nil
 	}
```
ishitasequeira, line 193: Both the `"Config not found"` and `"Config terminating"` cases silently `return nil`. This means if Config is accidentally deleted, the Deployment quietly loses its `ownerRef` and nothing recovers until a separate event triggers a reconcile.

What if `LifecycleReconciler` also owned `Config` creation? The guard logic would be straightforward:

* Config missing + Deployment not terminating → create Config (accidental deletion, self-heal)
* Config missing + Deployment has DeletionTimestamp → skip, do not recreate (intentional teardown)
* Config has DeletionTimestamp → requeue after a short delay and let GC propagate

This makes the reconciler fully self-healing for the accidental deletion case, avoids the `recreate-after-delete` race during intentional teardown, and removes the need for `SkipConfigBootstrap` in `main.go` altogether. 

Happy to discuss if there's a reason Config creation was kept in the bootstrap runnable rather than here.

maas-controller/cmd/manager/main.go
```diff
@@ -325,47 +326,103 @@ func getClusterServiceAccountIssuer(c client.Reader) (string, error) {
...
+				log.Info("ensured default-tenant exists", "namespace", tenantNamespace)
+			}
+
+			if tenantReferencesConfig(&tenant, &ct) {
+				return
+			}
+			base := tenant.DeepCopy()
+			if err := controllerutil.SetOwnerReference(&ct, &tenant, scheme); err != nil {
+				log.Error(err, "SetOwnerReference on existing Tenant")
 				return
```
ishitasequeira: Noticed a small inconsistency here.. New Tenants get `SetControllerReference (controller=true)` while existing Tenants missing the `ownerRef` (e.g. on upgrade from an older release) get `SetOwnerReference (non-controller)`. It's not a functional bug since the `tenantReferencesConfig` guard on line 413 prevents duplicates, but it's a bit surprising that the same relationship ends up with different ref types depending on the path.

One thing worth thinking about: since `LifecycleReconciler` already ensures the `Config → Deployment ownerRef` via `ensureDeploymentReferencesConfig` on every reconcile, it could do the same for Tenant. That would make the bootstrap-time patching on line 417 redundant and keep all ownerRef management in one place where it naturally self-heals if a ref goes missing. What do you think?

ishitasequeira (commented):
I like the overall direction of Config CR as an anchor. 

Leaving a few things to clean up and one possible improvement to ConfigReconciler that simplifies the whole ownership story.

liangwen12year:
Nit: `readyConfigOrWait` has a `ct.UID == ""` branch that patches status with `WaitingForConfigUID` and requeues after 5s, but there's no dedicated unit test for it. The existing tests cover `ConfigMissing` (Config not found) and `ConfigTerminating` (Config deleting), but not this edge case.

A simple test that creates a Config with `UID: ""` (or uses a fake client that doesn't auto-generate UIDs) and asserts the Tenant status gets `WaitingForConfigUID` + `RequeueAfter: 5s` would pin down the behavior.

maas-controller/pkg/controller/maas/tenant_reconcile.go
jland-redhat: Nope! Thought we added in 3.4 thanks for the heads up

ishitasequeira (dismissed):
/lgtm

ishitasequeira (approved):
/lgtm

PR #912, opened 2026-05-15
feat: move auth-policy to gateway level for multi-tenancy

ishitasequeira:
## Description

Moves MaaS authentication from **per-model HTTPRoute-scoped AuthPolicies** to a **single Gateway-scoped AuthPolicy**, reducing policy sprawl and enabling multi-tenant hostname isolation.

The work has two parts: controller code that auto-creates and maintains the gateway policy, and a `poc/` folder with reference artefacts from cluster testing.

[RHOAIENG-62571](https://redhat.atlassian.net/browse/RHOAIENG-62571)
---

### New: gateway-level AuthPolicy auto-creation

`reconcileGatewayAuthPolicy` creates/updates a singleton `maas-gateway-auth` Kuadrant AuthPolicy in the gateway namespace on every `MaaSAuthPolicy` reconciliation. It is built by `buildGatewayAuthPolicySpec` and contains:

- **API-key identity** — extracts the bearer token from `Authorization` header, resolves it against `maas-api` (`/internal/v1/api-keys/validate`), and caches the result with a configurable TTL.
- **Subscription lookup** — calls `maas-api` (`/internal/v1/subscriptions/select`) to find the active subscription for the (user, model) pair; result is cached with the same TTL.
- **`auth-valid` OPA rule** — allows if API key validated OR K8s token identity exists OR OIDC subject exists.
- **`subscription-valid` OPA rule** — allows if subscription is Active/Degraded, not deleted, and no error.
- **`tenant-gateway-isolation` stub** — always-allow placeholder; ready to be replaced with a real `maas-api` hostname check for multi-gateway Coke/Pepsi isolation.
- **Response header injection** — sets `X-MaaS-Subscription`, identity filter fields (`userId`, `username`, `groups`, `selected_subscription_key`) for downstream consumers.
- **CEL-based dynamic model identity** — extracts model name from `X-Gateway-Model-Name` header with `request.path` fallback; no static namespace/name baked into the gateway policy.

### Updated: per-model AuthPolicy now carries only group membership

`reconcileModelAuthPolicies` now emits a lightweight per-route AuthPolicy with **only** `require-group-membership`. All shared auth logic (key validation, subscription check, OPA) was moved to the gateway policy above.

---

## PoC reference artefacts (`poc/gateway-authpolicy/`)

| File | Purpose |
|------|---------|
| `01-gateway-authpolicy.yaml` | Human-readable reference for the full AuthPolicy spec the controller generates |
| `02-per-model-group-policy.yaml` | Reference for the per-model companion policy |

---

## How Has This Been Tested?

End-to-end tested on a live ROSA cluster with Kuadrant + Authorino + Istio.

Key results:
- ✅ Valid API key → 200 from model backend
- ✅ Invalid / missing API key → 401
- ✅ Expired or missing subscription → 403
- ✅ K8s SA token on `/v1/models` → 200
- ✅ Policy coexists with ODH-managed `gateway-auth-policy` (no conflict)
- ✅ Controller auto-creates `maas-gateway-auth` on reconcile

Notes on testing for reference are documented in: https://gist.github.com/ishitasequeira/e1fa5d3605151750ddd3decb3c9da6ab

---

## Merge criteria

## Summary by CodeRabbit

* **New Features**
  * Gateway namespace configuration and a singleton gateway-level auth policy that aggregates per-model allowlists
  * Dynamic model identity extraction for gateway-level authorization, subscription selection, and telemetry

* **Improvements**
  * Gateway-scoped cache key isolation and response shaping; gateway-only mode avoids generating per-model auth policies
  * E2E: gateway readiness gating with diagnostics; new gateway-related env defaults and timeout
  * Installer: non-fatal warning for unresolved image placeholder
  * Deployment check now targets gateway-level auth policy

* **Tests**
  * Updated and added unit/E2E tests to validate gateway-only aggregation, cache isolation, and adjusted waits/expectations


poc/gateway-authpolicy/01-gateway-authpolicy.yaml
```diff
@@ -0,0 +1,328 @@
...
+              }
+              # Kubernetes token authentication: check identity exists
+              allow {
+                object.get(input.auth.identity, "user", {}).username != ""
+              }
+              # OIDC token authentication: check JWT subject exists
+              allow {
+                object.get(input.auth.identity, "sub", "") != ""
+              }
+          cache:
```
jland-redhat: We would also need to include the path as part of the key for the cacheing

Also this is something we want to be super careful about when we do it for real. We need to REALLY make sure that folks can't trick the cache with subsequent request

poc/gateway-authpolicy/01-gateway-authpolicy.yaml
```diff
@@ -0,0 +1,328 @@
...
+
+        subscription-info:
+          priority: 1
+          metrics: false
+          http:
+            url: https://maas-api.MAAS_API_NAMESPACE.svc.cluster.local:8443/internal/v1/subscriptions/select
+            contentType: application/json
+            method: POST
+            body:
+              # CEL extracts namespace/model-name from path /<namespace>/<model-name>/...
```
jland-redhat: We are making the strong assumption that the path is in face Namespace then Model Name (and not really model name but the name of the LLMInferenceService, ExternalModel, or any future object we support) 


This may be fine especially since we eventually want to make this legacy and probably depreacate but would want to validate with @nirrozenbaum and @pierDipi

poc/gateway-authpolicy/01-gateway-authpolicy.yaml
```diff
@@ -0,0 +1,328 @@
...
+          metrics: false
+          opa:
+            rego: |
+              # Tenant hostname isolation.
+              # Each Gateway listener is bound to a specific tenant hostname
+              # (e.g. coke.best.com, pepsi.worst.com). A valid API key from
+              # tenant A must be rejected if it arrives on tenant B's gateway.
+              #
+              # PoC Phase 1: validate that the API key's tenant matches the
+              # gateway hostname by calling maas-api tenant lookup.
```
jland-redhat: If we are attached the gateway level anyway we should just pass in the gateway name and key based off that instead of hostname.

jland-redhat (commented):
Yeah I think the path based probably the way we would need to do it for now but want that confirmation mentioned above.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -415,394 +454,315 @@ type authPolicyRef struct {
...
+			"metrics":  false,
+			"priority": int64(2),
+		},
+	}
+
+	if oidc != nil {
+		authenticationRules["oidc-identities"] = map[string]any{
+			"jwt": map[string]any{
+				"issuerUrl": oidc.IssuerURL,
+				"ttl":       int64(300),
```
jrhyness, line 522: Do we need audiences here?

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -224,21 +224,44 @@ const (
...
+// so requestedModel is omitted and the subscription selector returns all accessible subscriptions.
+const celModelIdentity = `(request.path.startsWith("/llm/")` +
+	` ? request.path.split("/").filter(x, x != "")[0] + "/" + request.path.split("/").filter(x, x != "")[1]` +
+	` : ("x-gateway-model-name" in request.headers` +
+	` ? request.headers["x-gateway-model-name"]` +
+	` : ""))`
+
+// maasGatewayAuthPolicyName is the singleton AuthPolicy that targets the Gateway.
+// All MaaSAuthPolicy CRs share this one policy; model identity is resolved dynamically.
+const maasGatewayAuthPolicyName = "maas-gateway-auth"
```
jrhyness, line 240: Is it a true singleton, or an auth policy per gateway?  
Do we need a per-tenant name for the auth policy?

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
ishitasequeira: The existing `maas-api-auth-policy` on main branch doesn't set audiences either. Adding it would be a separate hardening task. Although, I do agree on having audiences added as a separate task for hardening the auth policy.

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
ishitasequeira: Currently one tenant = one gateway = one auth policy. The name should be per-gateway/per-tenant for multi-tenant.

jland-redhat: We will also need to make sure merging plays nicely together with that.

jland-redhat (approved):
Lgtm

PR #914, opened 2026-05-15
feat: integrate cluster TLS security profile for maas-controller and maas-api

dmytro-zaharnytskyi:
## Description

Integrates both MaaS components with the OpenShift cluster-wide TLS security profile
(`config.openshift.io/v1 APIServer`), as part of the Tier 2 TLS compliance effort
([RHOAIENG-61070](https://redhat.atlassian.net/browse/RHOAIENG-61070)).

**maas-controller:**
- Fetches the cluster TLS profile and TLS adherence policy at startup using
  `github.com/openshift/controller-runtime-common/pkg/tls`
  with retry logic (3 attempts, 10s timeout each)
- Distinguishes non-OpenShift clusters (`NoMatchError` -> graceful fallback) from
  transient OpenShift/RBAC/API failures, which are logged before falling back to
  the Intermediate TLS profile so the watcher can self-heal when the API recovers
- Converts the resolved OpenShift TLS profile into a controller-runtime `TLSOpts`
  callback via `NewTLSConfigFromProfile` and applies it to the webhook server on
  port 9443
- Sets webhook `NextProtos` to `h2` and `http/1.1`
- Registers the upstream `SecurityProfileWatcher` controller that monitors the
  `APIServer` resource for profile/adherence changes and triggers graceful
  shutdown so the pod restarts with the updated webhook TLS config
- Falls back to the Intermediate profile (TLS 1.2) on non-OpenShift clusters;
  watcher registration is skipped when the `APIServer` resource does not exist
- Adds RBAC (`get/list/watch` on `config.openshift.io/apiservers`)

**maas-api:**
- Fetches the cluster TLS profile at startup via the typed OpenShift config
  client (`openshift/api` + `openshift/client-go`) with retry logic (3 attempts,
  then falls back to the Intermediate TLS profile after persistent fetch errors so
  startup is not blocked by transient config API issues)
- Replaces the flag-only `MinVersion` in `buildTLSConfig()` with profile-driven
  `MinVersion` and `CipherSuites` for the HTTPS server on port 8443
- Sets API server `NextProtos` to `h2` and `http/1.1`
- Uses the Intermediate TLS profile when the OpenShift config API is unavailable;
  `--tls-min-version` / `TLS_MIN_VERSION` only applies when HTTPS or cluster
  config is unavailable
- Adds a typed OpenShift config informer watcher that cancels the server context
  on profile changes, causing a graceful shutdown/restart so updated TLS settings apply
- Logs watcher construction/start failures and continues with the current TLS
  profile so startup is not blocked by watcher availability
- Adds RBAC to both the main and supplemental ClusterRoles
- Adds defensive nil checks on exported functions (`FetchTLSProfile`, `NewWatcher`)

**Shared TLS profile behavior:**
- Supports OpenShift named profiles: `Old`, `Intermediate`, and `Modern`
- Supports `Custom` profiles with explicit cipher list and `minTLSVersion`
- Uses OpenShift's `configv1.TLSProfiles` definitions instead of local profile
  copies
- Maps OpenSSL/IANA cipher names from OpenShift profiles to Go `crypto/tls`
  constants
- Skips TLS 1.3 cipher suites when setting `tls.Config.CipherSuites`, because Go
  manages TLS 1.3 cipher suites internally
- Deep-copies predefined profile cipher slices before returning them so callers
  cannot mutate package-level profile definitions
- Uses the Intermediate profile when no cluster profile is configured or the
  cluster profile cannot be fetched during startup

**Documentation:**
- Adds a "Cluster TLS Security Profile" section in `tls-configuration.md`
- Updates the controller description to reflect the webhook TLS listener on 9443
- Documents non-OpenShift fallback behavior and profile-change restart behavior
- Clarifies the current gateway-to-maas-api certificate verification limitation:
  Gateway API `BackendTLSPolicy` is the standard direction, but the managed
  Gateway provider still needs a supported CA-reference path for this deployment

**Dependency approach:**
- `maas-controller` follows the RHOAI controller-runtime pattern via
  `controller-runtime-common`, which requires a controller-runtime/Kubernetes
  dependency refresh. Webhook setup was updated for controller-runtime v0.23.x.
- `maas-api` follows the non-controller-runtime pattern via direct
  `openshift/api` and `openshift/client-go` usage.

**Not in scope / known limitations:**
- `DestinationRule` `insecureSkipVerify: true` for gateway -> `maas-api` TLS
  origination is unchanged in this PR. Istio can verify upstream certificates
  with `caCertificates` or `credentialName`, but the managed gateway proxy does
  not currently mount or receive the OpenShift service-CA bundle that signs the
  `maas-api` serving certificate. Removing this requires platform wiring that
  makes the service CA available to the gateway proxy, or a managed Gateway
  implementation that supports `BackendTLSPolicy` with the required CA reference.
- The cleanup CronJob TLS overlay still uses `curl -k` when calling
  `https://maas-api:8443/internal/v1/api-keys/cleanup`. Removing this requires a
  service-CA trust-bundle strategy for the cleanup container.
- `maas-controller` metrics remains HTTP on 8080. This PR applies the central TLS
  profile to the webhook TLS server on 9443, which is the controller TLS server
  surface introduced/enabled here.

## Risk analysis

- **Risk rating**: 4
- **Why**: TLS changes affect server startup, webhook TLS configuration, API HTTPS
  configuration, and graceful restart behavior. The profile watcher introduces
  new restart paths for both components, and the controller webhook server now
  receives explicit TLS settings from the cluster profile. The PR also refreshes
  Kubernetes/controller-runtime/OpenShift dependencies to use the expected central
  TLS helpers and typed clients. This behavior is not covered by the default e2e
  smoke path (`test/e2e/scripts/prow_run_smoke_test.sh` does not exercise profile
  changes, Modern profile handshakes, TLS profile fallback/self-heal behavior, or
  dependency upgrade surface). Mitigations include non-OpenShift fallback
  detection, log-and-continue Intermediate fallback behavior, focused unit tests
  for parsing/conversion and server config, and manifest validation.

## How Has This Been Tested?

- Focused local validation:
  - From `maas-controller/`:
    `GOTOOLCHAIN=auto GOSUMDB=sum.golang.org GOCACHE=/tmp/maas-go-build GOMODCACHE=/tmp/maas-go-mod go test ./pkg/webhook ./cmd/manager`
  - From `maas-api/`:
    `GOTOOLCHAIN=auto GOSUMDB=sum.golang.org GOCACHE=/tmp/maas-go-build GOMODCACHE=/tmp/maas-go-mod go test ./internal/tlsprofile ./cmd`
  - From repo root:
    `./scripts/ci/validate-manifests.sh`
- Previous validation on this PR:
  - `make -C maas-controller lint`
  - `make -C maas-controller test`
  - `make -C maas-controller generate manifests && make -C maas-controller verify-codegen`
  - `make -C maas-api lint`
  - `make -C maas-api test`
- Deployed to OpenShift cluster (devshift) with custom images:
  - Both components log that the cluster TLS security profile was fetched
  - TLS watcher controller starts and watches `config.openshift.io/APIServer`
  - RBAC verified: both component service accounts can get
    `apiservers.config.openshift.io`
  - `maas-api` serves TLS 1.2 and TLS 1.3 on port 8443 with the Intermediate
    profile
  - `maas-api` logs profile fetch failures and starts with the Intermediate TLS
    profile while the watcher can self-heal after the cluster profile becomes
    available
  - Profile change detection could not be fully tested on managed OpenShift
    because `APIServer` modification was blocked by admission policy, but watcher
    startup and registration were confirmed

## Merge criteria

- [x] Testing instructions have been added in the PR body.

[RHOAIENG-61070]: https://redhat.atlassian.net/browse/RHOAIENG-61070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


maas-api/cmd/main.go
```diff
@@ -257,3 +267,93 @@ func debugCORSConfig() cors.Config {
...
+		log.Warn("TLS profile produced no TLS 1.2 cipher suites; Go defaults will be used for TLS 1.2 negotiation")
+	}
+
+	watcher, watchErr := tlsprofile.NewWatcher(restConfig, profile, func(oldProfile, newProfile tlsprofile.ProfileSpec) {
+		log.Info("TLS security profile changed, initiating graceful shutdown to reload",
+			"oldType", string(oldProfile.Type), "newType", string(newProfile.Type))
+		cancel()
+	})
+	if watchErr != nil {
+		log.Info("Could not start TLS profile watcher", "error", watchErr)
```
ishitasequeira: nit: Make this `log.Error` or `log.Warn`

maas-api/cmd/main.go
```diff
@@ -257,3 +267,93 @@ func debugCORSConfig() cors.Config {
...
+	} else {
+		go func() {
+			if err := watcher.Start(ctx.Done()); err != nil {
+				log.Error("TLS profile watcher failed", "error", err)
+			}
+		}()
```
ishitasequeira: If Start returns an error after it's been running for a while, we currently just log it and continue. In such cases, the watcher would be silently dead for the rest of this pod's lifetime.

Since we already call `cancel()` on a profile change to trigger a graceful restart, it might make sense to do the same here?
```
go func() {
    if err := watcher.Start(ctx.Done()); err != nil {
        log.Error("TLS profile watcher failed, initiating restart", "error", err)
        cancel()
    }
}()
```
This way a watcher failure gets the same treatment as a profile change and a clean restart rather than silent degradation.

Also a minor nit: since `Start` always needs to be called in a goroutine, it might be worth having the watcher own that internally with something like `StartAsync(ctx)`, just to make it harder to accidentally deadlock in the future. Totally up to you though.

maas-controller/cmd/manager/main.go
```diff
@@ -466,6 +507,23 @@ func main() {
...
+	k8sClient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		setupLog.Error(err, "unable to create pre-manager Kubernetes client for TLS profile fetch")
+		os.Exit(1)
+	}
+	tlsProfile, tlsProfileAvailable := fetchTLSProfileWithRetry(ctx, k8sClient)
+	if tlsProfileAvailable {
+		setupLog.Info("fetched cluster TLS security profile",
+			"type", tlsProfile.Type, "minTLSVersion", tlsProfile.MinTLSVersion)
+	}
```
ishitasequeira: If the startup fetch fails, the SecurityProfileWatcher reconciler never gets registered. So there's no retry and no recovery until someone manually restarts the pod.

I noticed the reconciler would do this same fetch on its first reconcile pass anyway since controller-runtime fires an initial reconcile for the existing cluster object. Was there a reason to do it upfront in `main.go` rather than just letting the reconciler handle it? Seems like registering it unconditionally would give you retry-on-failure for free without any extra work.

maas-api/cmd/main.go
dmytro-zaharnytskyi: fixed a graceful restart, but left nit for now. Thanks!

maas-controller/go.mod
```diff
@@ -131,12 +124,15 @@ require (
...
 	k8s.io/utils v0.0.0-20260319190234-28399d86e0b5 // indirect
 	knative.dev/serving v0.48.1 // indirect
-	sigs.k8s.io/gateway-api-inference-extension v1.3.1 // indirect
+	sigs.k8s.io/gateway-api-inference-extension v0.3.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 )
+
+replace github.com/kserve/kserve => github.com/opendatahub-io/kserve v0.0.0-20260112171902-47894470ea49
```
ishitasequeira, line 142: Any reason to need a replace with `opendatahub-io/kserve` instead of relying on upstream github.com/kserve/kserve?

dmytro-zaharnytskyi: Yep, we need that (and I also don't like it :smile: )

With upstream github.com/kserve/kserve v0.19.0, the controller packages fail to compile because github.com/kedacore/keda/v2@v2.18.0 still calls the old ctrl.NewWebhookManagedBy(mgr) API, which is incompatible with controller-runtime v0.23.x.

I don't see a better way now, but would appreciate if you have any good idea

maas-api/cmd/main.go
```diff
@@ -267,3 +277,94 @@ func debugCORSConfig() cors.Config {
...
+
+	return profileMinVersion, profileCipherSuites, nil
+}
+
+// fetchTLSProfileWithRetry attempts to fetch the OpenShift TLS security profile.
+// If the config.openshift.io API doesn't exist (non-OpenShift), returns
+// available=false with nil error. For transient errors on OpenShift, retries a
+// few times and returns a non-nil error if all attempts fail (fail-closed).
+func fetchTLSProfileWithRetry(ctx context.Context, log *logger.Logger, restConfig *rest.Config) (tlsprofile.ProfileSpec, bool, error) {
+	const maxRetries = 3
```
jland-redhat: Do we want this to be hard coded?

maas-controller/pkg/platform/tenantreconcile/apply.go
```diff
@@ -73,7 +74,17 @@ func ApplyRendered(ctx context.Context, c client.Client, scheme *runtime.Scheme,
 		unstructured.RemoveNestedField(u.Object, "status")
 		// ForceOwnership is intentional: maas-controller is the sole manager for
 		// Tenant platform resources, ensuring a clean field-manager handoff.
-		if err := c.Patch(ctx, u, client.Apply, client.FieldOwner(ssaFieldOwner), client.ForceOwnership); err != nil {
+		if err := c.Apply(ctx, client.ApplyConfigurationFromUnstructured(u), client.FieldOwner(ssaFieldOwner), client.ForceOwnership); err != nil {
+			if apimeta.IsNoMatchError(err) && isOptionalAPIGroup(u.GroupVersionKind().Group) {
```
jland-redhat: What does this have to do with TLS?

maas-controller/pkg/platform/tenantreconcile/constants.go
```diff
@@ -13,6 +13,20 @@ import (
 	maasv1alpha1 "github.com/opendatahub-io/models-as-a-service/maas-controller/api/maas/v1alpha1"
 )
 
+// OptionalAPIGroups lists API groups whose CRDs are installed by optional platform
+// components (e.g. COO for Perses). Resources in these groups are skipped gracefully
+// when their CRDs are not yet registered, instead of failing the Tenant reconcile.
+// The CRD watch in the controller re-triggers reconcile once the CRDs appear.
+var OptionalAPIGroups = map[string]bool{
```
jland-redhat: I think this may have been left over code

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -409,7 +409,7 @@ func (r *LifecycleReconciler) ensureLimitadorServiceMonitor(ctx context.Context)
 		return fmt.Errorf("set owner reference on ServiceMonitor: %w", err)
 	}
 
-	if err := r.Patch(ctx, sm, client.Apply, client.ForceOwnership, client.FieldOwner("maas-controller")); err != nil {
+	if err := r.Apply(ctx, client.ApplyConfigurationFromUnstructured(sm), client.ForceOwnership, client.FieldOwner("maas-controller")); err != nil {
```
jland-redhat, line 467: This is another place where I am not sure TLS is relevant

maas-api/cmd/main.go
```diff
@@ -267,3 +277,94 @@ func debugCORSConfig() cors.Config {
...
+		log.Warn("TLS profile produced no TLS 1.2 cipher suites; Go defaults will be used for TLS 1.2 negotiation")
+	}
+
+	watcher, watchErr := tlsprofile.NewWatcher(restConfig, profile, func(oldProfile, newProfile tlsprofile.ProfileSpec) {
+		log.Info("TLS security profile changed, initiating graceful shutdown to reload",
+			"oldType", string(oldProfile.Type), "newType", string(newProfile.Type))
+		cancel()
+	})
+	if watchErr != nil {
+		log.Warn("Could not start TLS profile watcher", "error", watchErr)
```
jland-redhat: I think we should be failing if our watcher fails here.

We may want to give a way for a user to explicitly turn off the watch for certain cases but I think we need to do more than a warn

jland-redhat (changes requested):

maas-api/cmd/main.go
dmytro-zaharnytskyi: nope, just followed the retry pattern, but agree, for cleaner solution, will change it to timeout or something like that

maas-controller/pkg/platform/tenantreconcile/apply.go
dmytro-zaharnytskyi: It appeared while adapting to the controller-runtime update, but agree, mostly out of scope (as well as 2 further comments), so will remove that for now

maas-api/cmd/main.go
dmytro-zaharnytskyi: agree, will change so it returns an error

.gitignore
```diff
@@ -21,6 +21,8 @@ coverage.html
 
 # Go binaries
 bin/
+maas-controller/manager
```
jland-redhat, line 24: Why were these changes added? Does not seem relevant to the PR

maas-controller/pkg/webhook/aitenant_webhook.go
```diff
@@ -21,7 +21,6 @@ import (
 	"errors"
```
jland-redhat, line 21: Why is anything in any of these  webhook files changed? 

Nothing here seems relevant to TLS

maas-controller/pkg/webhook/maasauthpolicy_webhook.go
```diff
@@ -21,7 +21,6 @@ import (
 	"errors"
```
jland-redhat, line 21: Same comment as before where I don't see anything relevant to TLS

maas-controller/pkg/webhook/maassubscription_webhook.go
```diff
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/runtime"
```
jland-redhat, line 24: Same comment as before where I don't see anything relevant to TLS

maas-controller/pkg/controller/maas/self_deployment_controller.go
jland-redhat: Please don't resolve without responding

maas-api/go.mod
```diff
@@ -22,10 +24,10 @@ require (
 	go.uber.org/zap v1.27.0
 	golang.org/x/sync v0.20.0
 	google.golang.org/grpc v1.81.1
-	k8s.io/api v0.34.1
-	k8s.io/apimachinery v0.34.1
-	k8s.io/client-go v0.34.1
-	k8s.io/utils v0.0.0-20250820121507-0af2bda4dd1d
+	k8s.io/api v0.35.1
```
jland-redhat, line 27: If these changes are not required for the TLS change please revert them

jland-redhat (changes requested):
I stopped reviewing this. There are a TON of changes here unrelated to TLS.

Please make sure that your PRs are limited to what is required for the story so they are easier to review.

bartoszmajsak:
This PR should arguably be two or three PRs:

- TLS profile integration - the actual feature, which is looking good and seems to be backward compatible
- controller-runtime v0.23.x migration - Patch->Apply, typed webhooks, event recorder - although mechanical, carries upgrade risk which we could defer
- behavioral changes - namespace deletion semantics, alias resolution, endpoint preference, cleanup ordering - these are independent design decisions that should have their own review and rollout and do not look like related to TLS

We should keep PRs scoped to make our lives easier - both for reviewers, but then also when testing changes.

dmytro-zaharnytskyi:
> This PR should arguably be two or three PRs:
> 
> * TLS profile integration - the actual feature, which is looking good and seems to be backward compatible
> * controller-runtime v0.23.x migration - Patch->Apply, typed webhooks, event recorder - although mechanical, carries upgrade risk which we could defer
> * behavioral changes - namespace deletion semantics, alias resolution, endpoint preference, cleanup ordering - these are independent design decisions that should have their own review and rollout and do not look like related to TLS
> 
> We should keep PRs scoped to make our lives easier - both for reviewers, but then also when testing changes.

Work on splitting right now :hugs:

dmytro-zaharnytskyi:
Splitting into a few new PRs

PR #915, opened 2026-05-17
fix: remove namespace field from model refs in GET /subscriptions response

liangwen12year:
## Summary

- Remove the `namespace` (project) field from `ModelRefInfo` in the GET /subscriptions API response
- The field exposed internal cluster metadata that platform consumers do not need

## Breaking Change

Removes the `namespace` field from `modelRefs` objects in the `GET /v1/subscriptions` response. This field exposed internal cluster topology (which namespace a MaaSModelRef lives in) which is not relevant to API consumers and could leak implementation details. Clients should use `name` to identify model refs.

- **Impact**: Clients parsing `namespace` from subscription model refs will receive `undefined`/missing field instead
- **Migration**: Remove any client-side references to `modelRefs[].namespace` — use `modelRefs[].name` to identify model refs
- **Reason**: Defense in depth — internal namespace metadata should not be exposed in public API responses

## Changes

- **`maas-api/internal/subscription/types.go`**: Change `ModelRefInfo.Namespace` JSON tag from `"namespace,omitempty"` to `"-"` — excluded from JSON serialization, still available internally for model matching in the subscription selector
- **`maas-api/openapi3.yaml`**: Remove `namespace` property from the model ref schema
- **`test/e2e/tests/test_subscription_list_endpoints.py`**: Add assertion that no model ref in the response contains a `namespace` field

## Test plan

- [x] `ModelRefInfo.Namespace` still populated internally from MaaSSubscription CRD spec
- [x] Model matching in `selector.go` still works (uses Go struct field, not JSON)
- [x] E2E test asserts `namespace` absent from response
- [ ] `go build ./...` passes
- [ ] `go test ./internal/subscription/...` passes
- [ ] E2E: GET /subscriptions returns model refs without `namespace`

Ref: [RHAISTRAT-1547](https://redhat.atlassian.net/browse/RHAISTRAT-1547)
https://redhat.atlassian.net/browse/RHOAIENG-62491

## Summary by CodeRabbit

* **Bug Fixes**
  * Model reference objects in subscription API responses no longer include the namespace field.

* **Documentation**
  * API schema updated to remove the namespace property from model reference definitions and clarify name examples.

* **Tests**
  * End-to-end tests strengthened to verify model reference responses do not expose a namespace field.

### DIFF
--- a/maas-api/internal/subscription/types.go
+++ b/maas-api/internal/subscription/types.go
@@ -54,7 +54,7 @@ type SubscriptionInfo struct {
 // ModelRefInfo represents a model reference with its rate limits.
 type ModelRefInfo struct {
 	Name            string           `json:"name"`
-	Namespace       string           `json:"namespace,omitempty"`
+	Namespace       string           `json:"-"`
 	DisplayName     string           `json:"display_name,omitempty"`
 	Description     string           `json:"description,omitempty"`
 	TokenRateLimits []TokenRateLimit `json:"token_rate_limits,omitempty"`
--- a/.github/workflows/openapi-validation.yml
+++ b/.github/workflows/openapi-validation.yml
@@ -112,8 +112,9 @@ jobs:
           git fetch origin "$BASE_REF"
           git show "origin/$BASE_REF:maas-api/openapi3.yaml" > base-spec.yaml
 
-          # Run breaking change detection
-          oasdiff breaking base-spec.yaml maas-api/openapi3.yaml > breaking-changes.txt || true
+          # Run breaking change detection (--warn-ignore suppresses acknowledged warnings)
+          oasdiff breaking base-spec.yaml maas-api/openapi3.yaml \
+            --warn-ignore maas-api/oasdiff-warn-ignore.txt > breaking-changes.txt || true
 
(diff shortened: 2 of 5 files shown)


maas-api/internal/subscription/types.go
```diff
@@ -54,7 +54,7 @@ type SubscriptionInfo struct {
 // ModelRefInfo represents a model reference with its rate limits.
 type ModelRefInfo struct {
 	Name            string           `json:"name"`
-	Namespace       string           `json:"namespace,omitempty"`
+	Namespace       string           `json:"-"`
```
EgorLu, line 57: @liangwen12year do we also need to update the openapi spec?

liangwen12year: I updated the openapi spec, nice catch.

liangwen12year:
The `namespace` field removal from the API spec is intentional.  So the OpenAPI validation failure is expected.

EgorLu (dismissed):
LGTM

EgorLu (dismissed):
LGTM
Edit: I now see that the OpenAPI test still needs to be overwritted

chaitanya1731 (approved):
LGTM

PR #920, opened 2026-05-19
chore: add isManaged check for all resources managed by maas-controller

ishitasequeira:
## Description

Audits all resource types reconciled by `maas-controller` to confirm the `opendatahub.io/managed=false` opt-out annotation is respected on both the write (create/update) and delete paths.

## How Has This Been Tested?

New unit tests in `pkg/reconciler/externalmodel/reconciler_test.go` cover the added behaviour:

- `TestManagedAnnotation_Service` — write-path opt-out: pre-existing Service with `managed=false` keeps its sentinel `ExternalName` after reconcile.
- `TestManagedAnnotation_ServiceEntry` — write-path opt-out: pre-existing ServiceEntry spec is untouched.
- `TestManagedAnnotation_HTTPRoute` — write-path opt-out: pre-existing HTTPRoute parent ref is untouched.
- `TestManagedAnnotation_DestinationRule_DeletePath` — delete-path opt-out: stale DestinationRule with `managed=false` survives the TLS→no-TLS transition that would otherwise call `deleteIfExists`.
- `TestIsManaged` — unit-tests the helper across all edge cases (nil annotations, absent key, `false`, `true`, other values, case-sensitivity).

Each test includes the three-way table `absent / managed=false / managed=true` to confirm the positive and negative cases.
https://redhat.atlassian.net/browse/RHOAIENG-60911

## Merge criteria:

## Release Notes

* **New Features**
  * Resources can now opt-out of automatic management by setting the `opendatahub.io/managed=false` annotation. When applied, the controller will skip updates and deletions for annotated resources.

* **Tests**
  * Added comprehensive test coverage for opt-out annotation behavior across service management, gateway routing, and resource lifecycle operations.

### DIFF
--- a/maas-controller/pkg/controller/maas/annotations.go
+++ b/maas-controller/pkg/controller/maas/annotations.go
@@ -16,10 +16,15 @@ limitations under the License.
 
 package maas
 
-import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/opendatahub-io/models-as-a-service/maas-controller/pkg/platform/tenantreconcile"
+)
 
 // ManagedByODHOperator is used to denote if a resource/component should be reconciled - when missing or true, reconcile.
-const ManagedByODHOperator = "opendatahub.io/managed"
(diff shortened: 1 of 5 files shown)


jland-redhat (approved):
lgtm

PR #924, opened 2026-05-20
feat: add old tier resource cleanup to migration script

EgorLu:
## Summary

- Extends `scripts/migrate-tier-to-subscription.sh` with a cleanup phase that deletes old tier-based resources (AuthPolicy, TokenRateLimitPolicy, ConfigMap, tier annotations) **before** creating new subscription CRs
- Adds `--cleanup-only` flag for standalone cleanup without CR generation
- Updates `docs/content/migration/tier-to-subscription.md` to document the new flag and rewrite Phase 4 with automated cleanup + manual fallback

Resolves [RHOAIENG-62437](https://redhat.atlassian.net/browse/RHOAIENG-62437) 

## Details

The cleanup phase handles 4 resources:

| Resource | Kind | Namespace |
|----------|------|-----------|
| `gateway-auth-policy` | AuthPolicy | openshift-ingress |
| `gateway-tier-rate-limits` | TokenRateLimitPolicy | openshift-ingress |
| `tier-to-group-mapping` | ConfigMap | maas-api |
| `alpha.maas.opendatahub.io/tiers` | annotation on LLMInferenceService | model namespace (default: llm) |

Key behaviors:
- **Idempotent**: re-running on an already-clean cluster exits 0 with "already clean" messages
- **Dry-run**: `--cleanup-only --dry-run --verbose` previews what would be deleted
- **Normal flow**: cleanup runs automatically before CR generation
- **Standalone**: `--cleanup-only` skips `--tier`/`--models`/`--rate-limit` validation

## Test plan

- [x] `bash -n` syntax check passes
- [x] `--help` shows `--cleanup-only` flag and examples
- [x] `--cleanup-only --dry-run --verbose` correctly previews resources without deleting
- [x] `--cleanup-only --verbose` deletes all 3 tier resources on a live cluster
- [x] Idempotent re-run shows "already clean" (not false "Deleted" messages)
- [x] `--dry-run` (without `--cleanup-only`) still requires `--tier`
- [x] Manifest validation passes (`./scripts/ci/validate-manifests.sh`)
- [ ] E2E: run full migration with cleanup on a cluster with tier resources + LLMInferenceServices with tier annotations

## Risk analysis

- **Risk rating:** 3
- **Why:** Modifies a customer-facing migration script that deletes cluster resources. The deletions target hardcoded, well-known resource names (no wildcards), and all operations use `--ignore-not-found` for safety. Tested on a live cluster. Rating is 3 (not higher) because the script is a manual pre-enablement tool, not part of the automated operator upgrade path — customers explicitly choose to run it.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Documentation**
  * Updated the tier-to-subscription migration guide with clearer Phase 4 instructions, including the required post-Phase 3 validation step and an automated cleanup process plus a shorter manual fallback.

* **New Features**
  * Added an automated cleanup script for removing legacy tier resources and tier annotations.
  * Supports `--dry-run`, `--verbose`, and `--help`, provides progress output, and exits non-zero if any cleanup step fails.

[RHOAIENG-62437]: https://redhat.atlassian.net/browse/RHOAIENG-62437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/content/migration/tier-to-subscription.md
+++ b/docs/content/migration/tier-to-subscription.md
@@ -85,9 +85,9 @@ Run both old and new systems in parallel, validate the new system, then switch o
 
 **Approach:**
 1. Install maas-controller (creates gateway defaults)
-2. Create new MaaS CRs alongside existing tier configuration
+2. Create new MaaS CRs alongside existing tier configuration (`migrate-tier-to-subscription.sh`)
 3. Validate new system works correctly
-4. Remove old tier-based configuration
+4. Remove old tier-based configuration (`cleanup-tier-resources.sh`)
 
 ### Option B: Full Cutover (Requires Downtime)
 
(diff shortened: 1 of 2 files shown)


EgorLu:
@coderabbitai review

scripts/migrate-tier-to-subscription.sh
```diff
@@ -216,6 +228,103 @@ while [[ $# -gt 0 ]]; do
...
+                log_error "Failed to delete TokenRateLimitPolicy gateway-tier-rate-limits in openshift-ingress"
+                cleanup_failed=true
+            fi
+        fi
+    else
+        log_verbose "TokenRateLimitPolicy gateway-tier-rate-limits not found in openshift-ingress (already clean)"
+    fi
+
+    # 3. Delete tier-to-group-mapping ConfigMap
+    if kubectl get configmap tier-to-group-mapping -n maas-api >/dev/null 2>&1; then
```
chaitanya1731: Is the namespace `maas-api` correct? I think this needs to be revisited.

scripts/migrate-tier-to-subscription.sh
```diff
@@ -339,6 +448,12 @@ if ! validate_resource_name "$SUBSCRIPTION_NAMESPACE" "Subscription namespace" 6
     exit 1
 fi
 
+# Run cleanup phase before generating new CRs
+cleanup_tier_resources || {
```
chaitanya1731: Also AI flagged a major concern that I agree.. 
In normal (non-`--cleanup-only`) mode, `cleanup_tier_resources()` runs before CR generation and `--apply`. 

This means the old `gateway-auth-policy` and `gateway-tier-rate-limits` are deleted before new subscription CRs exist on the cluster. Two risks:
1. **Brief auth/rate-limit gap** — between deletion of the old policies and reconciliation of the new CRs, models may be unauthenticated or unthrottled. On a production cluster this could be seconds to minutes depending on controller reconcile time.
2. **Broken state on failure** — if CR generation or `kubectl apply` fails after cleanup, the cluster is left with neither old nor new policies in place, requiring manual recovery.
This also contradicts the phased approach in the migration doc itself, where Phase 4 (remove old config) comes *after* Phase 3 (validate new system is working).
**Suggestions (pick one):**
- Move the automatic cleanup to run *after* successful `--apply` and a brief validation check, so new policies are confirmed active before old ones are removed.
- Add a `--skip-cleanup` flag (defaulting to cleanup-enabled) so users can opt out in the normal flow and run `--cleanup-only` separately once validated.
- At minimum, gate the automatic cleanup behind a confirmation prompt (similar to how namespace creation is handled in interactive mode).

Can you check..

chaitanya1731 (commented):
@EgorLu added few comments..

scripts/migrate-tier-to-subscription.sh
EgorLu: Mmm I agree. I thought about the issue that this is not an "atomic" operation, and can't be undone on failure, but the part when the models remain "unprotected" is even more important.

I'll extract the cleanup from the migration script and make it a separate flow. Then document it so that it's clear for the user.

Migrate -> validate -> clear.

EgorLu:
## Addressing review feedback

Thanks @chaitanya1731 for the review!

**Re: `maas-api` namespace** — Confirmed correct. The `tier-to-group-mapping` ConfigMap lives in `maas-api`. Verified on a live cluster.

**Re: cleanup ordering risk** — Great catch, fully agree. I've reworked the approach:

**Before:** Cleanup was embedded in the migration script and ran automatically before CR generation (risky gap + broken state on failure).

**After:** Cleanup is now a **standalone script** (`scripts/cleanup-tier-resources.sh`), completely separate from the migration script. The migration script (`migrate-tier-to-subscription.sh`) is unchanged — it only generates/applies CRs.

The documented workflow is now a clear 3-phase process:
1. **Migrate** — create subscription CRs (`migrate-tier-to-subscription.sh --apply`)
2. **Validate** — verify the new system works (Phase 3)
3. **Cleanup** — remove old tier resources (`cleanup-tier-resources.sh`) only after validation passes

This eliminates both risks: no auth gap (old policies stay until explicitly removed) and no broken state (cleanup is a separate, reversible step).

jland-redhat (approved):
lgtm

jland-redhat:
lgtm

PR #926, opened 2026-05-20
ci: replace PR-based stable→rhoai promotion with direct merge commit

chaitanya1731:
## Summary

Replace the PR-based stable → rhoai branch promotion workflow with a direct merge commit approach that requires zero merge conflicts.

Ref: [RHOAIENG-63731](https://redhat.atlassian.net/browse/RHOAIENG-63731)

## Description

The current `promote-stable-to-rhoai` workflow creates a pull request from `stable` to `rhoai`, requiring manual review and approval. Since `stable` is already a validated branch, this adds unnecessary overhead.

- Replace PR creation/update logic with a direct `git merge --no-ff` into `rhoai`
- Add strict conflict detection via a dry-run merge before the actual merge
- Fail with a clear report of conflicting files if any conflicts exist
- Push the merge commit directly to `rhoai` only when the merge is clean
- Update permissions from `contents: read` + `pull-requests: write` to `contents: write`

## How it was tested

- Tested in fork by triggering `workflow_dispatch` from the feature branch
- Verified successful merge commit with no conflicts (stable synced from upstream)
- Confirmed job summary correctly reports merged commits
- Workflow run: https://github.com/chaitanya1731/models-as-a-service/actions/runs/26180802720

## Risk analysis

- **Risk rating**: 2
- **Why**: The change only affects an on-demand CI workflow (workflow_dispatch) for branch promotion. It does not touch application code, CRDs, or deployment manifests. The blast radius is limited to the rhoai branch sync process. The strict conflict check ensures no unintended changes can be pushed. Tested end-to-end in fork.

## Summary by CodeRabbit

* **Chores**
  * Updated the stable-to-rhoai promotion workflow to directly merge changes between branches instead of creating pull requests, with automatic conflict detection and conditional execution based on commit availability.

PR #927, opened 2026-05-20
ci: harden promote-stable-to-rhoai merge step

chaitanya1731:
## Summary

Follow-up to #926 — addresses review feedback from @ishitasequeira.

## Description

- Add `set -euo pipefail` to the "Merge stable into rhoai" step to ensure any silent command failure aborts execution before `git push` can run against bad state.
- Guard the dry-run `git merge --abort` with `|| true` to prevent spurious "no merge in progress" errors if the conflict path already aborted the merge.

## How it was tested

- Verified workflow syntax is valid
- Logic reviewed: `set -euo pipefail` catches unset variables (`-u`), pipe failures (`-o pipefail`), and non-zero exits (`-e`) before the push step
- `|| true` only applies to the abort after a successful dry-run merge, which is the only path that reaches it

## Risk analysis

- **Risk rating**: 1
- **Why**: Minimal change (2 lines) in a CI workflow that only runs on-demand. Adds strictly defensive behavior — cannot introduce regressions.

Ref: [RHOAIENG-63731](https://redhat.atlassian.net/browse/RHOAIENG-63731)

## Bug Fixes
* Improved merge conflict handling resilience in CI/CD workflows to prevent error masking during cleanup operations, ensuring clearer failure diagnostics.

### DIFF
--- a/.github/workflows/promote-stable-to-rhoai.yml
+++ b/.github/workflows/promote-stable-to-rhoai.yml
@@ -80,11 +80,12 @@ jobs:
             exit 1
           fi
 
-          git merge --abort
+          git merge --abort || true
 
       - name: Merge stable into rhoai
         if: steps.check.outputs.skip != 'true'
         run: |
+          set -euo pipefail
           git checkout rhoai
           git reset --hard origin/rhoai
 


chaitanya1731:
@coderabbitai review

ishitasequeira (approved):
lgtm

PR #933, opened 2026-05-21
feat: add AIGateway CRD and reconciler for multi-tenancy bootstrap

dmytro-zaharnytskyi:
## Description

Implements the initial `AITenant` slice for multi-tenancy Phase 1 ([RHOAIENG-63338](https://redhat.atlassian.net/browse/RHOAIENG-63338)).

`AITenant` is the infra-level bootstrap CR for creating a dedicated MaaS tenant environment. It lives in a controller-configured infrastructure namespace, defaulting to `ai-tenants`, and is intended to be created by cluster/MaaS infrastructure administrators. The controller creates the infrastructure namespace if it is missing.

This follows the two-CR model from the multi-tenancy syncs:

- **AITenant** (infrastructure namespace) configures tenant bootstrap: tenant namespace, tenant Gateway, OIDC mirror values, and tenant-admin RBAC.
- **Tenant** (tenant namespace) remains the temporary MaaS runtime config object and is created automatically by the `AITenant` reconciler as `Tenant/default-tenant`.

### What's added

- **AITenant API type** (`maas-controller/api/maas/v1alpha1/aitenant_types.go`): spec includes tenant namespace config, Gateway template fields (name, class, optional domain, optional TLS), OIDC settings, and RBAC subjects. `spec.tenantNamespace.name` is immutable via CEL. Status reports phase, tenant namespace, resolved Gateway ref, and Ready condition.
- **AITenant reconciler** (`maas-controller/pkg/controller/maas/aitenant_controller.go`): creates or labels the tenant namespace, creates or adopts the tenant Gateway, creates the temporary `Tenant/default-tenant`, creates tenant-admin Role/RoleBinding and per-object AITenant read Role/RoleBinding, updates status, and uses a finalizer for child cleanup.
- **Configured AITenant infra namespace**: `--aitenant-namespace` defaults to `ai-tenants`. `AITenant` CRs are rejected outside this namespace, and the manager creates/maintains the namespace automatically.
- **Tenant Gateway lifecycle**: the controller creates the Gateway in the configured Gateway namespace, defaults the Gateway name from the `AITenant` name, and mirrors the resolved ref into `Tenant/default-tenant`. If the Gateway already exists and is unowned, the controller adopts/reconciles it for migration by applying AITenant metadata and the desired Gateway spec. Adopted Gateways are not marked as controller-created.
- **Gateway cleanup semantics**: controller-created Gateways are marked with `maas.opendatahub.io/created-by-aitenant=true` and deleted on `AITenant` deletion. Pre-existing/adopted Gateways are preserved on deletion and only AITenant labels/annotations are removed.
- **Tenant namespace deletion behavior**: deleting an `AITenant` removes controller-owned child resources and strips controller-added labels/annotations from the tenant namespace, but leaves the namespace itself in place.
- **AITenant-managed tenant bridge behavior**: `Tenant/default-tenant` created by `AITenant` is labeled as AITenant-managed so existing tenant platform rendering can run for that tenant. The full ADR placement of per-tenant `maas-api` services in `ai-gateway-infra` is not implemented in this PR.
- **Generated CRD, RBAC, and deepcopy** for the new API/controller behavior.
- **Docs updates** for the `AITenant` CRD reference and controller flags.
- **E2E lifecycle coverage** for AITenant create and delete paths, split into separate tests.

### What's explicitly deferred

- Tenant CR rename (`MaaSContext`/`MaaSConfig`, pending ADR/follow-up).
- Dynamic cache/watch expansion for arbitrary tenant namespaces.
- Automatic default `AITenant` creation from existing `models-as-a-service` resources.
- Tenant-wide key revocation on tenant deletion.
- Moving mirrored Gateway/OIDC values out of `Tenant.spec` once the MaaS config/status API is settled.
- ADR `ai-gateway-infra` layout for per-tenant `maas-api` services and `maas-db-config`.
- Database/schema work for tenant-scoped API keys.
- ADR namespace deletion behavior. This PR intentionally preserves tenant namespaces on delete per review feedback, while still removing controller-owned child resources and metadata.
- Full tenant functional e2e coverage for Policies, Subscriptions, Models, and inference requests in newly created tenant namespaces.

## Risk analysis

- **Risk rating**: 4
- **Why**: This PR introduces a new `AITenant` CRD and controller path that can create tenant namespaces and Gateway API Gateways, bootstrap tenant resources, manage tenant RBAC/metadata, and create/maintain the configured AITenant infrastructure namespace. It also reconciles pre-existing unowned Gateways for migration. Namespace deletion remains non-destructive and adopted Gateways are preserved on delete, but the blast radius is platform/operator lifecycle behavior and may not be fully covered by the default smoke path, so the rating is conservative.

## How Has This Been Tested?

**Unit tests**

- AITenant full bootstrap: tenant namespace, Gateway, Tenant config, RBAC creation, status, labels, and annotations.
- Configured infrastructure namespace placement: valid default namespace and rejection outside `--aitenant-namespace`.
- Default tenant namespace allowed from the infra namespace.
- Pre-existing tenant namespace with `create=false`.
- Missing namespace with `create=false` sets Pending.
- Protected namespace rejection: ODH app namespace and same-as-infra namespace.
- Missing Gateway is created with default Gateway name/class/listeners.
- Pre-existing Gateway adoption with AITenant labels/annotations and desired spec reconciliation, without marking it as controller-created.
- Shared upsert AlreadyExists create-race handling: re-read and patch the raced object instead of returning early.
- Gateway ownership conflict rejection.
- Tenant adoption/update for pre-existing `Tenant/default-tenant`.
- Namespace ownership conflict rejection.
- Deletion cleanup: controller-created Gateway, Tenant, and RBAC children removed, adopted Gateway preserved with metadata cleanup, tenant namespace preserved.
- Deletion metadata cleanup: controller-added tenant namespace labels/annotations are removed.
- RBAC subject validation.
- Idempotency.
- Full OIDC mirroring and nil OIDC behavior.
- Child name truncation.
- Manager startup helper creates the configured AITenant namespace.

**E2E tests**

- `test_aitenant_create_bootstrap_resources`: creates an `AITenant` in the configured infra namespace and verifies tenant namespace, Gateway, Tenant config, RBAC, and status.
- `test_aitenant_delete_cleans_up_bootstrap_resources`: deletes an `AITenant` and verifies Gateway/Tenant/RBAC cleanup, tenant namespace preservation, and metadata cleanup.
- Added TODO breadcrumb for follow-up e2e coverage that Policies, Subscriptions, Models, and inference requests work end-to-end in a newly created AITenant tenant namespace.

**Verification commands run**

```bash
GOCACHE=/tmp/maas-go-build GOTOOLCHAIN=auto GOSUMDB=sum.golang.org go test ./...
python -B -m py_compile test/e2e/tests/test_aitenant_lifecycle.py
GOCACHE=/tmp/maas-go-build GOTOOLCHAIN=auto GOSUMDB=sum.golang.org make -C maas-controller generate manifests
./scripts/ci/validate-manifests.sh
git diff --check
```

## Merge criteria

## Summary by CodeRabbit

* **New Features**
  * Added AIGateway custom resource: tenant namespace provisioning, gateway templating, TLS certificate refs, optional OIDC, tenant-admin RBAC, status (phase/Ready) and validation rules.

* **New Features**
  * Added controller to manage AIGateway lifecycle (namespace, gateway, tenant, RBAC, finalizers) with deterministic child naming.

* **Documentation**
  * Added AIGateway tenant bootstrap architecture doc and site navigation entry.

* **Tests**
  * Added tests for domain-driven HTTP/HTTPS gateway listener creation.

* **Chores**
  * Safer apply behavior: skip conflicting server-side apply for externally-owned resources.


maas-controller/api/maas/v1alpha1/aigateway_types.go
```diff
@@ -0,0 +1,227 @@
...
+	Hostname string `json:"hostname,omitempty"`
+
+	// Port is the listener port.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port int32 `json:"port"`
+
+	// Protocol is the listener protocol.
+	// +kubebuilder:validation:Enum=HTTP;HTTPS;TLS;TCP;UDP
+	Protocol string `json:"protocol"`
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: any value not matching the enum is rejected regardless of length, no? And non-enum string fields in this CRD (Name, Namespace, etc) already have MaxLength markers where appropriate

maas-controller/api/maas/v1alpha1/aigateway_types.go
```diff
@@ -0,0 +1,227 @@
...
+	// Name is the subject name.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Namespace is required only for ServiceAccount subjects.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=63
+	Namespace string `json:"namespace,omitempty"`
+}
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: The controller already validates this at runtime in rbacSubjects()

And the suggested "namespace is only allowed for ServiceAccount subjects" is incorrect - Kubernetes RBAC allows specifying namespace on any subject kind (it's simply ignored for User/Group)

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,762 @@
...
+	if len(name) <= 63 {
+		return name
+	}
+	sum := sha256.Sum256([]byte(aigwName))
+	hash := hex.EncodeToString(sum[:])[:8]
+	budget := 63 - len(prefix) - len(suffix) - len(hash) - 2
+	if budget < 1 {
+		return prefix + hash + "-" + suffix
+	}
+	trimmed := strings.Trim(aigwName[:budget], "-.")
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: Consulted with AI here, seems valid:

"""
This panic is mathematically impossible. The function only reaches line 746 when len(name) > 63 (line 737), which means len("aigateway-" + aigwName + "-" + suffix) > 63, so len(aigwName) > 52 - len(suffix). Meanwhile budget = 63 - 10 - len(suffix) - 8 - 2 = 43 - len(suffix). Since 52 - len(suffix) > 43 - len(suffix), we always have len(aigwName) > budget + 9, guaranteeing the slice is safe. The budget < 1 guard on line 743 handles the remaining edge case.
"""

maas-controller/pkg/controller/maas/aitenant_controller_test.go
```diff
@@ -0,0 +1,908 @@
...
+
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "openshift-ingress", Name: "team-del"}, &gatewayapiv1.Gateway{}))).To(BeTrue())
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "team-del-maas", Name: maasv1alpha1.TenantInstanceName}, &maasv1alpha1.Tenant{}))).To(BeTrue())
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "team-del-maas", Name: tenantAdminRoleName(aigw)}, &rbacv1.Role{}))).To(BeTrue())
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "team-del-maas", Name: tenantAdminRoleName(aigw)}, &rbacv1.RoleBinding{}))).To(BeTrue())
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "ai-gateway-system", Name: aigatewayAccessRoleName(aigw)}, &rbacv1.Role{}))).To(BeTrue())
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, client.ObjectKey{Namespace: "ai-gateway-system", Name: aigatewayAccessRoleName(aigw)}, &rbacv1.RoleBinding{}))).To(BeTrue())
+
+	g.Expect(apierrors.IsNotFound(cl.Get(ctx, key, &maasv1alpha1.AIGateway{}))).To(BeTrue())
+}
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: Fixed

maas-controller/pkg/controller/maas/aigateway_controller_test.go
```diff
@@ -0,0 +1,908 @@
...
+	g.Expect(cl.Get(context.Background(), client.ObjectKey{Name: tenantAdminRoleName(aigw), Namespace: "team-a-maas"}, &tenantBinding)).To(Succeed())
+	g.Expect(tenantBinding.Subjects).To(ContainElement(rbacv1.Subject{
+		Kind:     rbacv1.GroupKind,
+		APIGroup: rbacv1.GroupName,
+		Name:     "team-a-admins",
+	}))
+
+	var aigwBinding rbacv1.RoleBinding
+	g.Expect(cl.Get(context.Background(), client.ObjectKey{Name: aigatewayAccessRoleName(aigw), Namespace: "ai-gateway-system"}, &aigwBinding)).To(Succeed())
+	g.Expect(aigwBinding.RoleRef.Name).To(Equal(aigatewayAccessRoleName(aigw)))
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: Fixed. Added assertions verifying both tenantRole and aigwRole rules contain no wildcards (*)

docs/content/architecture-internals/aigateway-tenant-bootstrap.md
```diff
@@ -0,0 +1,66 @@
...
+
+When an `AIGateway` is reconciled, `maas-controller` creates or updates:
+
+1. the tenant namespace, unless `spec.tenantNamespace.create=false`
+2. the tenant `Gateway`
+3. `Tenant/default-tenant` in the tenant namespace
+4. tenant-admin RBAC in the tenant namespace
+5. per-object RBAC for the specific `AIGateway` in the infra namespace
+
+`spec.tenantNamespace.name` and `spec.gateway.namespace` are immutable in the CRD schema. The controller also rejects `AIGateway` objects in the protected ODH application namespace and in the legacy tenant namespace so bootstrap objects stay in a separate infra namespace.
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: Updated the wording

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,744 @@
...
+		}
+	}
+	return ref
+}
+
+func gatewayListenersFor(aigw *maasv1alpha1.AIGateway) []gatewayapiv1.Listener {
+	fromAll := gatewayapiv1.NamespacesFromAll
+	routes := &gatewayapiv1.AllowedRoutes{
+		Namespaces: &gatewayapiv1.RouteNamespaces{From: &fromAll},
+	}
```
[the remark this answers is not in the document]
dmytro-zaharnytskyi: seems like wrong as per ADR:
"""
Namespace Flexibility: Users can deploy models in any namespace. They may use separate namespaces per tenant, shared namespace for multiple tenants, per-project or per-team namespaces.
"""

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,744 @@
...
+		if !ok {
+			return fmt.Errorf("expected Role, got %T", obj)
+		}
+		applyAIGatewayMetadata(role, aigw)
+		role.Rules = []rbacv1.PolicyRule{
+			{
+				APIGroups:     []string{maasv1alpha1.GroupVersion.Group},
+				Resources:     []string{"aigateways"},
+				ResourceNames: []string{aigw.Name},
+				Verbs:         []string{"get", "update", "patch"},
```
jrhyness: Code Rabbit pointed this one out.   `update` and `patch` should be removed.  There isn't any reason to have them, is there?

maas-controller/api/maas/v1alpha1/aigateway_types.go
```diff
@@ -0,0 +1,196 @@
+/*
+Copyright 2025.
```
jrhyness: 2026

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,744 @@
+/*
+Copyright 2025.
```
jrhyness: 2026

jrhyness (commented):
Looking good.   It would be nice to get a decision on the naming, but it can be changed later. 


There are still some Code Rabbit comments that haven't been addressed despite everything showing resolved.   The first one, especially is important. 



  1. RBAC privilege escalation (aigateway_controller.go:462)
  - Finding: Change verbs from ["get", "update", "patch"] to read-only
  - Status: Still grants write permissions
  - Current: Verbs: []string{"get", "update", "patch"}
  - Should be: Verbs: []string{"get"}
  - Severity: HIGH - allows tenant admins to modify platform bootstrap CR

  2. AllowedRoutes namespace restriction (aigateway_controller.go:315-317)
  - Finding: Change from NamespacesFromAll to NamespacesFromSelector matching tenant namespace
  - Status: Still allows all namespaces
  - Current: fromAll := gatewayapiv1.NamespacesFromAll used for all listeners
  - Should: Use selector matching LabelTenantNamespace with value from aigw.Spec.TenantNamespace.Name
  - Severity: MEDIUM - allows route attachment from any namespace instead of just the tenant namespace

  3. Legacy tenant namespace validation (aigateway_controller.go:183-189)
  - Finding: Reject when spec.tenantNamespace.name == r.TenantNamespace
  - Status: Missing validation
  - Current: Only checks if aigw.Namespace == r.TenantNamespace (line 177)
  - Should also: Check aigw.Spec.TenantNamespace.Name == r.TenantNamespace
  - Severity: MEDIUM - could allow writing tenant resources into legacy shared namespace

See https://github.com/opendatahub-io/models-as-a-service/pull/933#pullrequestreview-4376121576

dmytro-zaharnytskyi:
> Looking good. It would be nice to get a decision on the naming, but it can be changed later.
> 
> There are still some Code Rabbit comments that haven't been addressed despite everything showing resolved. The first one, especially is important.
> 
> 1. RBAC privilege escalation (aigateway_controller.go:462)
> 
> * Finding: Change verbs from ["get", "update", "patch"] to read-only
> * Status: Still grants write permissions
> * Current: Verbs: []string{"get", "update", "patch"}
> * Should be: Verbs: []string{"get"}
> * Severity: HIGH - allows tenant admins to modify platform bootstrap CR
> 
> 2. AllowedRoutes namespace restriction (aigateway_controller.go:315-317)
> 
> * Finding: Change from NamespacesFromAll to NamespacesFromSelector matching tenant namespace
> * Status: Still allows all namespaces
> * Current: fromAll := gatewayapiv1.NamespacesFromAll used for all listeners
> * Should: Use selector matching LabelTenantNamespace with value from aigw.Spec.TenantNamespace.Name
> * Severity: MEDIUM - allows route attachment from any namespace instead of just the tenant namespace
> 
> 3. Legacy tenant namespace validation (aigateway_controller.go:183-189)
> 
> * Finding: Reject when spec.tenantNamespace.name == r.TenantNamespace
> * Status: Missing validation
> * Current: Only checks if aigw.Namespace == r.TenantNamespace (line 177)
> * Should also: Check aigw.Spec.TenantNamespace.Name == r.TenantNamespace
> * Severity: MEDIUM - could allow writing tenant resources into legacy shared namespace
> 
> See [#933 (review)](https://github.com/opendatahub-io/models-as-a-service/pull/933#pullrequestreview-4376121576)

Regarding resolved CodeRabbit comments - i left answers in every comment I left without addressing, so no blind resolving here :))

Regarding your points, agree with 1 and 3 - will fix. But 2nd point i think we should rely on adr here, we have "Users can deploy models in any namespace." in it and considering that kserve creates httpRoutes in the model's namespace (not the tenant namespace), and those routes must attach to the tenant Gateway. Restricting to NamespacesFromSelector matching only the tenant namespace would break model traffic. Correct me if I'm wrong

jrhyness:
> Regarding resolved CodeRabbit comments - i left answers in every comment I left without addressing, so no blind resolving here :))

I think it was just the "Some comments are outside the diff and can’t be posted inline due to platform limitations." bit that left it looking like everything was resolved.  

I agree with you on number 2.

jrhyness:
As agreed, let's hold off for a couple days for a naming decision.   Otherwise, it looks good to me.

maas-controller/pkg/platform/tenantreconcile/apply.go
```diff
@@ -43,6 +44,16 @@ func ApplyRendered(ctx context.Context, c client.Client, scheme *runtime.Scheme,
 			continue
 		}
 
+		// Skip resources whose live cluster copy is already owned by a different
+		// controller (e.g. ODH operator's ModelsAsService component). SSA-applying
+		// over them would fail on immutable fields like spec.selector and produce
+		// conflicting controller:true ownerReferences.
```
jrhyness, line 50: What resources does this apply to?    What failed in testing?  

Is this just for the default/models-as-a-service tenant?   Do we need to transfer ownership of that?

dmytro-zaharnytskyi: What resources /  failed? 

When deploying with --deployment-mode operator, the ODH maas component creates maas-api and its supporting resources (SA, Clusterrole, etc) -and they all come with controller:true ownerRef. -> The Tenant reconciler then tries to SSA-apply the same set with its own ownerRef and different spec.selector labels. k8s rejects both the dual controller ownerRef and the immutable selector change, so the Tenant gets stuck on PlatformReconcileFailed.

Just the default tenant?
 Yep - only affects models-as-a-service/default-tenant when running alongside the operator. (In kustomize mode there's no external controller, so nothing gets skipped and behavior is unchanged. We discovered this issue when I tried to deploy with ODH, used kustomize before)

Regarding ownership transfer - I don't think it s needed. If the operator already owns a resource, we just leave it alone

jrhyness: It feels wrong.  Not the fix specifically - just having tenant ownership split between the operator and the tenant reconciler ( and keeping them in sync ).

dmytro-zaharnytskyi: yep, but it is the temp measure. when the ODH already owns the maas-api resources, the Tenant reconciler should not try to re-apply/adopt them.

The longer-term direction, as I understood, is still ADR option 2 ("Dedicated maas-api per Tenant"), where each tenant’s maas-api lifecycle is owned by the tenant/AIGateway flow instead of being split with the ODH operator. Working on this pr (https://github.com/opendatahub-io/models-as-a-service/pull/950) and it moves us toward that by adding dynamic tenant namespace discovery, but the ownership split fully goes away only once the per-tenant maas-api platform deployment/ownership work lands. And then this check should be removed.

jrhyness: Ah, ok. Thanks,

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_aitenants.yaml
```diff
@@ -0,0 +1,299 @@
...
+                      - name
+                      type: object
+                    maxItems: 128
+                    type: array
+                type: object
+              tenantNamespace:
+                description: TenantNamespace identifies the namespace where tenant
+                  administrators manage MaaS objects.
+                properties:
+                  cleanupOnDelete:
```
jland-redhat: Is this something we want to even give the options of?

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,747 @@
...
+			Namespace: aigw.Spec.TenantNamespace.Name,
+		},
+	}
+	return r.upsert(ctx, tenant, aigw, func(obj client.Object) error {
+		t, ok := obj.(*maasv1alpha1.Tenant)
+		if !ok {
+			return fmt.Errorf("expected Tenant, got %T", obj)
+		}
+		applyAIGatewayMetadata(t, aigw)
+		t.Spec.GatewayRef = gatewayRef
```
jland-redhat: This is maybe fine for the scope of this PR but I don't know long term we want to set the Spec.GatewayRef (or Spec.ExternalOIDC) via the controller.

If these are just going to be informational for all intents and purposes then we should move this into status, since the expectation is normally fields in spec are mutable.

Could we either make this change or add a comment here to change this in the future and create a followup jira

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,747 @@
...
+	if equality.Semantic.DeepEqual(base, &existing) {
+		return ref, nil
+	}
+	if err := r.Patch(ctx, &existing, client.MergeFrom(base)); err != nil {
+		return ref, fmt.Errorf("patch Gateway %s/%s: %w", ref.Namespace, ref.Name, err)
+	}
+	return ref, nil
+}
+
+func (r *AIGatewayReconciler) mutateGateway(gw *gatewayapiv1.Gateway, aigw *maasv1alpha1.AIGateway) error {
```
jland-redhat: Wait why are we  modifying the gateway here...

We read info from the gateway but we are not making changes to the Gateway.

docs/content/architecture-internals/aigateway-tenant-bootstrap.md
```diff
@@ -0,0 +1,75 @@
+# AIGateway Tenant Bootstrap
```
jland-redhat: This does not really belong in the docs in this state.

We really want to make sure we are explaining exactly what the AIGateway is doing and how it work. 

For now lets hold off on this but lets add the new object to the CRD docs to describe the field info, and we can make sure to capture how to use it as part of the documentation PR

maas-controller/pkg/controller/maas/aigateway_controller.go
```diff
@@ -0,0 +1,747 @@
...
+		tlsMode := gatewayapiv1.TLSModeTerminate
+		return []gatewayapiv1.Listener{{
+			Name:          "https",
+			Port:          443,
+			Protocol:      gatewayapiv1.HTTPSProtocolType,
+			Hostname:      &hostname,
+			AllowedRoutes: routes,
+			TLS: &gatewayapiv1.GatewayTLSConfig{
+				Mode: &tlsMode,
+				CertificateRefs: []gatewayapiv1.SecretObjectReference{{
```
jland-redhat: Again we should not be modifying the gateway.

jland-redhat (changes requested):
Made a few comments but the biggest one here is that we should **NOT** be modifying the gateway in any way.

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -0,0 +1,647 @@
...
+	return r.upsert(ctx, tenant, aitenant, func(obj client.Object) error {
+		t, ok := obj.(*maasv1alpha1.Tenant)
+		if !ok {
+			return fmt.Errorf("expected Tenant, got %T", obj)
+		}
+		applyAITenantMetadata(t, aitenant)
+		// TODO: Move these mirrored platform values out of Tenant spec in a
+		// follow-up Jira once the MaaS config/status API is settled. The current
+		// post-render path still reads Tenant.spec.gatewayRef and externalOIDC.
+		t.Spec.GatewayRef = gatewayRef
```
jland-redhat, line 312: Did we create the jira?

jland-redhat (changes requested):
One more request please add some E2E test to validate creation/deletion of the AITenant please

test/e2e/tests/test_aitenant_lifecycle.py
```diff
@@ -0,0 +1,258 @@
...
+            "maas.opendatahub.io/aitenant-name",
+            "maas.opendatahub.io/aitenant-namespace",
+        ]
+        gateway_before = _metadata_values(
+            _oc_json(["get", "gateway", GATEWAY_NAME, "-n", GATEWAY_NAMESPACE, "-o", "json"]),
+            watched_metadata_keys,
+        )
+
+        try:
+            _create_namespace(infra_ns)
```
jland-redhat: If it is small can we add the code to create the namespace automatically out of the box (similar to how we are doing models-as-a-service currently.

test/e2e/tests/test_aitenant_lifecycle.py
```diff
@@ -0,0 +1,258 @@
...
+            assert _get_json_or_none("role", object_admin_role, infra_ns) is not None
+            assert _get_json_or_none("rolebinding", object_admin_role, infra_ns) is not None
+
+            gateway_after_create = _metadata_values(
+                _oc_json(["get", "gateway", GATEWAY_NAME, "-n", GATEWAY_NAMESPACE, "-o", "json"]),
+                watched_metadata_keys,
+            )
+            assert gateway_after_create == gateway_before, "AITenant must not mutate Gateway metadata"
+
+            _delete(AITENANT_KIND, aitenant_name, infra_ns)
```
jland-redhat: Can we make create and delete two different test

And can we add a TODO to make validate that all the basic functionality  Policies/Subscriptions/Models/Inference) works with the newly created tenant (Writing those test may be a little out of scope for this but want to have a breadcrumb so we don't forget)

jland-redhat (changes requested):

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -0,0 +1,705 @@
...
+	}
+	if ref.Name == "" {
+		return ref, errors.New("spec.gateway.name is required when AITenant name is empty")
+	}
+
+	var gateway gatewayapiv1.Gateway
+	key := client.ObjectKey{Namespace: ref.Namespace, Name: ref.Name}
+	if err := r.get(ctx, key, &gateway); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ref, fmt.Errorf("gateway %s/%s not found: create or configure the Gateway before creating AITenant %s/%s", ref.Namespace, ref.Name, aitenant.Namespace, aitenant.Name)
```
jrhyness: https://github.com/opendatahub-io/architecture-decision-records/pull/138/changes#diff-eb68695e5d6e13ec23c3f3a7560aea41e92863e0251d8dfe80bd379556fc7869R71

We're supposed to be creating the gateway?

jrhyness (commented):
@dmytro-zaharnytskyi , this isn't creating maas-api in `ai-gateway-infra` yet, right?

maas-controller/pkg/controller/maas/aitenant_controller.go
dmytro-zaharnytskyi: regarding gateway, I messed it up a bit while fixing RBAC and other stuff today's morning, will revert it soon

Regarding maas-api in ai-gateway-infra - no, it doesn't, we will implement that, but it is out of scope (for the current task)

jland-redhat (approved):
lgtm

jrhyness (approved):
lgtm

PR #896, opened 2026-05-13
fix: use odh-stable image tags in odh overlay and add dev overlay

EgorLu:
https://redhat.atlassian.net/browse/RHOAIENG-60590

## Summary

- Change `deployment/overlays/odh/params.env` image tags from `:latest` to `:odh-stable` so the ODH operator (which reads from the `stable` branch) deploys images matching the stable codebase
- Change base kustomization `newTag` values to `odh-stable` for consistency
- Create `deployment/overlays/dev/params.env` with `:latest` tags for the deploy script's `--dev` flag
- Update `maas-api/deploy/overlays/dev/` to wrap the odh overlay with `:latest` image overrides
- Add `--dev` flag to `deploy.sh` to select the dev overlay
- Document the branch-to-image-tag mapping in `docs/content/contributing/release-strategy.md`
- Update `CONTRIBUTING.md` Tekton pipeline table (outdated after PR #877)

## Test plan

- [x] `kustomize build deployment/overlays/odh` — N/A (no kustomization.yaml since #840)
- [x] `kustomize build maas-api/deploy/overlays/odh` produces `:odh-stable` tags
- [x] `kustomize build maas-api/deploy/overlays/dev` produces `:latest` tags
- [x] `diff` between maas-api odh and dev overlays shows only image tag lines differ
- [x] `./scripts/ci/validate-manifests.sh` passes for all kustomization files (tested with kustomize 5.7.1 to match CI)
- [x] `deploy.sh --help` shows `--dev` flag
- [x] `deploy.sh --dev --dry-run` sets `DEV_MODE=true`

## Summary by CodeRabbit

* **New Features**
  * Added `--dev` flag to deployment script to enable development-mode image deployments.

* **Documentation**
  * Updated CI pipeline documentation with clarified image tag conventions for different branches and overlays.
  * Added testing guide section explaining development vs. stable image selection for deployments.
  * Added release strategy documentation covering image tag conventions and overlay layout.

PR #939, opened 2026-05-25
fix: restore --dev flag in deploy script argument parser

EgorLu:
https://redhat.atlassian.net/browse/RHOAIENG-64574

## Summary

- Restore the `--dev)` case in `parse_arguments()` — removed accidentally in #864
- Update default image tag from `:latest` to `:odh-stable` when `--dev` is not passed

## Context

PR #896 added `--dev` flag support to `deploy.sh`. PR #864 (refactor of Tenant reconciler params.env) subsequently removed the `--dev)` case from `parse_arguments()` while keeping the `DEV_MODE` variable and help text. This causes `deploy.sh --dev` to fail with "Unknown option: --dev".

## Risk analysis

- **Risk rating:** 1
- **Why:** Single-file script change restoring a flag parser case that was accidentally removed. The `DEV_MODE` variable, help text, and consuming code already exist — only the argument parser entry was missing.

## Test plan

- [x] `deploy.sh --dev --dry-run` — no longer errors with "Unknown option"
- [x] `DEV_MODE` correctly set to `true` when `--dev` is passed
- [x] Default (no `--dev`) uses `:odh-stable` image tags
- [x] `--dev` switches to `:latest` image tags

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -330,6 +330,10 @@ parse_arguments() {
         POSTGRES_CONNECTION="$2"
         shift 2
         ;;
+      --dev)
+        DEV_MODE="true"
+        shift
+        ;;
       --help|-h)
         show_help
         exit 0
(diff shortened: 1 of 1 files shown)


EgorLu:
@coderabbitai review

jrhyness (approved):
lgtm

PR #951, opened 2026-06-02
feat: add a new field to the db schema for `tenant`

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-62763

## Description
- added a new migration `0004_add_tenant_column.up.sql`. Implemented with forward migration in mind and made it idempotent. Since we atomically backfill all existing rows with the empty-string. No separate backfill step needed.
- updated `types.go`. Namely, updated the `ApiKey` struct
- add `tenant` param in other files and add "" for most of the tests to fit the new method signature
**NOTE:** The tenant column is a plain `TEXT` field intentionally kept independent for now, since the exact tenant identifier format (namespace name, AIGateway name, stable slug) is still in discussion.

## How Has This Been Tested?
Additional unit tests have been added

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * API keys now support tenant association so keys can be linked to a tenant.
  * API responses and storage surface tenant information for keys.
  * Existing API keys continue to work and default to an empty tenant value.

* **Database**
  * Added migration and rollback to add/remove tenant column and index.

* **Tests**
  * Test suite updated and a new end-to-end tenant test added.


maas-api/db/schema/0004_add_tenant_column.up.sql
```diff
@@ -0,0 +1,6 @@
+-- Schema for API Key Management: 0004_add_tenant_column.up.sql
+-- Description: Add tenant column — binds each API key to a tenant identifier
+
+-- Add tenant column (idempotent). DEFAULT '' backfills pre-existing rows with the
+-- sentinel empty-string value so the migration is safe on populated tables.
+ALTER TABLE api_keys ADD COLUMN IF NOT EXISTS tenant TEXT NOT NULL DEFAULT '';
```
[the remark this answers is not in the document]
yu-teo: The user story explicitly allows empty string as the default value. Additionally, tenant format is to be determined. Baking in a magic string like `<migrated>` now creates a default value to deal with later. An empty string is a cleaner "not yet assigned" marker that the S4 follow-up can naturally replace when it wires real tenant resolution. Adding `DROP DEFAULT` would break the idempotency pattern. If the migration runs and then service.go creates keys before S4 lands, those `INSERT`s pass "" and would violate the `CHECK` constraint. The current design intentionally allows the column to be inert until S4 adds real logic.
The right time to add a CHECK (tenant <> '') constraint is in S4, after the service layer is updated to populate a real tenant value.

jrhyness:
To be clear,  S4  will add:

  - Tenant context extraction from request
  - Update all store methods to use real tenant value
  - Add AND tenant = $X to all SELECT/UPDATE/DELETE queries
  - Backfill: UPDATE api_keys SET tenant = 'models-as-a-service' WHERE tenant = ''
 
?

The index and unique constraint could be added now, I think? 

ref: 
https://github.com/opendatahub-io/architecture-decision-records/pull/138/changes#diff-eb68695e5d6e13ec23c3f3a7560aea41e92863e0251d8dfe80bd379556fc7869R315

maas-api/db/schema/0004_add_tenant_column.up.sql
ishitasequeira, line 1: Given that we are now GA, it would be useful to add rollback script (`.down.sql` file) as well for users to be able to rollback easily.

yu-teo: @ishitasequeira we never had one. Is this a good time to introduce a down migration? Wouldn't it be inconsistent and potentially misleading (users might expect rollback support for all migrations)?

yu-teo:
> The index and unique constraint could be added now, I think?
> 
> ref: https://github.com/opendatahub-io/architecture-decision-records/pull/138/changes#diff-eb68695e5d6e13ec23c3f3a7560aea41e92863e0251d8dfe80bd379556fc7869R315

I will be adding a simple index with `IF NOT EXISTS` to keep it idempotent, matching the pattern of the existing migrations. However, I am not too sure which constrain to introduce at this point in time. Copuld you help identifying it or do we leave it until work on S4 has started?

jrhyness:
> > The index and unique constraint could be added now, I think?
> > ref: https://github.com/opendatahub-io/architecture-decision-records/pull/138/changes#diff-eb68695e5d6e13ec23c3f3a7560aea41e92863e0251d8dfe80bd379556fc7869R315
> 
> I will be adding a simple index with `IF NOT EXISTS` to keep it idempotent, matching the pattern of the existing migrations. However, I am not too sure which constrain to introduce at this point in time. Copuld you help identifying it or do we leave it until work on S4 has started?

I was referring to the key_hash unique constraint in the ADR.

```
-- Add unique constraint: key hash must be unique within tenant
-- (allows same key pattern across tenants if needed, though not recommended)
ALTER TABLE api_keys ADD CONSTRAINT uk_api_keys_hash_tenant UNIQUE (key_hash, tenant_id);
```

We currently have a unique index which is more strict.  

```
CREATE UNIQUE INDEX IF NOT EXISTS idx_api_keys_key_hash ON api_keys(key_hash);
```

I suggest we _not_ change the uniqueness constraint.  I'll confirm with Marius in his PR.

maas-api/db/schema/0004_add_tenant_column.up.sql
yu-teo: Discussed on slack, committed a change adding a rollback functionality, resolving.

jrhyness (approved):
lgtm

ishitasequeira (approved):
lgtm

PR #953, opened 2026-06-02
fix: revert "feat: move external cronjob to run timely cleanup to goroutin…

jland-redhat:
Reverting changes for https://github.com/opendatahub-io/models-as-a-service/pull/934/changes


ishitasequeira (approved):
/lgtm

PR #954, opened 2026-06-03
feat: move external cronjob to run timely cleanup to goroutine, part II

zdtsw:
## Description

Completes the migration from the external `maas-api-key-cleanup` CronJob to the in-process cleanup loop in maas-api (follow-up to #934). 

in detail :
- Removes the dead CronJob
- Cleanup: params, RBAC, deploy script, controller patch code
- Adds general function `PruneLegacyCleanupResources` so the orphaned CronJob + `maas-api-cleanup-restrict`
NetworkPolicy are deleted on Tenant reconcile
- run CI test on changes not only maas-controller path
- renamed const to make it clear the same image is used for Tenent not Cleanup, but kept the key in params.env old name to avoid change in ai-gateway-opreator( probalby good to do the change as well, but can be a followup later)

Co-authored-by  @jland-redhat

## How Has This Been Tested?

run 
```
cd maas-controller; make test
cd maas-api; make test

```
## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Ephemeral API key cleanup now runs inside **maas-api** as a background process, configurable via `CLEANUP_INTERVAL_MINUTES` (default: 15; `-1` disables).
* **Chores**
  * Removed the standalone key-cleanup CronJob and restrictive cleanup NetworkPolicy; upgrades now prune legacy cleanup resources after MaaS API becomes ready.
  * Updated controller/RBAC and CI workflow behavior to run for all pull requests.
* **Documentation**
  * Updated API key administration, internal endpoint, and troubleshooting guidance for the in-process cleanup.
* **Tests**
  * E2E/unit tests updated to validate `CLEANUP_INTERVAL_MINUTES` and in-process cleanup; removed CronJob/NetworkPolicy checks.


zdtsw:
@jland-redhat to follow up: do we want to preceed with this PR now?

jland-redhat:
I still want to get this in the goal will be for 3.6-EA1

zdtsw:
> I still want to get this in the goal will be for 3.6-EA1

ok, cool, let me get the code in shape, and we can revisit in a while.

zdtsw:
@Jaland / @ishitasequeira check again if we still wanna this into 3.6 ea1 cycle

PR #962, opened 2026-06-03
feat: add source field to ModelRefInfo in subscription response

liangwen12year:
## Description
Derive internal/external classification from MaaSModelRef.spec.modelRef.kind (ExternalModel → external, LLMInferenceService → internal) and populate it in the enrichModelRefs step. Updates OpenAPI spec accordingly.

Ref: [RHOAIENG-62495](https://redhat.atlassian.net/browse/RHOAIENG-62495)

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Model references in API responses now include a new optional "source" property with values "internal" or "external".
* **Documentation**
  * API schema updated to document the new "source" field and its allowed values.
* **Tests**
  * New tests validate that the "source" field is populated correctly for internal, external, and unknown model kinds.

### DIFF
--- a/maas-api/internal/subscription/selector.go
+++ b/maas-api/internal/subscription/selector.go
@@ -298,9 +298,10 @@ func (s *Selector) enrichModelRefs(refs []ModelRefInfo, index map[string]*unstru
 				refs[i].Description = annotations[constant.AnnotationDescription]
 			}
 			kind, _, _ := unstructured.NestedString(u.Object, "spec", "modelRef", "kind")
-			if kind == "ExternalModel" {
+			switch kind {
+			case "ExternalModel":
 				refs[i].Source = "external"
-			} else if kind != "" {
+			case "LLMInferenceService":
 				refs[i].Source = "internal"
 			}
 		}
--- a/maas-api/internal/subscription/selector.go
+++ b/maas-api/internal/subscription/selector.go
@@ -297,6 +297,12 @@ func (s *Selector) enrichModelRefs(refs []ModelRefInfo, index map[string]*unstru
 				refs[i].DisplayName = annotations[constant.AnnotationDisplayName]
 				refs[i].Description = annotations[constant.AnnotationDescription]
 			}
+			kind, _, _ := unstructured.NestedString(u.Object, "spec", "modelRef", "kind")
+			if kind == "ExternalModel" {
+				refs[i].Source = "external"
+			} else if kind != "" {
+				refs[i].Source = "internal"
+			}
 		}
 	}
 }
--- a/maas-api/internal/subscription/selector_test.go
+++ b/maas-api/internal/subscription/selector_test.go
@@ -883,3 +883,86 @@ func createSubscriptionWithTRLPStatus(name string, groups []string, phase string
 	}
 	return obj
 }
+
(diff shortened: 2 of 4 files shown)


maas-api/internal/subscription/selector.go
```diff
@@ -297,6 +297,12 @@ func (s *Selector) enrichModelRefs(refs []ModelRefInfo, index map[string]*unstru
 				refs[i].DisplayName = annotations[constant.AnnotationDisplayName]
 				refs[i].Description = annotations[constant.AnnotationDescription]
 			}
+			kind, _, _ := unstructured.NestedString(u.Object, "spec", "modelRef", "kind")
+			if kind == "ExternalModel" {
+				refs[i].Source = "external"
+			} else if kind != "" {
+				refs[i].Source = "internal"
```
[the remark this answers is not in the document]
chaitanya1731: I agree with this comment here. We anyway have two kinds right now `LLMInferenceService` and `ExternalModel`, so tis should work in theory.. But if a new kind is added future then this would misclassify it.. May be we can either add another `else`?

maas-api/internal/subscription/selector.go
```diff
@@ -297,6 +297,12 @@ func (s *Selector) enrichModelRefs(refs []ModelRefInfo, index map[string]*unstru
...
+			}
```
chaitanya1731:
```suggestion
			} else if kind == "LLMInferenceService" {
				refs[i].Source = "internal"
			} else {
			    refs[i].Source = "unknown"
			}
```

chaitanya1731: This is just a suggestion. Not sure if this will break something.

maas-api/internal/subscription/selector.go
```diff
@@ -297,6 +297,12 @@ func (s *Selector) enrichModelRefs(refs []ModelRefInfo, index map[string]*unstru
...
 		}
 	}
```
chaitanya1731, line 308: Can we also add some unit tests in `selector_test.go` covering the `enrichModelRefs` function or the new Source field. Given the PR adds classification logic with branching, a test should verify:
ExternalModel kind → source = "external"
LLMInferenceService kind → source = "internal"
Empty kind → source is empty/omitted/unknown..

maas-api/internal/subscription/selector.go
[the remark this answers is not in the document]
liangwen12year: Good point. Updated to use an explicit `switch` on `ExternalModel` and `LLMInferenceService` — unknown kinds now leave `source` empty (omitted from JSON via `omitempty`) rather than being misclassified.

maas-api/internal/subscription/selector.go
liangwen12year: Added `TestEnrichModelRefsSource` in `selector_test.go` with 4 test cases:
- `ExternalModel` → `"external"`
- `LLMInferenceService` → `"internal"`
- Empty kind → empty (omitted)
- Unknown kind (`SomeNewKind`) → empty (omitted)

All passing.

chaitanya1731 (approved):
lgtm

jland-redhat (approved):
lgtm

jland-redhat:
lgtm

EgorLu (approved):

PR #966, opened 2026-06-05
fix: stop subscription description falling back to displayName

liangwen12year:
## Description
Remove the `description→displayName→name` fallback chain in `toSubscriptionInfo` and `ResponseToSubscriptionInfo`. When no description annotation is set, return empty string and let the frontend decide what to display. 

Resolves [RHOAIENG-66621](https://redhat.atlassian.net/browse/RHOAIENG-66621)

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription descriptions now reflect only the `openshift.io/description` annotation and no longer fall back to other fields when it’s missing.
* **Documentation**
  * Updated OpenAPI documentation: `subscription_description` may be empty and is no longer marked as required.
* **Tests**
  * Adjusted subscription listing tests to validate the new “no fallback” behavior, including renamed/repurposed coverage.
* **Chores**
  * Updated OpenAPI diff validation to acknowledge the documented API change.

### DIFF
--- a/maas-api/internal/subscription/handler_test.go
+++ b/maas-api/internal/subscription/handler_test.go
@@ -1061,8 +1061,8 @@ func TestListSubscriptions_MultipleAccessible(t *testing.T) {
 		found[s.SubscriptionIDHeader] = s.SubscriptionDescription
 	}
 
-	if desc, ok := found["free-sub"]; !ok || desc != "Free Tier" {
-		t.Errorf("expected free-sub with description 'Free Tier' (fallback from display-name), got %q", desc)
+	if desc, ok := found["free-sub"]; !ok || desc != "" {
+		t.Errorf("expected free-sub with empty description (no fallback), got %q", desc)
 	}
 	if desc, ok := found["premium-sub"]; !ok || desc != "High limits for production" {
 		t.Errorf("expected premium-sub with description 'High limits for production', got %q", desc)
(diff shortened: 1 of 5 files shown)


maas-api/internal/subscription/handler_test.go
chaitanya1731: One nit..  The test name now `TestListSubscriptions_DescriptionFallback` is now misleading, as it asserts that fallback does not occur... Can we rename it appropriately to accurately reflect the new test behavior.

chaitanya1731 (commented):
I think this will also need changes in the `openapi3.yaml` spec here - https://github.com/opendatahub-io/models-as-a-service/blob/94edca0023b5e4c0f2c12e73506cb3cdf59eff78/maas-api/openapi3.yaml#L794 
 
Itis also marked as required field. If we are ok to return empty or `""` then we may need to update it in the `required`?  it may be worth noting in the spec that the value can be an empty string, so consumers don't assume it will always be meaningful.
https://github.com/opendatahub-io/models-as-a-service/blob/94edca0023b5e4c0f2c12e73506cb3cdf59eff78/maas-api/openapi3.yaml#L825

maas-api/internal/subscription/handler_test.go
liangwen12year: Good catch — renamed to `TestListSubscriptions_DescriptionNoFallback`.

liangwen12year: Renamed to `TestListSubscriptions_DescriptionNoFallback` in 9d242a7.

liangwen12year:
@chaitanya1731 Good catches — both addressed in 9d242a7:

1. **OpenAPI description updated**: removed the fallback wording, now reads: *"Human-readable description from the openshift.io/description annotation. May be empty if not set."*
2. **Removed from `required`**: `subscription_description` is no longer listed as required since it can legitimately be an empty string.
3. **Test renamed**: `TestListSubscriptions_DescriptionFallback` → `TestListSubscriptions_DescriptionNoFallback`.

chaitanya1731 (approved):
lgtm

chaitanya1731 (approved):
lgmt

ishitasequeira (commented):
Overall change looks good. CI needs to be fixed.

chaitanya1731 (approved):
lgtm

ishitasequeira (approved):
lgtm

PR #995, opened 2026-06-14
feat: add Loki-based usage dashboards with Perses datasources

tgitelman:
## Summary

Add Perses dashboards and Loki datasources for usage monitoring, providing both admin-wide and user-scoped views of model inference activity.

### Admin dashboard (`usage-admin-loki-dashboard`)
- Direct `loki` datasource (SA token auth + TLS to LokiStack gateway)
- Dropdowns: User, Subscription, Model (`LokiLabelValuesVariable`, COO 1.5+)
- Panels: Total tokens, Total requests, Total rate limited, Success rate, Active users
- Token consumption time-series chart (stacked area, `view_by` toggle: model/subscription)
- Usage breakdown table grouped by model/subscription/key_name (`MergeSeries` transform)
- `customAllValue: ".*"` on ListVariables for absent-label matching

### User dashboard (`usage-user-loki-dashboard`)
- `scoped-loki` datasource routed through `loki-query-proxy` — injects `user_id` filtering based on the caller's Kubernetes token (TokenReview)
- No User dropdown — queries are automatically scoped to the logged-in user
- Same panels as admin minus Active users; wider 4-panel grid layout
- Datasource named `scoped-loki` to avoid Perses prefix collision with `loki`

### Files changed (6)

| File | Change |
|------|--------|
| `dashboards/kustomization.yaml` | Add user dashboard + scoped datasource resources |
| `dashboards/loki-datasource-scoped.yaml` | New: `scoped-loki` datasource → `loki-query-proxy-user` |
| `dashboards/usage-user-loki-dashboard.yaml` | New: user-scoped Perses dashboard |
| `usage-logs/kustomization.yaml` | New: admin dashboard kustomization |
| `usage-logs/loki-datasource.yaml` | New: admin `loki` datasource (direct to LokiStack) |
| `usage-logs/usage-admin-loki-dashboard.yaml` | New: admin Perses dashboard |

### Dependencies

- OTel pipeline deployed (EnvoyFilter → OTel Collector → Loki) for log data
- Loki query proxy (PR #999) for user-scoped dashboard
- LokiStack with `streamLabels.resourceAttributes` configured

## Risk analysis

- **Risk rating**: 1
- **Why**: Dashboard and datasource manifests only — no controller code, no runtime behavior changes. Covered by manual verification on cluster. Risk is limited to dashboard query correctness.

### DIFF
--- a/deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
+++ b/deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
@@ -1,0 +1,478 @@
+# Admin usage dashboard — queries Loki structured logs via LogQL.
+# Uses $__range to bind LogQL windows to the native time picker.
+# LokiLabelValuesVariable for model/subscription dropdowns (COO 1.5+).
+# CEL filter restricts logs to inference-only paths; no probe traffic to filter.
+# customAllValue: ".*" on ListVariables ensures "All" matches entries with
+# absent labels (e.g. user_id when capturing is disabled). Do not remove.
+apiVersion: perses.dev/v1alpha1
+kind: PersesDashboard
+metadata:
+  name: dashboard-4-maas-usage-logs-admin
+  namespace: opendatahub
+  labels:
+    app.kubernetes.io/managed-by: maas-controller
+    app.kubernetes.io/component: perses
+spec:
+  display:
+    name: "All usage (logs)"
+    description: >-
+      Admin view of model usage. Shows token consumption, request counts,
+      rate limiting, and per-user breakdown across all users and subscriptions.
+  duration: 1h
+  variables:
+    - kind: ListVariable
+      spec:
+        name: user
+        display:
+          name: "User"
+        allowMultiple: true
+        allowAllValue: true
+        customAllValue: ".*"
+        defaultValue: "$__all"
+        plugin:
+          kind: LokiLabelValuesVariable
+          spec:
+            datasource:
+              kind: LokiDatasource
+              name: usage-logs
+            labelName: user_id
+            matchers:
+              - '{service_name="maas-gateway"}'
+    - kind: ListVariable
+      spec:
+        name: subscription
+        display:
+          name: "Subscription"
+        allowMultiple: true
+        allowAllValue: true
+        customAllValue: ".*"
+        defaultValue: "$__all"
+        plugin:
+          kind: LokiLabelValuesVariable
+          spec:
+            datasource:
+              kind: LokiDatasource
+              name: usage-logs
+            labelName: subscription
+            matchers:
+              - '{service_name="maas-gateway", subscription!="-"}'
+    - kind: ListVariable
+      spec:
+        name: model
+        display:
+          name: "Model"
+        allowMultiple: true
+        allowAllValue: true
+        customAllValue: ".*"
+        defaultValue: "$__all"
+        plugin:
+          kind: LokiLabelValuesVariable
+          spec:
+            datasource:
+              kind: LokiDatasource
+              name: usage-logs
+            labelName: model
+            matchers:
+              - '{service_name="maas-gateway", model!="-"}'
+    - kind: ListVariable
+      spec:
+        name: view_by
+        display:
+          name: "View by"
+        allowMultiple: false
+        allowAllValue: false
+        defaultValue: "model"
+        plugin:
+          kind: StaticListVariable
+          spec:
+            values:
+              - value: "model"
+                label: By model
+              - value: "subscription"
+                label: By subscription
+
+  panels:
+    totalTokens:
+      kind: Panel
+      spec:
+        display:
+          name: "Total tokens"
+          description: "Total tokens consumed over the selected time window."
+        plugin:
+          kind: StatChart
+          spec:
+            calculation: last
+            format:
+              unit: decimal
+              shortValues: true
+              decimalPlaces: 0
+        queries:
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs
+                  query: >-
+                    sum(sum_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model", response_type="hit"}
+                    | user_id=~"$user"
+                    | unwrap tokens_total [$__range]))
+                    or vector(0)
+
+    totalRequests:
+      kind: Panel
+      spec:
+        display:
+          name: "Total requests"
+          description: "Total API requests over the selected time window."
+        plugin:
+          kind: StatChart
+          spec:
+            calculation: last
+            format:
+              unit: decimal
+              shortValues: true
+              decimalPlaces: 0
+        queries:
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs
+                  query: >-
+                    sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model"}
+                    | user_id=~"$user" [$__range]))
+                    or vector(0)
+
+    totalRateLimited:
+      kind: Panel
+      spec:
+        display:
+          name: "Total rate limited"
+          description: "Total rate-limited requests over the selected time window."
+        plugin:
+          kind: StatChart
+          spec:
+            calculation: last
+            format:
+              unit: decimal
+              shortValues: true
+              decimalPlaces: 0
+        queries:
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs
+                  query: >-
+                    sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model", response_type="rate_limit"}
+                    | user_id=~"$user" [$__range]))
+                    or vector(0)
+
+    successRate:
+      kind: Panel
+      spec:
+        display:
+          description: "Percentage of authorized vs total requests over the selected time window."
+          name: Success rate
+        plugin:
+          kind: StatChart
+          spec:
+            calculation: last
+            format:
+              decimalPlaces: 1
+              unit: percent-decimal
+        queries:
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs
+                  query: >-
+                    (sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model", response_type="hit"}
+                    | user_id=~"$user" [$__range]))
+                    /
+                    sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model"}
+                    | user_id=~"$user" [$__range])))
+                    or vector(1)
+
+    activeUsers:
+      kind: Panel
+      spec:
+        display:
+          name: "Active users"
+          description: "Distinct users with API activity over the selected time window."
+        plugin:
+          kind: StatChart
+          spec:
+            calculation: last
+            format:
+              unit: decimal
+              decimalPlaces: 0
+        queries:
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs
+                  query: >-
+                    count(sum by (user_id) (count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription", model=~"$model"}
+                    | user_id=~"$user" | user_id!="" | user_id!="-" [$__range])))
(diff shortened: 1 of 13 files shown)


deployment/components/observability/loki/loki-ca.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: loki-gateway-serving-ca
+  namespace: opendatahub
+  labels:
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: perses
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
```
ahadas: did you check if we can enable unencrypted communication from within the cluster? I saw we have unencrypted communication with the Prometheus instance of ODH - I think it makes sense here as well, if possible

deployment/components/observability/loki/loki-datasource.yaml
```diff
@@ -0,0 +1,36 @@
...
+    display:
+      name: "Loki (Admin)"
+    default: true
+    plugin:
+      kind: "LokiDatasource"
+      spec:
+        proxy:
+          kind: HTTPProxy
+          spec:
+            url: https://lokistack-sample-gateway-http.logging4.svc.cluster.local:8080/api/logs/v1/application
```
ahadas: this should not rely on the namespace being `logging4`

deployment/components/observability/loki/loki-rbac.yaml
```diff
@@ -0,0 +1,18 @@
+# Grant Perses SA read access to application logs via LokiStack gateway.
```
ahadas: this probably needs to move to the opendatahub-operator instead

deployment/components/observability/loki/loki-secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: loki-secret
+  namespace: opendatahub
+  labels:
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: perses
+  annotations:
+    kubernetes.io/service-account.name: loki-query-proxy
```
ahadas: I guess we don't need this here, right?

deployment/components/observability/observability/dashboards/loki-datasource.yaml
```diff
@@ -0,0 +1,36 @@
+# Admin Loki datasource — direct to LokiStack, no user filtering.
```
ahadas: the datasource needs to be placed in observability/dashboards folder, next to the dashboard

deployment/components/observability/observability/dashboards/usage-admin-loki-dashboard.yaml
```diff
@@ -0,0 +1,401 @@
...
+        display:
+          name: "Subscription"
+        allowMultiple: true
+        allowAllValue: true
+        defaultValue: "$__all"
+        plugin:
+          kind: StaticListVariable
+          spec:
+            values:
+              - "dev-subscription-a"
```
ahadas: version 1.5 of cluster observability operator is expected later this week, we need to test their fix for dynamic values in those lists

deployment/components/observability/loki/loki-ca.yaml
tgitelman: LokiStack gateway in openshift-logging mode enforces HTTPS exclusively ([PR #6288](https://github.com/grafana/loki/pull/6288)). There's no HTTP port to connect to. This matches the Prometheus/Thanos Querier pattern — the ODH operator's own cluster-prometheus-datasource also uses HTTPS + service CA + bearer token to Thanos Querier (same pattern we follow here for Loki). The CA ConfigMap + TLS config are required.

ahadas: > This matches the Prometheus/Thanos Querier pattern — the ODH operator's own cluster-prometheus-datasource also uses HTTPS + service CA + bearer token to Thanos Querier (same pattern we follow here for Loki)

the Prometheus instance I referred to is not the cluster-prometheus-datasource but the one that is specific to ODH

tgitelman: The ODH-specific Prometheus instance (prometheus-odh-monitoring) is deployed by the ODH operator itself and exposes HTTP (port 9090) without TLS — the operator controls its own deployment. LokiStack is different: it's deployed by the Loki Operator (CLO), and the gateway component enforces HTTPS + mTLS by default with no option to disable it in openshift-logging mode. There's no HTTP port we can connect to. If this changes upstream (e.g., LokiStack adds an option for in-cluster HTTP), we can drop the CA ConfigMap and simplify.

deployment/components/observability/loki/loki-datasource.yaml
tgitelman: Fixed — URL updated to logging-loki-gateway-http.openshift-logging.svc.cluster.local (standard LokiStack namespace).

deployment/components/observability/loki/loki-rbac.yaml
tgitelman: Agreed — removed. The ClusterRoleBinding (perses-sa → cluster-logging-application-view) is platform-level RBAC, not MaaS-specific. It should be provisioned by the opendatahub-operator when it enables Loki integration for Perses. Removed loki-rbac.yaml, loki-secret.yaml, and loki-ca.yaml from this PR — the dashboard and datasource CRs remain (MaaS-owned), the infra goes to the operator.

deployment/components/observability/loki/loki-secret.yaml
tgitelman: Correct — removed. Same reasoning as #3: the SA token secret for LokiStack authentication is platform infra. The opendatahub-operator should provision this alongside the RBAC when setting up Perses with Loki access. The datasource CR still references the secret by name (secret: "loki-secret"), but the actual resource is operator-owned.

deployment/components/observability/observability/dashboards/loki-datasource.yaml
tgitelman: Done — loki-datasource.yaml moved to observability/dashboards/ next to the dashboard, following the same co-location pattern as kuadrant-prometheus-datasource.yaml. The entire loki/ directory has been removed from this PR.

deployment/components/observability/observability/dashboards/usage-admin-loki-dashboard.yaml
tgitelman: Done — COO 1.5.0 shipped with LokiLabelValuesVariable plugin ([perses/plugins#651](https://github.com/perses/plugins/pull/651)). All three variables (user, subscription, model) now use LokiLabelValuesVariable with dynamic label discovery from Loki. Tested and verified on cluster — dropdowns are dynamically populated and searchable.

deployment/components/observability/loki/loki-ca.yaml
ahadas: ok

deployment/components/observability/observability/dashboards/loki-datasource.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: loki
+  namespace: opendatahub
+  labels:
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: perses
+spec:
+  config:
+    display:
+      name: "Loki (Admin)"
```
ahadas: that's not a good name to display
```suggestion
      name: "Usage"
```
and the existing dashboard could change to something like `Usage (metrics)` or `Usage (legacy)`
or maybe we can just keep it as `Usage` and show either the logs-based or the metrics-based dashboard (this requires some changes in the controller) - wdyt?

deployment/components/observability/observability/dashboards/loki-datasource.yaml
```diff
@@ -0,0 +1,37 @@
...
+    default: true
+    plugin:
+      kind: "LokiDatasource"
+      spec:
+        proxy:
+          kind: HTTPProxy
+          spec:
+            # LokiStack gateway URL — adjust name/namespace to match your LokiStack CR.
+            url: https://logging-loki-gateway-http.openshift-logging.svc.cluster.local:8080/api/logs/v1/application
```
ahadas: in your comment on the previous thread you mentioned that you pick this URL since it's a standard location - but we rather want to use the ODH-specific instance so we can assume it's deployed in the monitoring namespace 
```suggestion
            url: https://logging-loki-gateway-http.opendatahub.svc.cluster.local:8080/api/logs/v1/application
```
we just need to make sure we change `opendatahub` to the setting in DSCI

deployment/components/observability/observability/dashboards/usage-admin-loki-dashboard.yaml
```diff
@@ -0,0 +1,433 @@
...
+  name: usage-admin-loki-dashboard
+  namespace: opendatahub
+  labels:
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: perses
+spec:
+  display:
+    name: "Usage Admin Dashboard"
+    description: >-
+      Admin view of MaaS API usage. Shows token consumption, request counts,
```
ahadas: it's not really MaaS API usage but rather model usage

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -0,0 +1,433 @@
...
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: loki
+                  query: >-
+                    sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription"}
+                    | user_id=~"$user" | user_id!="-" [$__range]))
+                    or vector(0)
```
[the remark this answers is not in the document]
ahadas: yeah, we need to include the model here - either from the response or from the request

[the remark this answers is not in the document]
tgitelman: Fixed. Added model=~"$model" to all stat panels: totalRequests, totalRateLimited, successRate (both numerator and denominator), and activeUsers.

Note: CodeRabbit's concern that "429 entries carry no model label" is not applicable — our Lua capture_model filter extracts the model from the request body before rate limiting occurs, so the model stream label is present on all log entries including rate-limited ones.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -0,0 +1,433 @@
...
+                    name: loki
+                  query: >-
+                    (sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription"}
+                    | response_code=~"2.." | user_id=~"$user" | user_id!="-" [$__range]))
+                    /
+                    sum(count_over_time({service_name="maas-gateway",
+                    subscription=~"$subscription"}
+                    | user_id=~"$user" | user_id!="-" [$__range])))
+                    or vector(1)
```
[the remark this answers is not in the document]
tgitelman: Keeping or vector(1). The fallback only fires when there is literally zero traffic in the time window. With no requests, no failures occurred, so 100% success is mathematically correct (vacuous truth). Changing to or vector(0) would show 0% success during idle periods (e.g. off-hours), which is more misleading — it suggests a total outage when the system is simply idle. An actual outage produces failing requests, which means the division itself returns a real sub-1.0 value and the or fallback never fires.

deployment/components/observability/observability/dashboards/loki-datasource.yaml
tgitelman: Fixed. Changed from "Loki (Admin)" to "Usage".

deployment/components/observability/observability/dashboards/loki-datasource.yaml
tgitelman: Fixed. Changed from openshift-logging to opendatahub in the datasource URL. The service name remains a __LOKI_GATEWAY_SVC__ sed placeholder.

deployment/components/observability/observability/dashboards/usage-admin-loki-dashboard.yaml
tgitelman: Fixed. Changed from "MaaS API usage" to "model usage" in both admin and user dashboards.

deployment/components/observability/usage-logs/loki-datasource.yaml
```diff
@@ -0,0 +1,38 @@
...
+      kind: "LokiDatasource"
+      spec:
+        proxy:
+          kind: HTTPProxy
+          spec:
+            # LokiStack gateway URL — adjust name/namespace to match your LokiStack CR.
+            url: https://__LOKI_GATEWAY_SVC__.opendatahub.svc.cluster.local:8080/api/logs/v1/application
+            allowedEndpoints:
+              - endpointPattern: "/loki/api/v1/.*"
+                method: GET
```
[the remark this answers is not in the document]
tgitelman: All Loki v1 GET endpoints are read-only query/label/series operations — there's nothing destructive or sensitive behind the wildcard. Explicitly listing each endpoint adds maintenance burden without meaningful security gain (if a new Perses panel type or COO version calls a different Loki endpoint, it would silently break). The method: GET constraint already prevents any write/delete operations. Keeping the wildcard for maintainability.

ahadas (changes requested):
please squash and add `Co-authored-by: Arik Hadas <ahadas@redhat.com>`

ahadas:
replaced by https://github.com/opendatahub-io/models-as-a-service/pull/1203

PR #999, opened 2026-06-15
feat: add Loki query proxy for user-scoped log isolation

tgitelman:
## Description
Adding a reverse-proxy that will be used by a usage dashboard for non-admin users based on data in Loki.

## How Has This Been Tested?
Tests manually

## Merge criteria:

### DIFF
--- a/deployment/components/observability/usage-logs/tenancy-proxy-code.yaml
+++ b/deployment/components/observability/usage-logs/tenancy-proxy-code.yaml
@@ -1,0 +1,538 @@
+# Loki Query Proxy -- Python stdlib-only implementation
+# Runs on ubi9/python-312 base image with zero external dependencies.
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: loki-query-proxy-source
+  namespace: opendatahub
+  labels:
+    app: loki-query-proxy
+    app.kubernetes.io/managed-by: maas-controller
+    app.kubernetes.io/component: loki-proxy
+data:
+  main.py: |
+    #!/usr/bin/env python3
+    """Loki Query Proxy - stdlib-only implementation."""
+    import contextvars
+    import hashlib
+    import json
+    import logging
+    import os
+    import ssl
+    import sys
+    import time
+    import uuid
+    from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
+    from threading import Lock
+    from urllib.parse import urlparse, parse_qs, urlencode, urljoin
+
+    from auth import extract_from_bearer_token, create_ssl_context
+    from rewriter import inject_user_filter
+
+    # Context variable for request ID
+    request_id_var = contextvars.ContextVar('request_id', default='')
+
+    # Simple in-memory cache for query results
+    query_cache = {}
+    cache_lock = Lock()
+    CACHE_TTL_SECONDS = 30  # Cache responses for 30 seconds
+
+    class RequestIDFilter(logging.Filter):
+        """Inject request ID into log records."""
+        def filter(self, record):
+            record.request_id = request_id_var.get() or '-'
+            return True
+
+    # Configure logging with request ID support
+    log_level = os.getenv('LOG_LEVEL', 'INFO').upper()
+    logging.basicConfig(
+        level=getattr(logging, log_level, logging.INFO),
+        format="%(asctime)s [%(levelname)s] [%(request_id)s] %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+    )
+    # Add filter to all handlers so request_id is available in all log records
+    for handler in logging.getLogger().handlers:
+        handler.addFilter(RequestIDFilter())
+    logger = logging.getLogger(__name__)
+
+    # Global configuration
+    ssl_context = None
+    token_review_url = None
+
+    ALLOWED_PATHS = {
+        "/loki/api/v1/query",
+        "/loki/api/v1/query_range",
+        "/loki/api/v1/index/stats",
+        "/loki/api/v1/index/volume",
+        "/loki/api/v1/index/volume_range",
+        "/loki/api/v1/labels",
+        "/loki/api/v1/series",
+        "/health",
+        "/ready",
+    }
+
+    HOP_BY_HOP_HEADERS = {
+        "connection",
+        "keep-alive",
+        "proxy-authenticate",
+        "proxy-authorization",
+        "te",
+        "trailers",
+        "transfer-encoding",
+        "upgrade",
+        "authorization",
+        "content-length",
+    }
+
+
+    def is_allowed_path(path):
+        """Check if path is allowed."""
+        if path in ALLOWED_PATHS:
+            return True
+        if path.startswith("/loki/api/v1/label/") and path.endswith("/values"):
+            return True
+        return False
+
+
+    class ProxyHandler(BaseHTTPRequestHandler):
+        """HTTP request handler for the Loki proxy."""
+
+        protocol_version = "HTTP/1.1"
+
+        def log_message(self, format, *args):
+            """Suppress default logging - we handle it ourselves."""
+            pass
+
+        def send_json_error(self, msg, code):
+            """Send JSON error response."""
+            self.send_response(code)
+            self.send_header("Content-Type", "application/json")
+            self.end_headers()
+            response = {
+                "status": "error",
+                "error": msg,
+                "errorType": "proxy",
+                "data": None,
+            }
+            self.wfile.write(json.dumps(response).encode())
+
+        def do_GET(self):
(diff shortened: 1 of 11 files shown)


deployment/components/observability/loki-proxy/deployment-user.yaml
```diff
@@ -0,0 +1,79 @@
...
+  template:
+    metadata:
+      labels:
+        app: loki-query-proxy
+        mode: user
+    spec:
+      serviceAccountName: loki-query-proxy
+      containers:
+      - name: proxy
+        image: registry.access.redhat.com/ubi9/go-toolset:1.25
```
ahadas: this means that the proxy will run in production on the go-toolset image - which is not that great (that's an image that is meant for development). I think it is worth trying to convert it to a language, e.g., python, with a minimal image that can run the proxy

deployment/components/observability/usage-logs/tenancy-proxy-code.yaml
```diff
@@ -0,0 +1,465 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: loki-query-proxy-source
+  labels:
+    app: loki-query-proxy
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: loki-proxy
+data:
```
ahadas: @jland-redhat what's your view on this? we did this trick in several places (not in rhoai) to avoid adding images for "temporary" stuff - I think golang is not a great choice here, but in general - would you agree going with this approach?

tgitelman:
@coderabbitai review

deployment/components/observability/usage-logs/tenancy-proxy.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: loki-query-proxy-user
+  labels:
+    app: loki-query-proxy
+    mode: user
```
ahadas: this assumes we may go through this proxy also as admin - but I think it would make more sense to have two different datasources, and the one for admin's will go directly to Loki, wdyt?

deployment/components/observability/loki-proxy/deployment-user.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: loki-query-proxy-user
```
ahadas: I'd go with something with tenancy in its name, like:

```suggestion
  name: usage-logs-tenancy-proxy
```

deployment/components/observability/loki-proxy/rbac.yaml
```diff
@@ -0,0 +1,58 @@
...
+    app: loki-query-proxy
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: loki-proxy
+rules:
+- apiGroups: ["authentication.k8s.io"]
+  resources: ["tokenreviews"]
+  verbs: ["create"]
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get", "list"]
```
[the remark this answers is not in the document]
ahadas: I think this comment makes sense

deployment/components/observability/loki-proxy/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+    mode: user
+    app.kubernetes.io/managed-by: maas-observability
+    app.kubernetes.io/component: loki-proxy
+spec:
+  type: ClusterIP
+  ports:
+  - port: 8080
+    targetPort: 8080
+    protocol: TCP
+    name: http
```
ahadas: why not https?

deployment/components/observability/loki-proxy/deployment-user.yaml
```diff
@@ -0,0 +1,77 @@
...
+        app: loki-query-proxy
+        mode: user
+    spec:
+      serviceAccountName: loki-query-proxy
+      containers:
+      - name: proxy
+        image: registry.access.redhat.com/ubi9:9.5-1734497371
+        command:
+        - python3
+        - /opt/app-root/src/proxy/main.py
```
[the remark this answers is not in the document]
ahadas: done

ahadas (approved):

mariusdanciu:
Why do we need a separate reverse-proxy instance as opposed of doing the same logic inside an IPP filter ?

ahadas:
> Why do we need a separate reverse-proxy instance as opposed of doing the same logic inside an IPP filter ?

the queries from the dashboard don't go through the gateway, they go directly from the dashboard to Loki

mariusdanciu:
> the queries from the dashboard don't go through the gateway, they go directly from the dashboard to Loki

Yes this is obvious, but why not use the same gateway ?

ahadas:
> > the queries from the dashboard don't go through the gateway, they go directly from the dashboard to Loki
> 
> Yes this is obvious, but why not use the same gateway ?

the question is not clear to me - use the same gateway in which case then?

mariusdanciu:
> the question is not clear to me - use the same gateway in which case then?

When the user wants to view the usage info in a Perses dashboard, why not use the tenant gateway and do the loki proxy inside IPP ? Additionally do you consider the tenant info when query the Loki logs ?

For MaaS multitenancy we use different gateway per tenant and also different IPP ext_proc instance per tenant in order to ensure traffic isolation. However the Loki ext_proc instance seems to be shared for all tenants but do correct me if I'm wrong.

ahadas:
> When the user wants to view the usage info in a Perses dashboard, why not use the tenant gateway and do the loki proxy inside IPP ?

can you explain the potential value you see in doing that?
I guess we'd both agree that for admins, when the queries can go to Loki directly, there's not much sense in going through a gateway.
It feels like you see some value in running the proxy within IPP as opposed to running it as a separate pod - we haven't considered this, and I guess I'm missing what we'll get out of that

> Additionally do you consider the tenant info when query the Loki logs ?

not at the moment, we can add stuff if needed

> 
> For MaaS multitenancy we use different gateway per tenant and also different IPP ext_proc instance per tenant in order to ensure traffic isolation. However the Loki ext_proc instance seems to be shared for all tenants but do correct me if I'm wrong.

well, currently it would apply only to the default mass gateway (it won't be shared) but as I commented [here](https://github.com/opendatahub-io/models-as-a-service/pull/1032#issuecomment-4992498912), there's on going work to emit logs from other AI tenants as well

mariusdanciu:
> can you explain the potential value you see in doing that?
I guess we'd both agree that for admins, when the queries can go to Loki directly, there's not much sense in going through a gateway.

Are Perses dashboards for admins only or end users as well ? Even if we don't need to go via the tenant Gateway, we ca use the default Gateway and Envoy/Praxis can handle this logic of injecting additional info in the Loki queries. So there is still a gateway involved. 

> It feels like you see some value in running the proxy within IPP as opposed to running it as a separate pod - we haven't considered this, and I guess I'm missing what we'll get out of that

I'm not asking to change things now of course just this this type of logic seems to fit in a proxy filter without the need to mange additional PODs, images, RBACS, etc. This logic would work well with Praxis as well. Also we'd eliminate another network hop since the proxy filter can talk directly with Loki. Envoy/Praxis are already reverse proxies. So the flow would be Client -> Gateway Proxy filter -> Loki instead of Client -> Gateway -> Custom Loki Reverse Proxy -> Loki. Filters are a natural fit to enrich/massage incoming requests that are forward to the backend services (loki in this case)

ahadas:
replaced by https://github.com/opendatahub-io/models-as-a-service/pull/1203

ahadas:
> Are Perses dashboards for admins only or end users as well ?

both - we'll have a dashboard for admins where they can see usage of all users, and user-scoped dashboards for non-admin users (admin users can also open their user-scoped dashboard)
 
> > It feels like you see some value in running the proxy within IPP as opposed to running it as a separate pod - we haven't considered this, and I guess I'm missing what we'll get out of that
> 
> I'm not asking to change things now of course just this this type of logic seems to fit in a proxy filter without the need to mange additional PODs, images, RBACS, etc. This logic would work well with Praxis as well. Also we'd eliminate another network hop since the proxy filter can talk directly with Loki. Envoy/Praxis are already reverse proxies. So the flow would be Client -> Gateway Proxy filter -> Loki instead of Client -> Gateway -> Custom Loki Reverse Proxy -> Loki. Filters are a natural fit to enrich/massage incoming requests that are forward to the backend services (loki in this case)

ok, now I think I understand where you coming from - my understanding is that the current process is `Client -> Perses-proxy -> Custom Loki reverse proxy -> Loki`. the Perses proxy loads the datasources and routes requests from the console to their target, but it's not based on a k8s gateway

PR #1002, opened 2026-06-15
fix: skip storage-initializer for simulator LLMInferenceServices

jland-redhat:
Simulator LLMInferenceService manifests used in e2e and sample deployments were still triggering KServe’s storage-initializer init container to download hf://sshleifer/tiny-gpt2, even though llm-d-inference-sim never reads model weights. That download can time out in CI and block pods from becoming Ready.

This change sets storageInitializer.enabled: false on all simulator LLMIS resources and replaces the HuggingFace URI with a placeholder (hf://placeholder/no-model) that satisfies the schema but is never fetched.


ishitasequeira (approved):
lgtm

PR #1014, opened 2026-06-17
fix: pick default appNamespace to be same as maas-controller instead of hardcoded value

ishitasequeira:
## Description
pick default appNamespace to be same as maas-controller instead of hardcoded value

## Merge criteria:


jrhyness (approved):
lgtm.  Thanks.

PR #993, opened 2026-06-12
feat:  integrate multi-tenant components

jrhyness:
Integrates multi-tenant infrastructure from #968 (per-tenant maas-api) and  
  #912                                                                        
  (gateway-level AuthPolicy), ensuring all components work together correctly.
                                                                              
                                                                                                                             
  ## Summary                                                                  
                                                                              
  This PR makes multi-tenant components work together by:                     
                                                                              
  • Introducing dual tenant naming:  TenantIdentifierFor  (resources) vs      
  TenantNameFor  (database/headers)                                           
  • Setting default tenant  TENANT_NAME  to  "models-as-a-service"  for       
  database consistency                                                        
  • Fixing AuthPolicy CEL expressions that broke on management endpoints      
  • Ensuring backward compatibility for default tenant resources              
  • Addressing CodeRabbit security and code quality issues                    
                                                                              
  ## Key Changes                                                              
                                                                              
  ### 1. Dual Tenant Naming                                                   
                                                                              
  Split tenant identifier into two concepts:                                  
                                                                              
  •  TenantIdentifierFor(tenant) : Kubernetes resource naming                 
    • Default tenant:  ""  → resource name  maas-api  (backward compatible)   
    • AITenant-managed:  "redteam"  → resource name  maas-api-redteam         
  •  TenantNameFor(tenant) : Database queries, AuthPolicy headers,            
  TENANT_NAME  env var                                                        
    • Default tenant:  "models-as-a-service"  (matches DB default)            
    • AITenant-managed:  "redteam"                                            
                                                                              
                                                                              
  This split preserves backward-compatible resource names while maintaining   
  database consistency.                                                       
                                                                              
  ### 2. Default Tenant TENANT_NAME                                           
                                                                              
  Changed: Default tenant  TENANT_NAME  env var from  ""  to  "models-as-a-   
  service"                                                                    
                                                                              
  Why: Database  api_keys.tenant  column has  DEFAULT 'models-as-a-service'   
  (migration 0004). The env var must match for proper tenant filtering in     
  database queries and AuthPolicy header validation.                          
                                                                              
  ### 3. AuthPolicy Subscription-Info Fix                                     
                                                                              
  Fixed "no such key: subscription-info" errors on  /v1/models  requests. The 
  subscription-info  metadata is only fetched for  /llm/*  paths, but identity
  filter referenced it unconditionally.                                       
                                                                              
  Before (broken):                                                            
                                                                              
[...]

PR #1018, opened 2026-06-17
fix: skip tenant-cleanup finalizer on default-tenant

jland-redhat:
## Summary

- Stop adding `maas.opendatahub.io/tenant-cleanup` on `default-tenant`; its operands are torn down via `Config/default` owner references (#894).
- Strip a legacy `tenant-cleanup` finalizer on reconcile so upgraded clusters converge without manual patches.
- Keep `tenant-cleanup` on AITenant-managed tenants (`TenantIdentifierFor` ≠ `""`); per-tenant `maas-api-*` and `MaaSAuthPolicy` cleanup on individual delete is unchanged.

Fixes the platform-disable hang where `default-tenant` stayed `Terminating` with a no-op finalizer after `maas-controller` was removed (#993 regression on top of #894).

Follow-up (not in this PR): cluster-wide tenant finalization during `Config/default` teardown for AITenant-managed tenants — see `docs/jira/tenant-cleanup-on-platform-teardown.jira`.

## Test plan

- [x] `make -C maas-controller test`
- [x] `TestTenantReconcile_DefaultTenantDoesNotAddCleanupFinalizer`
- [x] `TestTenantReconcile_DefaultTenantStripsLegacyCleanupFinalizer`
- [x] `TestTenantReconcile_AITenantManagedAddsCleanupFinalizer`
- [x] Manual: on a cluster with legacy finalizer on `default-tenant`, reconcile removes it and Config delete proceeds
- [x] Manual: AITenant-managed tenant delete still cleans up `maas-api-*` and strips finalizer

## Risk analysis

**Risk rating: 2**

**Why:** Controller-only change with focused unit tests. Blast radius is limited to `default-tenant` finalizer behavior; AITenant delete path is untouched. Improves platform disable for the common case but does not address AITenant tenants stuck during full teardown (tracked in follow-up JIRA). Full operator disable flow is not covered by default Prow smoke.

Follow up Jira:
https://redhat.atlassian.net/browse/RHOAIENG-69775

@coderabbitai review

## Summary

* **Bug Fixes**
  * Improved tenant reconciliation so the tenant-cleanup finalizer is added or removed based on tenant type, ensuring the correct cleanup behavior during both normal reconciliation and deletion handling.
  * Default tenants now end up without the legacy cleanup finalizer, while AI-managed tenants retain it.

* **Tests**
  * Updated and expanded tenant reconciliation test coverage to validate finalizer presence/absence for default vs AI-managed tenants.

### DIFF
--- a/maas-controller/pkg/controller/maas/tenant_reconcile.go
+++ b/maas-controller/pkg/controller/maas/tenant_reconcile.go
@@ -52,6 +52,17 @@ const (
 	tenantFinalizer = "maas.opendatahub.io/tenant-cleanup"
 )
 
+// tenantUsesCleanupFinalizer reports whether this Tenant should carry tenant-cleanup.
+// The default platform tenant (no AITenant labels) relies on Config/default GC for teardown;
+// only AITenant-managed tenants need explicit per-tenant resource cleanup on delete.
+func tenantUsesCleanupFinalizer(tenant *maasv1alpha1.Tenant) (bool, error) {
+	tenantID, err := tenantreconcile.TenantIdentifierFor(tenant)
+	if err != nil {
+		return false, err
+	}
+	return tenantID != "", nil
+}
+
 func managementState(ann map[string]string) string {
 	if ann == nil {
 		return ""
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
lgtm

dmytro-zaharnytskyi (commented):
Correct me if I'm wrong but I think we have a gap for the ADR default-tenant path here..

So if we use unlabeled/bootstrap default tenant path we skip tenant-cleanup, which is right. But if the default tenant is managed via AITenant/models-as-a-service, the AITenant controller labels the tenant with managed-by-aitenant=true and tenant-name=models-as-a-service, then TenantIdentifierFor returns a non-empty value, so this helper still treats it as needing cleanup and can re-add the finalizer on reconcile

So maybe we should add a test for "default tenant in the configured platform namespace with AITenant labels"? (maybe later as follow up)

PR #1025, opened 2026-06-18
feat: add new module CRD ModelsAsService

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-66772

## Description

This PR:
1. Creates a minimal module CRD `ModelsAsService` within MaaS repo, with mandatory fields in `status` and only `managementState` in `spec`.
2. Implements the `status.conditions` by aggregating the corresponding ones in `Tenants`.

## How Has This Been Tested?

Deployed on a cluster, created a `ModelsAsService` and multiple `Tenants`, and verified the following scenarios:
- `Tenants` all active -> `ModelsAsService` `Ready=True` `Degraded=False`.
- Any `Tenant` degraded -> `ModelsAsService` `Ready=True` `Degraded=True`.
- Any `Tenant` pending -> `ModelsAsService` `Ready=False` `Degraded=False`.

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added a new cluster-scoped “Models as Service” custom resource with readiness/phase status, release listing, and kubectl-friendly columns (Ready/Reason/Age).
  * Introduced a controller that aggregates Tenant state into the singleton’s status conditions (ready, provisioning succeeded, degraded) and phase.
* **Bug Fixes**
  * Reduced churn by avoiding status writes when the computed result hasn’t changed.
* **Tests**
  * Added unit tests covering API registration, CRD contract validation, and condition message formatting/capping.
* **Documentation**
  * Updated startup/reconciliation comments to reflect the new aggregation flow.


maas-controller/api/components/v1alpha1/modelsasservice_types.go
```diff
@@ -0,0 +1,128 @@
...
+	ModelsAsServicePhaseReady    ModelsAsServicePhase = "Ready"
+	ModelsAsServicePhaseNotReady ModelsAsServicePhase = "Not Ready"
+)
+
+// ComponentRelease reports a single installed component or operator release.
+type ComponentRelease struct {
+	// Name identifies the component or operator.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
ryancham715: Thanks for the catch. I fixed the `phase` part of this by adding a kubebuilder enum and regenerating the CRD, so `status.phase` is now constrained to `Ready` / `Not Ready`.

For the `ComponentRelease` string fields, I’m not applying the suggested `MaxLength` change in this PR. Those fields live under `status.releases` (controller-written status, not user-provided spec input), and the shared ODH `common.ComponentRelease` type currently follows the same contract without length bounds. I’d prefer to handle that separately, if we want to tighten the status schema consistently across both MaaS and ODH.

deployment/base/maas-controller/crd/bases/components.platform.opendatahub.io_modelsasservices.yaml
```diff
@@ -0,0 +1,176 @@
...
+                default: Managed
+                description: |-
+                  ManagementState controls whether the module operator should reconcile or remove
+                  the MaaS platform resources represented by this contract.
+                enum:
+                - Managed
+                - Removed
+                type: string
+            type: object
+          status:
```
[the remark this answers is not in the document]
ryancham715: Review.

deployment/base/maas-controller/crd/bases/components.platform.opendatahub.io_modelsasservices.yaml
```diff
@@ -0,0 +1,176 @@
...
+                  by maas-controller.
+                format: int64
+                minimum: 0
+                type: integer
+              phase:
+                description: |-
+                  Phase is a high-level lifecycle summary expected by the module handler.
+                  Expected values are "Ready" and "Not Ready".
+                type: string
+              releases:
```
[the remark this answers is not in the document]
ryancham715: Review.

maas-controller/api/components/v1alpha1/modelsasservice_types_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	statusSchema, ok := schema.Properties["status"]
+	if !ok {
+		t.Fatal("CRD schema is missing status")
+	}
+	for _, field := range []string{"phase", "observedGeneration", "conditions", "releases"} {
+		if _, ok := statusSchema.Properties[field]; !ok {
+			t.Fatalf("CRD status is missing %q", field)
+		}
+	}
```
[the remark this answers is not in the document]
ryancham715: Review.

somya-bhatnagar (dismissed):
LGTM

deployment/base/maas-controller/crd/bases/components.platform.opendatahub.io_modelsasservices.yaml
```diff
@@ -0,0 +1,182 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.4
+  name: modelsasservices.components.platform.opendatahub.io
+spec:
+  group: components.platform.opendatahub.io
+  names:
+    kind: ModelsAsService
```
Jaland, line 11: Since we are making this change anyway can we please rename this to ModelsAsAService (The lack of an A has been driving me crazy)

deployment/base/maas-controller/crd/bases/components.platform.opendatahub.io_modelsasservices.yaml
```diff
@@ -0,0 +1,182 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.4
+  name: modelsasservices.components.platform.opendatahub.io
+spec:
+  group: components.platform.opendatahub.io
```
Jaland, line 9: Is this what other teams are doing? And does it make sense?

Just kind of a strange thing to have the group `components.platform.opendatahub.io` when it lives in the MaaS repo

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
+/*
```
Jaland, line 1: We have a "ManagmentState" but it seems to never be referenced in this file.

Firstly does ManagmentState make sense or should we base the managment state on nthe existinance of the ModelsAsAAService object (defer to whatever the doc says on this)?

Secondly if it is a field we are including we should respect it.

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		Message:            message,
+		ObservedGeneration: generation,
+		LastTransitionTime: metav1.Now(),
+	})
+}
+
+func defaultModelsAsServiceReleases() []componentsv1alpha1.ComponentRelease {
+	return []componentsv1alpha1.ComponentRelease{
+		{
+			Name:    "Models as a Service",
```
Jaland, line 213: Right now this is just static info can we also include a version number here?

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+func (s moduleTenantSummary) provisioningSucceeded() bool {
+	return s.total > 0 && len(s.unprovisioned) == 0
+}
+
+func summarizeModuleTenants(tenants []maasv1alpha1.Tenant) moduleTenantSummary {
+	summary := moduleTenantSummary{}
+
+	for i := range tenants {
+		tenant := tenants[i]
+		if tenant.Name != maasv1alpha1.TenantInstanceName {
```
Jaland, line 154: I am not totally following why we are treating the default tenant any different here.

Jaland (changes requested):
Lets a few change request but one other big thing is that when we check this in it is going to break the current flow for the operator install.

Do we have the other required commits lined up so we are not just leaving the code in a bad state while we wait for everything else?

And have we done any validation of the install of this through the new AI Opeartor?

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	if err := r.Status().Patch(ctx, &module, client.MergeFrom(original)); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ModelsAsServiceReconciler) reconcileStatus(ctx context.Context, module *componentsv1alpha1.ModelsAsService) error {
```
ishitasequeira, line 86: The spec defines Managed and Removed states, but Reconcile never inspects `module.Spec.ManagementState`. If the platform sets it to Removed, the controller still aggregates and patches status as if it's Managed. Should this short-circuit with a condition like `Ready=False, Reason=Removed`?

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		return err
+	}
+
+	summary := summarizeModuleTenants(tenants.Items)
+
+	module.Status.ObservedGeneration = module.Generation
+	module.Status.Releases = defaultModelsAsServiceReleases()
+	module.Status.Phase = componentsv1alpha1.ModelsAsServicePhaseNotReady
+	if summary.allReady() {
+		module.Status.Phase = componentsv1alpha1.ModelsAsServicePhaseReady
```
ishitasequeira, line 98: When total == 0, the controller sets Not Ready / NoTenantsFound and returns ctrl.Result{} — no requeue. It relies on the Tenant watch to trigger reconciliation when tenants appear. That works, but there's a window: if the ModelsAsService CR is created before the Tenant CRD is installed, the List succeeds with zero results and the controller goes silent until a Tenant event fires. A periodic requeue (e.g., 5 minutes) on the zero-tenant path would add resilience.

Also, should we consider ModelAsService to be `Ready` even if there are no tenants? We will still have Controller working in the right way.

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+
+func summarizeModuleTenants(tenants []maasv1alpha1.Tenant) moduleTenantSummary {
+	summary := moduleTenantSummary{}
+
+	for i := range tenants {
+		tenant := tenants[i]
+		if tenant.Name != maasv1alpha1.TenantInstanceName {
+			continue
+		}
+
```
ishitasequeira, line 157: This lists every Tenant across all namespaces, then discards any not named default-tenant. If we are summarizing the status across all tenants, why ignore others?

maas-controller/pkg/controller/maas/modelsasservice_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	original := module.DeepCopy()
+	if err := r.reconcileStatus(ctx, &module); err != nil {
+		return ctrl.Result{}, err
+	}
+	if equality.Semantic.DeepEqual(original.Status, module.Status) {
+		return ctrl.Result{}, nil
+	}
```
ishitasequeira, line 77: You compare `original.Status` vs `module.Status` before patching. But `setModuleCondition` always passes `metav1.Now()`, which changes the `LastTransitionTime` on every reconcile for new conditions. On the first reconcile this triggers a write (correct), but on subsequent reconciles `apimeta.SetStatusCondition` preserves the old timestamp when status hasn't changed, so DeepEqual will return true (correct). This works, but it's subtle enough that a test explicitly asserting "no status write on re-reconcile with unchanged tenants" would catch regressions if someone refactors the condition-setting logic.

somya-bhatnagar:
Closing this PR as it is no longer needed for the current modularization architecture.

With AGO PR #29 merged, MaaS is deployed and managed by the AI Gateway Operator (AGO), not the ODH operator directly. The ODH operator reads the AIGateway CR status set by AGO (deployment readiness) to surface MaaS health in DSC — the `ModelsAsService` CR as an intermediate status aggregator is not required in this design.

Fine-grained Tenant health visibility can be revisited as a follow-up if needed.

PR #1026, opened 2026-06-19
feat: add validating webhook to prevent multiple AITenants from using…

jrhyness:
… the same Gateway

## Description
Implements https://redhat.atlassian.net/browse/RHOAIENG-66132
- Validates gateway uniqueness at AITenant CREATE and UPDATE time
- Rejects conflicts with detailed error message naming the conflicting tenant
- Handles both explicit gateway names and default (AITenant name) resolution
- Gateway reference validation mirrors controller reconcile logic

Changes:
- Add Client and GatewayNamespace fields to AITenantValidator struct
- Implement validateGatewayUniqueness method to check for conflicts across all AITenants
- Enhance ValidateCreate to validate gateway uniqueness before allowing creation
- Enhance ValidateUpdate to validate new gateway reference when gateway changes
- Add gatewayRefFor helper method matching controller defaulting logic
- Update webhook configuration to include UPDATE operation

## How Has This Been Tested?
- Add unit tests covering create/update scenarios with gateway conflicts
- Verified on cluster - webhook correctly rejects duplicate gateway usage.

## Merge criteria:

## Release Notes

* **New Features**
  * Extended the AITenant admission webhook to handle `UPDATE` operations.

* **Bug Fixes**
  * Improved AITenant webhook validation to require full webhook configuration and enforce creation requests occur in the expected namespace.
  * Strengthened gateway isolation by rejecting `AITenant` create/update when another tenant already claims the same effective gateway reference (including cases with preexisting duplicates).

* **Tests**
  * Added unit tests for gateway name defaulting and collision handling on both create and update, including self-update allowances.


maas-controller/pkg/webhook/aitenant_webhook.go
```diff
@@ -80,3 +123,50 @@ func (v *AITenantValidator) ValidateDelete(ctx context.Context, obj runtime.Obje
...
+			existingGatewayRef.Name == targetGatewayRef.Name {
+			return fmt.Errorf(
+				"gateway %s/%s is already in use by AITenant %s/%s; "+
+					"each AITenant requires a dedicated Gateway for isolation; "+
+					"please create a new Gateway for this tenant or specify a different gateway name in spec.gateway.name",
+				targetGatewayRef.Namespace, targetGatewayRef.Name,
+				existingTenant.Namespace, existingTenant.Name,
+			)
+		}
+	}
```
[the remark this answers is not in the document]
jrhyness: Out-of-scope for this PR.

maas-controller/pkg/webhook/aitenant_webhook.go
```diff
@@ -51,25 +54,68 @@ func (v *AITenantValidator) ValidateCreate(ctx context.Context, obj runtime.Obje
...
+	if v.GatewayNamespace == "" {
+		return nil, errors.New("gateway namespace is not configured")
+	}
+
+	// Check if gateway reference changed
+	oldGatewayRef := v.gatewayRefFor(oldAITenant)
+	newGatewayRef := v.gatewayRefFor(newAITenant)
+
+	if oldGatewayRef.Namespace == newGatewayRef.Namespace && oldGatewayRef.Name == newGatewayRef.Name {
```
dmytro-zaharnytskyi: If two AITenants already share a gateway (pre-webhook data, etc) any unrelated update that doesn’t change spec.gateway returns early here and never hits validateGatewayUniqueness 🤔 so probably we need to correct this line

maas-controller/pkg/webhook/aitenant_webhook.go
```diff
@@ -80,3 +126,50 @@ func (v *AITenantValidator) ValidateDelete(ctx context.Context, obj runtime.Obje
...
+	return ref
+}
+
+// validateGatewayUniqueness checks that no other AITenant is using the same Gateway.
+// The oldAITenant parameter is nil for creates, and set for updates to exclude self-comparison.
+func (v *AITenantValidator) validateGatewayUniqueness(ctx context.Context, aitenant *maasv1alpha1.AITenant, oldAITenant *maasv1alpha1.AITenant) error {
+	targetGatewayRef := v.gatewayRefFor(aitenant)
+
+	// List all AITenant CRs in the AITenant namespace (where AITenant CRs live, e.g., ai-tenants).
+	var aitenantList maasv1alpha1.AITenantList
```
dmytro-zaharnytskyi, line 143: This is list-then-compare, not atomic. Two concurrent creates for different AITenant names but the same gateway can both list before either object exists, and both get admitted.

But seems like out of scope a bit anyway, so even if you agree on that - probably the best way is to set it as a follow up

dmytro-zaharnytskyi (commented):
🤗

maas-controller/pkg/webhook/aitenant_webhook.go
jrhyness: CodeRabbit flagged the same thing.  It seems unlikely, but possible.   Ok, I'll create a follow-up issue. 
Thanks.

maas-controller/pkg/webhook/aitenant_webhook.go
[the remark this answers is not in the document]
jrhyness: https://redhat.atlassian.net/browse/RHOAIENG-70581

dmytro-zaharnytskyi (approved):
perfecto lgtm

chaitanya1731 (approved):
LGTM

PR #1034, opened 2026-06-23
fix: validate user-specified API key name

ahadas:
## Description

The API key names can be specified by users and thus, we need to make sure they're valid. It's even more important now that they are returned by Authorino.

## How Has This Been Tested?

Tested manually

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * API key names are now cleaned up automatically by trimming surrounding whitespace.
  * Invalid API key names are rejected more consistently, including empty names, overly long names, and names with unsupported control characters.
  * Ephemeral API keys continue to support automatic name generation when no name is provided.

### DIFF
--- a/maas-api/internal/api_keys/handler.go
+++ b/maas-api/internal/api_keys/handler.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/gin-gonic/gin"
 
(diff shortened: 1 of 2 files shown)


maas-api/internal/api_keys/service.go
```diff
@@ -135,6 +137,17 @@ func (s *Service) CreateAPIKey(
...
+	name = strings.TrimSpace(name)
+	if len(name) == 0 {
+		return nil, ErrWhitespaceName
+	}
+	if utf8.RuneCountInString(name) > 128 {
+		return nil, ErrLongName
+	}
+	if invalidKeyNameCharsPattern.MatchString(name) {
+		return nil, ErrInvalidName
+	}
```
ishitasequeira: We seem to be adding this validations post API Key generation on line 135. Can we move the validations before generating an API Key?

Also, for ephemeral keys, we auto-generate a name if the `keyName = ""`. If a customer sets the keyName as `"   "`, it will fail the validation as it will not auto-generate a name.

ishitasequeira (changes requested):

maas-api/internal/api_keys/service.go
ahadas: > We seem to be adding this validations post API Key generation on line 135. Can we move the validations before generating an API Key?

ack, done

> Also, for ephemeral keys, we auto-generate a name if the `keyName = ""`. If a customer sets the keyName as `" "`, it will fail the validation as it will not auto-generate a name.

yep

ishitasequeira (approved):
lgtm

tgitelman:
lgtm

tgitelman (approved):
Looks Good

PR #1035, opened 2026-06-23
feat: add EnvoyFilter for OTel structured usage logging

tgitelman:
## Summary

EnvoyFilter for structured OTel usage logs on MaaS inference endpoints (`/v1/chat/completions`, `/v1/completions`), now with **controller-managed lifecycle** via `usageLogging` feature gate.

**Architecture:**

- `json_to_metadata` — extracts model (request + authoritative response) and token counts (response body)
- Lua `sse_tokens` — extracts tokens from streaming SSE responses via `bodyChunks()` (zero buffering)
- OTel ALS — emits structured attributes to the collector via gRPC
- **`Tenant.Spec.Telemetry.UsageLogging` feature gate** — controller deploys/removes EnvoyFilter at runtime, reads manifest from container filesystem, respects `telemetry.enabled` kill switch

**Identity attributes** — all sourced from Kuadrant WASM plugin's `FILTER_STATE`:

- `user_id`, `subscription`, `groups`, `key_id`, `key_name`, `organization_id`

**Emitted attributes (12 total):**
`response_code`, `response_type`, `user_id`, `subscription`, `groups`, `key_id`, `key_name`, `organization_id`, `tokens_total`, `tokens_prompt`, `tokens_completion`, `model`

## Test plan

- [x] Deployed on cluster, verified 200/429/streaming all emit correct data
- [x] `user_id`, `key_id`, `key_name`, `subscription`, `groups` confirmed populated from FILTER_STATE
- [x] Token extraction working for both JSON (non-streaming) and SSE (streaming)
- [x] Rate-limited (429) requests correctly emit with tokens=0 and request-extracted model
- [x] Feature gate `usageLogging: true` → EnvoyFilter created with correct ownerRef
- [x] Feature gate `usageLogging: false` → EnvoyFilter deleted
- [x] `telemetry.enabled: false` overrides `usageLogging: true` → EnvoyFilter deleted
- [x] Full pipeline verified: request → Envoy → OTel Collector → Loki (structured log with all 12 fields)

## Risk analysis

- **Risk rating**: 3
- **Why**: Adds CRD field (`usageLogging`) and controller logic to LifecycleReconciler. EnvoyFilter is async (if collector unreachable, requests unaffected). Feature is off by default. Covered by unit tests and on-cluster verification. Elevated from 2→3 because of CRD change and new controller reconcile path.

## Summary by CodeRabbit

* **New Features**
  * Added optional `usageLogging` configuration to the Config CRD.
  * When enabled, deploys an Istio gateway Envoy access-log filter that emits structured OpenTelemetry inference usage logs (model + token counts).
  * Supports both standard and streaming (SSE) chat/completion endpoints.
* **Bug Fixes**
  * Improved handling for missing/invalid payload data with safe default values and non-overwriting of existing metadata.
* **Tests**
  * Added unit coverage for enable/disable behavior and manifest-based endpoint patching.


ahadas (commented):
this version looks good to be
@tgitelman did you check this with streaming?

tgitelman:
> this version looks good to be @tgitelman did you check this with streaming?

Yes

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
+# MaaS Usage Logs EnvoyFilter
+#
+# Emits structured usage logs for inference requests via OTel Access Log Service.
+#
+# Identity:  FILTER_STATE(wasm.kuadrant.auth.identity.*) — set by Kuadrant Wasm
```
ahadas: need to drop the FILTER_STATE parts

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
...
+# Targeting: Gateway API targetRefs.
+#
+# Streaming (Server-Sent Events):
+#            json_to_metadata sees text/event-stream (not application/json)
+#            and fires on_error — zero buffering, tokens default to "0".
+#            A companion Lua filter (sse_tokens) then takes over: iterates
+#            bodyChunks() without buffering, scans each chunk for usage
+#            fields, and overwrites the "0" metadata with real token counts.
+#            The SSE stream passes through untouched to the client.
+#            Future: sse_to_metadata (Envoy 1.38+, not yet in OSSM) will
```
ahadas: what is this statement about the future based on? I'd either link this to the source or drop it

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
...
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: maas-default-gateway
+  configPatches:
+
+  # ── Cluster: gRPC endpoint to OTel Collector ──────────────────────────
+  # Address must match the OTel Operator generated service name:
+  #   <cr-name>-collector.<namespace> (see otel-collector-cr.yaml metadata.name)
```
ahadas: as the creation of the OTEL collector would likely change, let's drop this comment

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
...
+          - lb_endpoints:
+            - endpoint:
+                address:
+                  socket_address:
+                    address: usage-logs-collector.opendatahub.svc.cluster.local
+                    port_value: 4317
+
+  # ── json_to_metadata: Extract model + tokens from request/response ────
+  #
+  # Replaces both Lua filters with a single native Envoy filter.
```
ahadas: no need to say this replaces our tests with LUA filters

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
...
+                    string_value: "%REQ(X-MaaS-Username)%"
+                - key: subscription
+                  value:
+                    string_value: "%REQ(X-MaaS-Subscription)%"
+                - key: groups
+                  value:
+                    string_value: "%REQ(X-MaaS-Group)%"
+                - key: key_id
+                  value:
+                    string_value: "%FILTER_STATE(wasm.kuadrant.auth.identity.keyId:PLAIN)%"
```
ahadas, line 359: need to change this and the properties below that use FILTER_STATE

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,340 @@
...
+                - key: tokens_prompt
+                  value:
+                    string_value: "%DYNAMIC_METADATA(envoy.filters.http.json_to_metadata:tokens_prompt)%"
+                - key: tokens_completion
+                  value:
+                    string_value: "%DYNAMIC_METADATA(envoy.filters.http.json_to_metadata:tokens_completion)%"
+                # ── Model (json_to_metadata → dynamic metadata) ──
+                - key: model
+                  value:
+                    string_value: "%DYNAMIC_METADATA(envoy.filters.http.json_to_metadata:model)%"
```
ahadas, line 379: now we can also add key_name here

ahadas (changes requested):
the EnvoyFilter looks good overall
few comments inside and also need to update the title and description of the PR

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
ahadas: no need to, apparently it works with RHCL 1.3.4 and u/s Kuadrant 1.4.2

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,335 @@
...
+                - key: response_type
+                  value:
+                    string_value: '%CEL(response.code >= 200 && response.code < 300 ? "hit" : (response.code == 429 ? "rate_limit" : "error"))%'
+                # ── Identity ──
+                - key: user_id
+                  value:
+                    string_value: "%REQ(X-MaaS-Username)%"
+                - key: subscription
+                  value:
+                    string_value: "%REQ(X-MaaS-Subscription)%"
```
ahadas:
```suggestion
                    string_value: "%FILTER_STATE(wasm.kuadrant.auth.identity.selected_subscription:PLAIN)%"
```

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,335 @@
...
+                - key: response_code
+                  value:
+                    string_value: "%RESPONSE_CODE%"
+                - key: response_type
+                  value:
+                    string_value: '%CEL(response.code >= 200 && response.code < 300 ? "hit" : (response.code == 429 ? "rate_limit" : "error"))%'
+                # ── Identity ──
+                - key: user_id
+                  value:
+                    string_value: "%REQ(X-MaaS-Username)%"
```
ahadas:
```suggestion
                    string_value: "%FILTER_STATE(wasm.kuadrant.auth.identity.userid:PLAIN)%"
```

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,335 @@
...
+                - key: subscription
+                  value:
+                    string_value: "%REQ(X-MaaS-Subscription)%"
+                - key: groups
+                  value:
+                    string_value: "%REQ(X-MaaS-Group)%"
```
ahadas:
```suggestion
                    string_value: "%FILTER_STATE(wasm.kuadrant.auth.identity.groups:PLAIN)%"
```

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
```diff
@@ -0,0 +1,335 @@
+# MaaS Usage Logs EnvoyFilter
+#
+# Emits structured usage logs for inference requests via OTel Access Log Service.
+#
+# Identity:  user_id/subscription/groups from X-MaaS-* request headers;
+#            key_id/key_name/organization_id from auth identity filter.
```
ahadas: need to change here that everything about identity is taken from the auth identity filter

ahadas (changes requested):

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
tgitelman: added on line 308

ahadas (commented):
lgtm
now we need to see how to deploy it - I think it would make the most sense to add it to `ensureObservability` that is added in https://github.com/opendatahub-io/models-as-a-service/pull/1087. need to remember to adjust the address of the collector according to the monitoring-namespace setting

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_tenants.yaml
```diff
@@ -133,6 +133,14 @@ spec:
...
                         type: boolean
                     type: object
+                  usageLogging:
+                    default: false
+                    description: |-
+                      UsageLogging enables per-request structured OTel access logging
+                      for usage tracking (token counts, identity, model). When enabled, the
+                      controller deploys an EnvoyFilter on the tenant gateway that emits
+                      structured usage logs via OTel Access Log Service.
+                    type: boolean
```
ahadas: I see you've added it to the Tenant CRD, but as long as we generate the log at the gateway level, we can't have a per-tenant granularity, do we?

maas-controller/cmd/manager/main.go
```diff
@@ -763,6 +763,8 @@ func main() {
 		AITenantNamespace:           aitenantNamespace,
 		ObservabilityManifestsPath:  observabilityManifestsPath,
 		MonitoringNamespace:         monitoringNamespace,
+		GatewayName:                 gatewayName,
+		GatewayNamespace:            gatewayNamespace,
```
ahadas, line 792: incorrect conflict resolution?

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -106,6 +115,9 @@ func (r *LifecycleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if err := r.ensureObservabilityDashboards(ctx, log); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err := r.ensureUsageLogsEnvoyFilter(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
```
ahadas, line 118: this part should move to `ensureObservability` after rebase

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -439,26 +451,19 @@ func (r *LifecycleReconciler) ensureObservabilityDashboards(ctx context.Context,
...
 		if err := controllerutil.SetControllerReference(&cfg, &res, r.Scheme); err != nil {
 			return fmt.Errorf("set controller reference on %s %s: %w", res.GetKind(), res.GetName(), err)
 		}
 
 		if err := r.Patch(ctx, &res, client.Apply, client.ForceOwnership, client.FieldOwner("maas-controller")); err != nil {
 			if isOptionalAPIGroup(res.GroupVersionKind().Group) && (apimeta.IsNoMatchError(err) || apierrors.IsNotFound(err)) {
-				// CRD not yet registered for a known optional dependency (e.g. Perses CRDs
-				// installed by COO which may not be present yet). Skip so the rest of the
-				// platform manifests are applied and Tenant reconcile does not fail.
-				// The CRD watch will re-trigger reconcile once the CRDs appear.
```
ahadas, line 471: why is it removed?

tgitelman:
> lgtm now we need to see how to deploy it - I think it would make the most sense to add it to `ensureObservability` that is added in #1087. need to remember to adjust the address of the collector according to the monitoring-namespace setting


@ahadas Our usageLogging feature gate already adjusts the collector address per MonitoringNamespace — we just need to hook it into ensureObservability when #1087 merges. No code changes needed now.

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -48,6 +51,9 @@ import (
 // can strip it from older installs.
 const CleanupFinalizer = "maas.opendatahub.io/cleanup"
 
+// envoyFilterRelPath is the path to the EnvoyFilter manifest relative to ObservabilityManifestsPath.
+const envoyFilterRelPath = "otel-collector/envoy-otel-access-log.yaml"
```
ahadas: we need to think of adding another kustomization folder for structured logs to be able to disable metrics or logs

maas-controller/cmd/manager/main.go
tgitelman: Fixed — removed stale GatewayName field that was leftover from conflict resolution. Only GatewayNamespace remains.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Done — rebased on upstream/main and moved ensureUsageLogsEnvoyFilter into ensureObservability (called after ensureLimitadorServiceMonitor and ensureUsageDashboard).

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Accidental removal during rebase conflict resolution — restored the comment explaining optional CRD dependency handling (skip if CRD not registered, CRD watch will re-trigger).

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -444,16 +456,13 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
 		return err
 	}
 
-	// Render kustomization (reuses tenant reconciler's kustomize logic)
-	// TODO: move kustomize logic to a separate package and reuse it here.
```
ahadas, line 453: why removing this?

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -444,16 +456,13 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
 	resources, err := tenantreconcile.RenderKustomize(r.ObservabilityManifestsPath, r.MonitoringNamespace)
 	if err != nil {
 		return fmt.Errorf("render observability dashboards: %w", err)
 	}
 
-	// Apply each resource with Config as controller owner
 	for _, resource := range resources {
-		res := resource // avoid loop variable aliasing
+		res := resource
```
ahadas: what's the reason of removing the comments? they seem to still hold

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
 	return nil
 }
 
+// ensureUsageLogsEnvoyFilter deploys or removes the OTel usage logs EnvoyFilter based on
+// the Config's usageLogging feature gate. The EnvoyFilter emits structured per-request
+// usage logs (token counts, identity, model) to an OTel Collector via gRPC Access Log Service.
+func (r *LifecycleReconciler) ensureUsageLogsEnvoyFilter(ctx context.Context) error {
+	log := ctrl.LoggerFrom(ctx)
```
ahadas: I think it's better to inject the `log` instance as in other functions

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+		return r.deleteEnvoyFilterIfExists(ctx, log)
+	}
+
+	return r.applyUsageLogsEnvoyFilter(ctx, log)
+}
+
+func (r *LifecycleReconciler) deleteEnvoyFilterIfExists(ctx context.Context, log logr.Logger) error {
+	ef := &unstructured.Unstructured{}
+	ef.SetGroupVersionKind(tenantreconcile.GVKEnvoyFilter)
+	ef.SetName("maas-model-access-logs")
```
ahadas: let's make this name a constant?

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+	ef.SetNamespace(r.GatewayNamespace)
+
+	if err := r.Delete(ctx, ef); err != nil {
+		if apierrors.IsNotFound(err) || apimeta.IsNoMatchError(err) {
+			return nil
+		}
+		return fmt.Errorf("delete usage-logs EnvoyFilter: %w", err)
```
ahadas:
```suggestion
		return fmt.Errorf("failed to delete usage-logs EnvoyFilter: %w", err)
```

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+	}
+	log.Info("deleted usage-logs EnvoyFilter (usageLogging disabled)")
+	return nil
+}
+
+func (r *LifecycleReconciler) applyUsageLogsEnvoyFilter(ctx context.Context, log logr.Logger) error {
+	var cfg maasv1alpha1.Config
+	if err := r.Get(ctx, client.ObjectKey{Name: maasv1alpha1.ConfigInstanceName}, &cfg); err != nil {
```
ahadas: why querying the config here again rather than injecting it from ensureUsageLogsEnvoyFilter?

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+	if r.ObservabilityManifestsPath == "" {
+		log.Info("ObservabilityManifestsPath not configured, skipping usage-logs EnvoyFilter")
+		return nil
+	}
+
+	// ObservabilityManifestsPath points to the dashboards kustomize root
+	// (e.g. .../observability/observability/dashboards). The EnvoyFilter manifest
+	// lives under the observability component root, two levels up.
+	observabilityRoot := filepath.Dir(filepath.Dir(r.ObservabilityManifestsPath))
+	manifestFile := filepath.Join(observabilityRoot, envoyFilterRelPath)
```
ahadas: we can define another path for this, but as I commented elsewhere (forgot to submit that comment earlier, sorry about that) - I think we need to think how to structure the observability manifests there

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -476,6 +485,166 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+
+	lbEndpoints[0] = lbe0
+	ep0["lb_endpoints"] = lbEndpoints
+	endpoints[0] = ep0
+	if err := unstructured.SetNestedSlice(patch, endpoints,
+		"patch", "value", "load_assignment", "endpoints"); err != nil {
+		return fmt.Errorf("write back endpoints: %w", err)
+	}
+	configPatches[0] = patch
+	return unstructured.SetNestedSlice(ef.Object, configPatches, "spec", "configPatches")
```
ahadas, line 635: why do we need a patch here and not just to set the address field?

ahadas:
> > lgtm now we need to see how to deploy it - I think it would make the most sense to add it to `ensureObservability` that is added in #1087. need to remember to adjust the address of the collector according to the monitoring-namespace setting
> 
> @ahadas Our usageLogging feature gate already adjusts the collector address per MonitoringNamespace — we just need to hook it into ensureObservability when #1087 merges. No code changes needed now.

yeah, note that Claude generates such responses because it seems to miss the time they were written - for instance, this comment was written before it added the part that sets the address according to the monitoring namespace

tgitelman (commented):
Addressed in latest force-push (rebased on main).

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: `unstructured.SetNestedField` cannot traverse numeric slice indices — we need to manually extract each `[]any` level (endpoints[0], lb_endpoints[0]) to reach `socket_address.address`. Added a comment explaining this. If you prefer, we could alternatively do `strings.ReplaceAll` on the raw YAML bytes before decoding (replace a placeholder), but that felt more fragile.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed — restored both comments in latest push.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed — restored both comments in latest push.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed — now accepts `log logr.Logger` parameter, consistent with sibling functions.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Done — added `const envoyFilterName = "maas-model-access-logs"`, used in both delete and apply.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed — `applyUsageLogsEnvoyFilter` now receives `*Config` from the parent. Removed the redundant Get.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Added a TODO for introducing a dedicated ObservabilityRoot path. For now using `filepath.Dir(filepath.Dir(...))` with the comment explaining the fragility.

maas-controller/pkg/controller/maas/self_deployment_controller.go
ahadas: ok

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,259 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+
+	err := r.ensureUsageLogsEnvoyFilter(context.Background(), ctrl.Log)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	check := &unstructured.Unstructured{}
+	check.SetGroupVersionKind(tenantreconcile.GVKEnvoyFilter)
+	err = cl.Get(context.Background(), client.ObjectKey{
+		Name: "maas-model-access-logs", Namespace: gwNS,
+	}, check)
+	g.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "EnvoyFilter should be deleted when usageLogging=false")
```
ahadas: but when was it added?

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,259 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+		DeploymentNS:                depNS,
+		TenantSubscriptionNamespace: tenantNS,
+		GatewayNamespace:            gwNS,
+		MonitoringNamespace:         depNS,
+		ObservabilityManifestsPath:  "/nonexistent/path/that/does/not/exist",
+	}
+
+	err := r.ensureUsageLogsEnvoyFilter(context.Background(), ctrl.Log)
+	g.Expect(err).NotTo(HaveOccurred())
+}
```
ahadas: I think this is a useless test - it's not like someone will override the path...

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,259 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+	ep0, _ := endpoints[0].(map[string]any)
+	lbEndpoints, _, _ := unstructured.NestedSlice(ep0, "lb_endpoints")
+	g.Expect(lbEndpoints).NotTo(BeEmpty())
+	lbe0, _ := lbEndpoints[0].(map[string]any)
+	addr, _, _ := unstructured.NestedString(lbe0, "endpoint", "address", "socket_address", "address")
+	g.Expect(addr).To(Equal("usage-logs-collector.opendatahub.svc.cluster.local"),
+		"collector address should be patched with MonitoringNamespace")
+}
+
+func TestEnsureUsageLogsEnvoyFilter_DeletesExistingWhenDisabled(t *testing.T) {
```
ahadas: what's the difference between this and TestEnsureUsageLogsEnvoyFilter_TelemetryDisabledOverridesUsageLogging?

maas-controller/Dockerfile
```diff
@@ -11,6 +11,7 @@ ARG TARGETARCH
 
 WORKDIR /app
 COPY maas-controller/go.mod maas-controller/go.sum ./
+ENV GOTOOLCHAIN=auto
```
ahadas: why is this needed?

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Resolved — the EnvoyFilter manifest now uses a dedicated absolute path constant (`/deployment/components/observability/otel-collector/envoy-otel-access-log.yaml`), fully decoupled from the metrics kustomization in `ObservabilityManifestsPath`. Enablement is controlled by the `usageLogging` feature gate on the Config CR, so metrics and usage-logs can be toggled independently without a separate kustomize folder.

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_tenants.yaml
tgitelman: Moved to Config CR — the `usageLogging` feature gate is now cluster-wide on the Config CR (matching the cluster-wide scope of metrics/telemetry toggles). Since the EnvoyFilter is applied at the shared gateway level, per-tenant granularity isn't applicable until multi-gateway mode.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push — comments restored.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push — `log` instance is now injected as a parameter.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push — extracted to `const envoyFilterName = "maas-model-access-logs"`.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed in latest push — `cfg` is now passed from `ensureUsageLogsEnvoyFilter` instead of re-querying.

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -48,6 +51,12 @@ import (
 // can strip it from older installs.
 const CleanupFinalizer = "maas.opendatahub.io/cleanup"
 
+// envoyFilterManifestPath is the absolute path to the EnvoyFilter manifest inside the container.
+const envoyFilterManifestPath = "/deployment/components/observability/otel-collector/envoy-otel-access-log.yaml"
```
ahadas: ok, it's good that the path is now absolute but why having `otel-collector` there? I'd go with something like:

```suggestion
const envoyFilterManifestPath = "/deployment/components/observability/usage-logs/envoy-otel-access-log.yaml"
```

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,192 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+
+	err := r.ensureUsageLogsEnvoyFilter(context.Background(), ctrl.Log)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	check := &unstructured.Unstructured{}
+	check.SetGroupVersionKind(tenantreconcile.GVKEnvoyFilter)
+	err = cl.Get(context.Background(), client.ObjectKey{
+		Name: "maas-model-access-logs", Namespace: gwNS,
+	}, check)
+	g.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "EnvoyFilter should be deleted when usageLogging is disabled")
```
ahadas: I think this assumes that this test runs after TestEnsureUsageLogsEnvoyFilter_EnabledCreatesFilter that adds the envoy filter - if that's the case, I think we should avoid that so that each test can run successfully on its own

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,192 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+}
+
+func TestEnsureUsageLogsEnvoyFilter_NoConfigNoError(t *testing.T) {
```
ahadas: how can it be that this reconciler gets executed when there's no config?

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,192 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+		GatewayNamespace:            gwNS,
+		MonitoringNamespace:         depNS,
+		ObservabilityManifestsPath:  "../../../../deployment/components/observability/observability/dashboards",
+	}
+
+	err := r.ensureUsageLogsEnvoyFilter(context.Background(), ctrl.Log)
+	g.Expect(err).NotTo(HaveOccurred())
+}
+
+func boolPtr(b bool) *bool {
```
ahadas: we can use pointer.Bool() from k8s.io/utils/pointer instead

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
```diff
@@ -357,3 +358,192 @@ func TestLifecycleReconciler_LimitadorServiceMonitorCustomInterval(t *testing.T)
...
+		Scheme:                      s,
+		DeploymentName:              "maas-controller",
+		DeploymentNS:                depNS,
+		TenantSubscriptionNamespace: tenantNS,
+		GatewayNamespace:            gwNS,
+		MonitoringNamespace:         depNS,
+	}
+
+	err := r.ensureUsageLogsEnvoyFilter(context.Background(), ctrl.Log)
+	g.Expect(err).NotTo(HaveOccurred())
```
ahadas: don't you also want to ensure that there's no envoy filter after this test (after making sure that there's no envoy filter before its execution)?

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Fixed — applied your suggestion and moved the manifest to `deployment/components/observability/usage-logs/`.

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Fixed — each test is fully independent. `DeletesExistingWhenDisabled` creates its own pre-existing EnvoyFilter via `fake.NewClientBuilder().WithObjects(existingEF)` — it never depended on the previous test. Also trimmed all unused objects (dep, tenant) from test setup to make isolation clearer.

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Fixed — using `ptr.To()` from `k8s.io/utils/ptr` (the modern replacement for the deprecated `pointer` package).

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Fixed — `DisabledByDefault` now explicitly asserts no EnvoyFilter exists after execution. Also removed the redundant `DisabledNilDeletesStaleEF` test (same code path already covered by `DeletesExistingWhenDisabled` + `TestIsUsageLoggingEnabled`).

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Removed — that test was redundant (duplicate). Only one remains now.

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Removed — that test was dropped as redundant.

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Agreed — removed.

maas-controller/Dockerfile
tgitelman: Removed — it was a local build workaround (go-toolset:1.25 image has Go 1.25.7, go.mod requires 1.25.8). Not needed for upstream CI.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: Resolved — manifest moved to `deployment/components/observability/usage-logs/` and the controller now uses a hardcoded absolute path constant, fully decoupled from the metrics kustomization.

maas-controller/pkg/controller/maas/self_deployment_controller.go
tgitelman: We use server-side apply (`client.Apply` with `ForceOwnership`) which handles both create and update idempotently. The `patchClusterAddress` function modifies the in-memory unstructured object before applying — it's not a Kubernetes patch operation, just setting the address field in the parsed YAML before sending it to the API server.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

deployment/components/observability/otel-collector/envoy-otel-access-log.yaml
tgitelman: Fixed in latest push.

maas-controller/pkg/controller/maas/self_deployment_controller_test.go
tgitelman: Removed.

ahadas (commented):
looks good now, please squash all the commit to a single commit

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -481,6 +496,152 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+func (r *LifecycleReconciler) ensureUsageLogsEnvoyFilter(ctx context.Context, log logr.Logger) error {
+	var cfg maasv1alpha1.Config
+	if err := r.Get(ctx, client.ObjectKey{Name: maasv1alpha1.ConfigInstanceName}, &cfg); err != nil {
+		if apierrors.IsNotFound(err) {
+			return r.deleteEnvoyFilterIfExists(ctx, log)
+		}
+		return err
+	}
+
+	if !isUsageLoggingEnabled(cfg.Spec.UsageLogging) {
```
ahadas:
```suggestion
	if !ptr.Deref(cfg.Spec.UsageLogging, false) {
```

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -481,6 +496,152 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
...
+	if err := unstructured.SetNestedSlice(patch, endpoints,
+		"patch", "value", "load_assignment", "endpoints"); err != nil {
+		return fmt.Errorf("write back endpoints: %w", err)
+	}
+	configPatches[0] = patch
+	return unstructured.SetNestedSlice(ef.Object, configPatches, "spec", "configPatches")
+}
+
+// isUsageLoggingEnabled checks the Config spec for the usageLogging feature gate.
+func isUsageLoggingEnabled(usageLogging *bool) bool {
```
ahadas: and then we can drop this one

ahadas (approved):

ahadas (approved):

jrhyness (commented):
The usage-logs EnvoyFilter currently targets maas-default-gateway by name.  For multi-tenancy (where each AITenant gets its
  own gateway), consider moving this into the tenant reconciler as a follow-up — it already deploys per-tenant EnvoyFilters for payload-processing, so the pattern is established. That
  would give each tenant's gateway usage logging automatically and open the door to per-tenant opt-in/out later.

ahadas:
@jrhyness sure, that makes sense. We have few other PRs coming up for this feature, we'll squeeze it in the next one

jrhyness (approved):
lgtm

PR #1036, opened 2026-06-23
fix: fix ipp plugins config to match new version

nirrozenbaum:
## Description
fix ipp plugins config to match the new format.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated payload-processing to start using a centralized config file and standardized logging/tracing flags.
  * Updated pre-processing to use its centralized config file and aligned workload identity labels/naming for clarity.
* **Configuration**
  * Reworked plugin configuration to use a structured ConfigMap-based format with the default request/response processing pipeline.
  * Updated pre-processing runtime settings to match the provided gRPC port and CPU/memory request/limit values.

### DIFF
--- a/deployment/base/payload-processing/instance/deployment.yaml
+++ b/deployment/base/payload-processing/instance/deployment.yaml
@@ -28,7 +28,7 @@ spec:
             readOnlyRootFilesystem: true
           args:
             - --config-file
-            - "config/custom-ipp-config.yaml"
+            - "/config/custom-ipp-config.yaml"
             - --v=3
             - --tracing=false
           ports:
--- a/deployment/base/payload-processing/instance/deployment.yaml
+++ b/deployment/base/payload-processing/instance/deployment.yaml
@@ -27,39 +27,10 @@ spec:
                 - ALL
             readOnlyRootFilesystem: true
           args:
-            - "--streaming"
-            - --v
-            - "3"
-            - --plugin
-            - "$(MODEL_TO_HEADER)"
-            - --plugin
-            - "$(MODEL_PROVIDER_RESOLVER)"
-            - --plugin
-            - "$(API_TRANSLATION)"
-            - --plugin
-            - "$(APIKEY_INJECTION)"
+            - --config-file
+            - "config/custom-ipp-config.yaml"
+            - --v=3
             - --tracing=false
-          env:
-            - name: MODEL_TO_HEADER
-              valueFrom:
-                configMapKeyRef:
-                  name: payload-processing-plugins
-                  key: model-to-header-plugin
-            - name: MODEL_PROVIDER_RESOLVER
-              valueFrom:
-                configMapKeyRef:
-                  name: payload-processing-plugins
-                  key: model-provider-resolver-plugin
-            - name: API_TRANSLATION
-              valueFrom:
-                configMapKeyRef:
-                  name: payload-processing-plugins
-                  key: api-translation-plugin
-            - name: APIKEY_INJECTION
-              valueFrom:
-                configMapKeyRef:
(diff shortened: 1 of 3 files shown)


ishitasequeira (dismissed):
lgtm

ishitasequeira (commented):
_`pre-processing/deployment-patch.yaml` not updated_: the pod that's crashing is `payload-pre-processing`, and this patch still replaces args with `--plugin  "$(MODEL_TO_HEADER)"` + references the old ConfigMap key `model-to-header-plugin` which no longer exists after the ConfigMap restructure here. This deployment will still fail to start even after merging.                                                                                                                                                            
                                                                                                                                                                                     
_Needs its own `--config-file` switch:_ either a second entry in the ConfigMap with a pre-processing-only profile, or a named profile in the existing config.                         
   
_`maas-headers-guard` not wired into the profile_: it's declared as a plugin but never referenced in the `default` profile's request chain, so it won't execute. Should be first in `request:` (before `model-extractor`) since downstream plugins depend on it populating CycleState with the sanitized `x-maas-*` headers.

nirrozenbaum:
> _`pre-processing/deployment-patch.yaml` not updated_: the pod that's crashing is `payload-pre-processing`, and this patch still replaces args with `--plugin "$(MODEL_TO_HEADER)"` + references the old ConfigMap key `model-to-header-plugin` which no longer exists after the ConfigMap restructure here. This deployment will still fail to start even after merging.
> 
> _Needs its own `--config-file` switch:_ either a second entry in the ConfigMap with a pre-processing-only profile, or a named profile in the existing config.
> 
> _`maas-headers-guard` not wired into the profile_: it's declared as a plugin but never referenced in the `default` profile's request chain, so it won't execute. Should be first in `request:` (before `model-extractor`) since downstream plugins depend on it populating CycleState with the sanitized `x-maas-*` headers.

fixed. 
`maas-headers-guard` is a new plugin. no need to add it now, we might add it in the near future. 
separated the model to header plugin to be in the pre-processing and the other plugins in the ipp.

chaitanya1731 (approved):
lgtm

ishitasequeira (approved):
lgtm

liangwen12year (commented):
Looks good overall — clean migration to file-based config. Two questions:

1. `--streaming` flag was removed from both deployments. Is this now the default in the new IPP version, or is it no longer needed?

2. The new `custom-ipp-config.yaml` includes `api-translation` in the response pipeline, which wasn't explicitly configured before (plugins were only specified for requests via `--plugin`). Is this an intentional new behavior?

nirrozenbaum:
> Looks good overall — clean migration to file-based config. Two questions:
> 
> 1. `--streaming` flag was removed from both deployments. Is this now the default in the new IPP version, or is it no longer needed?
> 2. The new `custom-ipp-config.yaml` includes `api-translation` in the response pipeline, which wasn't explicitly configured before (plugins were only specified for requests via `--plugin`). Is this an intentional new behavior?

@liangwen12year thanks!
1. indeed. streaming flag is no longer needed (or supported), since we removed the option of buffering and now the default (and the only option) is always streaming. so no need to specify it.
2. you're right. that was one of the problems in previous version. we didn't have any way to specify if we want a plugin to run only on the request or the response. it was all or nothing. we fixed it by creating a new config api, which allows easy separation between request/response. this new config allows to set only request plugin, only response plugin or both. what you see in the config now is the actual behavior that was in previous version (although it wasn't explicitly specified as response plugin).

nirrozenbaum:
/hold
to avoid accidental merge. after #1040 is merged this one will be closed.

nirrozenbaum:
#1040 merged. closing.

PR #1049, opened 2026-06-25
fix: fall back to inference.opendatahub.io ExternalModel in MaasModel…

ishitasequeira:
## Description
- externalModelHandler.ReconcileRoute now tries `maas.opendatahub.io/ExternalModel` first, then falls back to `inference.opendatahub.io/ExternalModel` via unstructured client              
- Uses unstructured lookup to avoid adding a Go module dependency on `ai-gateway-payload-processing`
- Adds RBAC marker for inference.opendatahub.io externalmodels (codegen pending)                                                                    
                                                            
## Problem                                                                     
When a MaaSModelRef with `kind: ExternalModel` points to an `inference.opendatahub.io/ExternalModel`, the handler only looked up `maas.opendatahub.io/ExternalModel` — which doesn't exist — and failed with ExternalModel not found. This blocks the transition from legacy `maas.opendatahub.io` CRDs to the new inference.opendatahub.io CRDs.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Route reconciliation now prefers the currently supported ExternalModel API when present, with automatic fallback to the legacy API.
  * HTTPRoute resolution now derives the route name/namespace and gateway status from the supported ExternalModel status data for more consistent routing.
  * When both ExternalModel resources are unavailable, errors clearly reference both supported API groups.
* **Tests**
  * Expanded controller tests to cover inference-only, missing-both, and inference-precedence scenarios.

### DIFF
--- a/maas-controller/pkg/controller/maas/providers_external.go
+++ b/maas-controller/pkg/controller/maas/providers_external.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


maas-controller/pkg/controller/maas/providers_external.go
```diff
@@ -22,43 +22,78 @@ import (
...
+				}
+			}
 		}
+		log.Info("resolved ExternalModel from inference.opendatahub.io", "name", externalModelName, "namespace", model.Namespace)
+	} else {
 		return fmt.Errorf("failed to get ExternalModel %s: %w", model.Spec.ModelRef.Name, err)
 	}
 
-	routeName := modelnaming.ExternalModelResourceName(model.Spec.ModelRef.Name)
+	routeName := model.Spec.ModelRef.Name
```
jland-redhat: This feels like it is making some strong assumptions it would be great if we could remove some of those assumptions as part of these PRs (Have the HTTPRoute name actually reported in the status).

jland-redhat (changes requested):

liangwen12year (approved):
LGTM. Clean implementation of the inference/legacy ExternalModel fallback.

jland-redhat (approved):
lgtm

PR #1055, opened 2026-06-26
fix: select correct endpoint URL from LLMInferenceService status

jrhyness:
For: https://redhat.atlassian.net/browse/RHOAIENG-72036

Controller was selecting the first address from LLMInferenceService.Status.Addresses without checking if it contained the full model path. This caused models to have incomplete endpoint URLs (base URL only), which failed access probes in /v1/models.

LLMInferenceService status contains multiple addresses with different completeness:
- gateway-internal-model-routing: https://host/ (base URL, 3 slashes)
- gateway-internal: https://host/namespace/model (full path, 4+ slashes)

The order of these addresses can vary between environments. The old code blindly selected addresses[0], which worked when gateway-internal happened to be first but broke when gateway-internal-model-routing was first.

Modified getEndpointFromLLMISvc() to prefer addresses with full paths (slash count > 3) regardless of position, making endpoint selection robust to ordering.

Additional improvements made during debugging:

- Parallelized subscription probing in /v1/models handler to fix 504 Gateway Timeout when users have multiple subscriptions (changed from sequential N × 15s to concurrent ~15s)

- Fixed gin.Context thread-safety by capturing context before spawning goroutines

- Fixed loop variable capture bug (avoided address-of-loop-variable)

- Added _wait_for_model_ready() test helper and fixed race conditions in tests that create subscriptions for Pending models (tests were querying /v1/models before models transitioned to Ready with endpoints)

## How Has This Been Tested?

local e2e testing

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * `/v1/models` now evaluates multiple subscriptions concurrently, improving responsiveness when matching models are available via more than one subscription.
* **Bug Fixes**
  * Improved model-routing address selection by preferring non-root, model-specific URL paths and using more reliable fallbacks when an exact match isn’t available.
* **Testing**
  * End-to-end tests now wait for MaaSModelRef readiness (phase **Ready** plus a populated endpoint) before querying `/v1/models`, reducing flaky timing-based checks.


test/e2e/tests/test_helper.py
```diff
@@ -877,6 +877,48 @@ def _wait_for_maas_auth_policy_phase(name, expected_phase="Active", namespace=No
...
+            status = cr.get("status", {})
+            phase = status.get("phase")
+            endpoint = status.get("endpoint")
+
+            if phase == "Ready" and endpoint:
+                log.info(f"MaaSModelRef {namespace}/{model_ref} is Ready with endpoint: {endpoint}")
+                return cr
+
+            log.debug(f"MaaSModelRef {namespace}/{model_ref}: phase={phase}, endpoint={endpoint or 'none'}")
+        time.sleep(2)
```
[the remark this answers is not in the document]
jrhyness: Will not fix here.   `_get_cr()` is used throughout.

maas-controller/pkg/controller/maas/providers_llmisvc.go
```diff
@@ -225,8 +225,26 @@ func (h *llmisvcHandler) getEndpointFromLLMISvc(llmisvc *kservev1alpha1.LLMInfer
...
+		// Prefer URLs with non-empty paths beyond just "/"
+		// Base URLs like https://host/ have path="/" (length 1)
+		// Model endpoints like https://host/ns/model have path="/ns/model" (length > 1)
+		if len(addr.URL.Path) > 1 && addr.URL.Path != "/" {
+			return addr.URL.String()
+		}
+		if fallbackURL == "" {
+			fallbackURL = addr.URL.String()
+		}
+	}
```
liangwen12year, line 245: Nit: if `Status.URL` is ever the only path-qualified URL while all `Addresses[]` are base URLs, this returns the base URL fallback before checking `Status.URL`. In practice gateway-internal always 
 carries the full path so this shouldn't matter — but wanted to flag it. Can you confirm Status.URL never has the model path when Addresses[] doesn't?

jrhyness: Thanks @liangwen12year , no, can't confirm that.   How does https://github.com/opendatahub-io/models-as-a-service/pull/1055/changes/5677d98388885804584a31ec13781389c10bc331 look?

dmytro-zaharnytskyi (approved):
lgtm, haven't found anything worth fixing or changing

maas-controller/pkg/controller/maas/providers_llmisvc.go
liangwen12year: lgtm

liangwen12year (approved):

PR #1062, opened 2026-06-29
chore: add e2e tests for body-based routing

ishitasequeira:
## Description
E2E tests for body based routing feature
https://redhat.atlassian.net/browse/RHOAIENG-63908
## How Has This Been Tested?
Executed the tests locally

## Merge criteria:

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for external models, including path-based routing validation (wrong model path is rejected).
  * Added body-based routing end-to-end tests for external and tenant models, verifying the `model` field is required and must match the provisioned value (wrong/missing values are rejected).
  * Added IPP readiness gating and a readiness helper so these body-routing suites run only when required gateway processing components are deployed and ready.
  * Standardized the external model target used in setup and cleanup for consistent test behavior.


test/e2e/tests/test_external_models.py
```diff
@@ -43,6 +43,9 @@
 EXTERNAL_AUTH_POLICY = os.environ.get("E2E_EXTERNAL_AUTH_POLICY", "e2e-external-access")
 RECONCILE_WAIT = int(os.environ.get("E2E_RECONCILE_WAIT", "12"))
 
+GATEWAY_NAMESPACE = os.environ.get("GATEWAY_NAMESPACE", "openshift-ingress")
+TARGET_MODEL = "gpt-3.5-turbo"
```
yu-teo, line 49: @ishitasequeira `TARGET_MODEL = "gpt-3.5-turbo"` is defined as a module-level constant, but the same value "gpt-3.5-turbo" is hardcoded in the `external_models_setup` fixture's ExternalModel CR. 
If someone changes the fixture's `targetModel`, the body-routing tests will fail for a non-obvious reason, isn't that right? Should we avoid introducing such duplication of the variable?

test/e2e/tests/test_external_models.py
```diff
@@ -285,6 +304,102 @@ def test_request_forwarded_returns_200(self, external_models_setup):
...
+@requires_ipp
+class TestExternalModelBodyRouting:
+    """Verify that the model field in the request body drives routing.
+
+    IPP pre-processing extracts the ``model`` field from the JSON body and
+    sets the ``X-Gateway-Model-Name`` header.  HTTPRoute Rule 2 matches on
+    that header (exact: targetModel) and routes to the backend.
+    """
+
+    def test_body_model_routes_correctly(self, external_models_setup):
```
yu-teo: How exactly do we verify routing with this test? I see it asserts `status_code` not in `(401, 403)`, which is the same approach taken like already existing `test_request_forwarded_returns_200` egress test. I agree it checks whether auth passed, but not that body-based routing (IPP extracting the model field into X-Gateway-Model-Name) was the mechanism. Therefore, we need to check the response body or a routing-indicative header to avoid this test passing with **path-based** routing.

test/e2e/tests/test_external_models.py
```diff
@@ -62,6 +65,22 @@ def _patch_cr(kind: str, name: str, namespace: str, patch: dict):
...
 
+def _check_payload_processing_pods():
+    """Check if payload-processing (IPP) pods are deployed."""
+    for name in ("payload-pre-processing", "payload-processing"):
+        result = subprocess.run(
+            ["oc", "get", "deployment", name, "-n", GATEWAY_NAMESPACE,
+             "-o", "jsonpath={.status.readyReplicas}"],
+            capture_output=True, text=True,
+        )
+        if result.returncode != 0:
```
yu-teo: If `oc` isn't on `PATH` or the `kubeconfig` is wrong, `returncode != 0` just returns `False`. Hence, the test will just skip. Should we consider adding a log message of sorts (maybe `log.debug` with `result.stderr` will help debugging)?

yu-teo:
@ishitasequeira just a general question about how we structure these tests.
I notice we repeat the `URL construction`, `headers dict`, and `body setup` code in all three methods. If am not wrong, the only thing that changes between tests is the `model` field in the body (present, correct, wrong, or absent), correct?
Maybe we can implement these tests slightly different to cut down on the boilerplate for the setups?
A class-level helper would collapse it:
```
  class TestExternalModelBodyRouting:
      def _post_chat(self, setup, body):
          url = (
              f"{setup['gateway_url']}/{MODEL_NAMESPACE}/{EXTERNAL_MODEL_NAME}"
              f"/v1/chat/completions"
          )
          headers = {
              "Content-Type": "application/json",
              "Authorization": f"Bearer {setup['api_key']}",
          }
          return requests.post(url, headers=headers, json=body, timeout=30, verify=TLS_VERIFY)
```
So we can perform testing as follows?
```
      def test_body_model_routes_correctly(self, external_models_setup):
          r = self._post_chat(external_models_setup, {
              "model": TARGET_MODEL,
              "messages": [{"role": "user", "content": "hello"}],
          })
          assert r.status_code not in (401, 403)
```
What do you think? This would be helpful as long as we URL pattern does not change. So if I missed that part, then disregard this comment :)

test/e2e/tests/test_tenant_model_inference.py
```diff
@@ -274,3 +279,149 @@ def test_tenant_isolation_cross_gateway_blocked(self, tenant_inference_cases):
...
+        _wait_reconcile()
+
+        r = self._post_chat(gateway_url, case_a["model_path"], api_key, {
+            "model": "nonexistent-model",
+            "messages": [{"role": "user", "content": "hello"}],
+        })
+        assert r.status_code != 200, (
+            f"Expected rejection for wrong model in body, got 200. "
+            f"Body routing may not be active — request succeeded via path routing alone."
+        )
```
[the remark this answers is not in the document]
ishitasequeira: The expected behavior is to test this specifically for not receiving the right response only.

test/e2e/tests/test_tenant_model_inference.py
```diff
@@ -274,3 +279,149 @@ def test_tenant_isolation_cross_gateway_blocked(self, tenant_inference_cases):
...
+# ─── Body-based routing ────────────────────────────────────────────────────
+
+
+def _check_payload_processing_pods():
+    """Check if payload-processing (IPP) pods are deployed."""
+    for name in ("payload-pre-processing", "payload-processing"):
+        result = subprocess.run(
+            ["oc", "get", "deployment", name, "-n", GATEWAY_NAMESPACE,
+             "-o", "jsonpath={.status.readyReplicas}"],
+            capture_output=True, text=True,
```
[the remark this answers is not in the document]
ishitasequeira: The test is expected to be executed on openshift clusters where oc is present in PATH by default.

jland-redhat (approved):
lgtm

jland-redhat (dismissed):
lgtm

yu-teo (dismissed):
lgtm

jland-redhat (approved):
lgtm

PR #1069, opened 2026-06-30
fix: resolve body-routed model names in gateway AuthPolicy and subscription validation

### LINKED ISSUE #1068: fix: gateway-level AuthPolicy and subscription validation fail for body-routed model names
## Problem

When requests arrive at a single URL (e.g., `POST /v1/messages`) and the model is identified only by the `X-Gateway-Model-Name` header (set by the pre-auth ext_proc from the request body), two components fail to match the model:

### 1. Gateway AuthPolicy `model_access` Rego map
The `aggregateModelSubjectAllowlists` function builds keys using `namespace/name` format (e.g., `llm/claude-opus-4-8`), but the `model_identity` computed from the `X-Gateway-Model-Name` header is the raw model name (e.g., `claude-opus-4-8`). No match → request denied.

### 2. MaaS API `subscriptionIncludesModel`
The function expects `requestedModel` in `namespace/name` format. When it receives a raw model name (no `/`), `len(parts) != 2` → returns `false` → subscription validation fails.

## Impact
Body-routed requests (single-URL pattern via `ANTHROPIC_BASE_URL=/v1`) are blocked with 403 even when the user has valid group membership and subscription access.

### maas-controller
In `aggregateModelSubjectAllowlists`, after building the `namespace/name` keyed map, also add entries keyed by the ExternalModel's `spec.modelName` (resolved via unstructured lookup on `inference.opendatahub.io` ExternalModels).

### maas-api
- Add `ModelName` field to `ModelRefInfo` (populated from ExternalModel `spec.modelName` via a new `ExternalModelNameResolver`)
- Update `subscriptionIncludesModel` to also match by `ModelRefInfo.ModelName`
- Wire `ExternalModelNameResolver` (backed by dynamic client) into the subscription `Selector`

## Related
- Depends on pre-auth ext_proc: opendatahub-io/models-as-a-service#948
- Depends on `inference.opendatahub.io` ExternalModel fallback: opendatahub-io/models-as-a-service#1049
- Related IPP PR for body-name resolution: opendatahub-io/ai-gateway-payload-processing#359

noyitz:
## Summary
- maas-controller: `aggregateModelSubjectAllowlists` adds model name aliases (e.g., `claude-opus-4-8`) alongside `namespace/name` keys in the `model_access` Rego map
- maas-api: `subscriptionIncludesModel` now matches raw model names via `ExternalModelNameResolver`

## Problem
Fixes #1068 — body-routed requests (single-URL `/v1/messages` pattern) fail with 403 because the gateway AuthPolicy and subscription validation only match `namespace/name` model identifiers, not raw model names from `X-Gateway-Model-Name`.

## How Has This Been Tested?
Deployed and verified on sandbox311 cluster:
- All 3 Claude models (Opus, Sonnet, Haiku) via `ANTHROPIC_BASE_URL=.../v1` — 200 OK
- Auth correctly injects `X-MaaS-Username` from API key validation
- Subscription validation passes with raw model names
- Metering records usage with correct customer identity

## Risk analysis
- **Risk rating**: 3
- **Why**: Changes auth policy Rego generation and subscription validation — core auth path. Well-tested on live cluster but no unit test coverage yet for the new code paths.

## Merge criteria
- [x] Testing instructions have been added in the PR body.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model selection for aliases and names containing `/`, including body-routed model names.
  - Fixed health checks for aliased models when subscriptions are degraded.
  - Improved validation messages when a requested model does not match the subscription.
  - Corrected readiness enforcement for aliased models.
- **Enhancements**
  - Model reference responses can now include the model’s descriptive name.
  - Authorization allowlists now correctly recognize configured model aliases.


noyitz:
Addressed all 5 review findings in 112bf8a9:

- **Duplicate aliases merged** — aliases are now collected separately and unioned (dedup + sort) instead of first-writer-wins, so multiple ExternalModel refs sharing a `spec.modelName` keep all subjects. Covered by `TestAggregateModelSubjectAllowlistsModelNameAliases`.
- **`spec.modelName` validated** — via the existing `validateCELValue` before it becomes a `model_access` key; covered by `TestAggregateModelSubjectAllowlistsRejectsUnsafeModelName`.
- **Lookup failures propagated** — `resolveModelName` now returns errors so the reconcile retries; NotFound/missing-CRD remain benign (no alias). RBAC marker updated to `get;list` — note the aggregated role already granted `get` via another controller's marker, so no generated-manifest diff.
- **Degraded health-check alias path** — `checkModelHealth` resolves the requested model (namespace/name or body-routed raw name) to its canonical ref via a shared `findModelRef` helper, so aliases get the same TRLP check instead of `InvalidModelFormat`. Covered by `TestCheckModelHealthBodyRoutedAlias` + `TestFindModelRef` (incl. raw model names containing `/`).
- **Context threading** — `ResolveModelName` now takes a `ctx`; enrichment lookups are bounded by a 5s timeout since Selector's public API doesn't carry the request context (per the review's alternative). Threading the request context through the full Selector/handler surface is a wider refactor than this PR should carry.

Both modules build; full `go test ./...` passes in maas-api and maas-controller.

maas-api/internal/models/external_model_resolver.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	gvr := schema.GroupVersionResource{
+		Group:    "inference.opendatahub.io",
+		Version:  "v1alpha1",
+		Resource: "externalmodels",
+	}
+	obj, err := r.client.Resource(gvr).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return ""
+	}
```
ishitasequeira: Shall we log the error here to help with debugging?

noyitz: Good call — added logging on both error paths (commit 93e763db): a warning when the ExternalModel lookup fails, and a warning when `spec.modelName` can't be read. Also added a logger to the resolver struct so it follows the same pattern as the rest of the maas-api codebase.

maas-api/internal/models/external_model_resolver.go
```diff
@@ -0,0 +1,49 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/logger"
+)
+
+// ExternalModelResolver resolves ExternalModel CRD names to spec.modelName.
+type ExternalModelResolver struct {
```
jland-redhat: Does `maas-api` have permissions to get ExternalModels?

jland-redhat (changes requested):
I am not sure it makes sense to do add a live lookup of an ExternalModel to the inference path.

The way we are doing this with internal models is by populating the information inside of the status and the doing our validation against that status information since we are already looking up the MaaSModelRef anyway.

Can we keep this same basic flow for ExternalModels?


```
Internal Model Flow

Client body "model" → ext_proc promotes it to X-Gateway-Model-Name
AuthPolicy Rego/CEL reads that header (path fallback for /ns/name/...)
Lookup: model_access[header_value]
Control plane (already for LLMISVC)

ModelRef reconciler sets status.resolvedModelAlias (publisher ID from LLMISVC status)
AuthPolicy reconcile copies that into model_access as an alt key via resolveHeaderModelKeys
Also keeps the namespace/name key for path routing
So validation is: header value == one of the model_access keys, and those alt keys come from ModelRef status — not a live ExternalModel watch.
```

noyitz:
@jland-redhat Addressed your review — removed the live ExternalModel lookup entirely and aligned with the internal model flow:

- **Deleted** `external_model_resolver.go` (the data-plane live lookup)
- **Removed** `WithExternalModelResolver` wiring from `main.go`
- **Simplified** `enrichModelRefs` in `selector.go` — no more live resolver call

The control plane already handles it: `externalModelHandler.ResolveModelAlias()` sets `status.resolvedModelAlias` during reconciliation, `resolveHeaderModelKeys` adds it as an alt key in the auth policy `model_access` map, and `resolveModelAlias()` in the selector maps body model names back to `namespace/name` using the MaaSModelRef cache. Same flow as LLMInferenceService.

maas-api no longer needs RBAC permissions on ExternalModel CRDs.

Net change: -84 lines, +5 lines.

maas-api/internal/subscription/types.go
```diff
@@ -59,6 +59,7 @@ type SubscriptionInfo struct {
 type ModelRefInfo struct {
 	Name            string           `json:"name"`
 	Namespace       string           `json:"-"`
+	ModelName       string           `json:"model_name,omitempty"`
```
ishitasequeira, line 62: `ModelRefInfo.ModelName` is used in `findModelRef()` as a fallback match, but after the `ExternalModelResolver` removal in `ca87e3c6`, nothing populates it.. `enrichModelRefs()` doesn't set it. The tests work because they hardcode ModelName in the fixtures, so the gap isn't surfaced.

Since `enrichModelRefs()` already looks up each ref in the MaaSModelRef index, and that index carries `status.resolvedModelAlias`, would it make sense to wire ModelName from there? That'd be a one-liner in enrichModelRefs().

Separately, is there a scenario where ModelName would come from user input on the subscription spec rather than being derived from the model CR? Just want to make sure we're not building two paths to set the same thing.

jland-redhat:
So I am seeing a similar thing as @ishitasequeira which I am not totally following how the MaaSModelRef is getting the correct model name.

Looking through our E2E test I am noticing the only one that validate ExternalModels authorization passes correctly uses path based routing. 

Could we please add at least one test that validates the BodyBasedRouting works?

And preferably a separate test validating the alias is populated (just so it is obvious if that is the part that is failing in the future)

noyitz:
@jland-redhat Thanks for the review. Two things addressed:

1. **Ishita's ModelName gap fixed** (commit `9b1dfa63`): `enrichModelRefs()` now populates `ModelName` from `MaaSModelRef.status.resolvedModelAlias` — the control-plane path you asked for.

2. **E2e tests**: `test_external_models.py` already has a `TestExternalModelBodyRouting` class that validates body-based routing (wrong model rejected, missing model rejected). For the alias population test — I can add a unit test in `maas-controller` that verifies `externalModelHandler.ResolveModelAlias()` returns the correct value and it ends up in `status.resolvedModelAlias`. Would that cover what you're looking for, or do you want it at the e2e level?

maas-api/internal/subscription/types.go
noyitz: Good catch — you're right, `findModelRef()` and `ModelName` are part of this PR and nothing populated `ModelName` after the `ExternalModelResolver` removal.

Fixed in `9b1dfa63`: `enrichModelRefs()` now populates `ModelName` from `status.resolvedModelAlias` in the MaaSModelRef informer cache — exactly as you suggested. One-liner, same index lookup that's already there.

To your second question: no, `ModelName` never comes from user input on the subscription spec. It's always derived from the model CR (set by the controller during reconciliation). Single path.

maas-api/internal/models/external_model_resolver.go
noyitz: No longer needed — the live ExternalModel lookup was removed in `ca87e3c6`. maas-api now relies entirely on `MaaSModelRef.status.resolvedModelAlias` from the informer cache (populated by the controller at reconcile time). No RBAC on ExternalModel CRDs required.

noyitz:
@jland-redhat Here's a testing guide with working simulator CRDs and curl commands:

**PR:** https://github.com/opendatahub-io/ai-gateway-payload-processing/pull/414

It has ExternalModel + ExternalProvider + Secret manifests for the shared simulator (`3-132-132-211.sslip.io`) covering:
- `sim-chat` — OpenAI Chat Completions (`/v1/chat/completions`)
- `sim-messages` — Anthropic Messages (`/v1/messages`)
- Curl commands for path-based routing, body-based routing (BBR), streaming, and wrong-model rejection

The simulator supports openai and anthropic provider types out of the box (no GPU, no real API keys needed — uses `llm-katan-openai-key` / `llm-katan-anthropic-key`).

jland-redhat (approved):
lgtm

PR #1070, opened 2026-06-30
fix: scope gateway-default-auth to model inference paths only

ishitasequeira:
## Description
- Scopes `gateway-default-auth` deny-all to model inference paths only by adding a `when` predicate with `celModelIdentityAvailable`, so management endpoints (`/v1/*`, `/maas-api/*`) remain accessible on clusters with zero MaaSAuthPolicy CRs
- Adds unit test verifying the recreated policy includes the `when` predicate
- Adds e2e tests validating OPA rego allows empty `model_identity`, `subscription-valid` is gated by model identity, and `gateway-default-auth` structure when present                                                

Fixes: https://redhat.atlassian.net/browse/RHOAIENG-72349

## Test plan                                                          
- [x] Unit: `TestMaaSAuthPolicyReconciler_MultiplePoliciesDeletion` passes
- [x] E2E: `TestGatewayAuthPolicyManagementEndpointAccess` passes
- [x] Manual: deploy on fresh cluster with no MaaSAuthPolicy, confirm API Keys page loads without 403                  

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default gateway auth policy so deny-all only applies to model inference routes, while management endpoints remain accessible when no auth policies are configured.
  * On controller startup, automatically upgrades or recreates the existing `gateway-default-auth` policy to include the correct inference scoping behavior.
  * Improved gateway propagation retries during API key setup to handle transient auth/gateway readiness races.
* **Tests**
  * Extended controller unit tests to verify the policy `when` predicate is present and exactly scoped to model identity availability.
  * Added/expanded E2E coverage to validate gateway policy rego and management endpoint exclusions, plus the updated retry behavior.


maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
```diff
@@ -1464,6 +1464,8 @@ func (r *MaaSAuthPolicyReconciler) deleteGatewayDefaultAuthPolicy(ctx context.Co
 
 // ensureGatewayDefaultAuthPolicy recreates the static deny-all gateway-default-auth policy
 // after the last MaaSAuthPolicy is removed, so unconfigured model routes remain denied.
```
jland-redhat, line 1467: Sounds like this may only get applied if an AuthPolicy gets removed, will we apply this policy on startup and override the old policy?

test/e2e/tests/test_gateway_scoped_authpolicy.py
```diff
@@ -120,3 +121,63 @@ def test_only_one_gateway_authpolicy_named_maas_gateway_auth(self):
...
+            "subscription-valid must have a 'when' predicate so it only runs for "
+            "model inference, not management endpoints"
+        )
+
+    def test_gateway_default_auth_scoped_if_present(self):
+        """If gateway-default-auth exists, it must scope deny-all to model paths only."""
+        default_auth = get_json_or_none(
+            "authpolicy", "gateway-default-auth", GATEWAY_NAMESPACE
+        )
+        if default_auth is None:
```
dmytro-zaharnytskyi, line 169: If I remember right in CI we deploy MaaSAuthPolicy, etc, so gateway-default-auth is gone, hence test_gateway_default_auth_scoped_if_present skips every run, and the other two tests only inspect maas-gateway-auth. So I wonder whether we need to add a setup where gateway-default-auth is actually present (delete all auth policies first, for example)?

maas-controller/pkg/controller/maas/maasauthpolicy_controller.go
dmytro-zaharnytskyi, line 1480: Seems like an existing gateway-default-auth without spec.defaults.when never gets patched. The current unit test pre-creates a bare policy, but reconciliation then deletes it when maas-gateway-auth is created and only checks the later recreate path. So maybe we need to patch the existing policy when the predicate is missing or stale, and add a test for that path?

jland-redhat (dismissed):
Lgtm

chaitanya1731 (approved):
lgtm

PR #1073, opened 2026-06-30
feat: add debounce last update at api keys

ishitasequeira:
## Problem
A customer running 256 concurrent requests against a single API key hit a `context deadline exceeded` while maas-api tried to update `last_used_at` in Postgres.  Every successful validation — including for ephemeral keys — fires an async goroutine that issues:

```sql
UPDATE api_keys SET last_used_at = $1 WHERE id = $2 AND tenant = $3
```

With 256 goroutines racing to UPDATE the same row, Postgres row-level locking causes all but the first to queue up. The 2-second goroutine timeout expires before most of them acquire the lock.

## Solution
Add a lightweight in-memory debounce to Service (sync.Map keyed by key ID) so only one UpdateLastUsed goroutine is spawned per key per configurable window instead of one per request.
```
Authorino cache (60 s TTL)
  └─ validate called on cache miss
       ├─ [SYNC]  GetByHash → SELECT (unchanged)
       └─ [ASYNC, debounced] UpdateLastUsed → UPDATE  ← at most 1/window per key
```
**Key properties:**
- Default window: 60 s (LAST_USED_DEBOUNCE_SECS=60), matching Authorino's metadata-cache TTL — no observable loss of last_used_at precision relative to what callers already see.
- LAST_USED_DEBOUNCE_SECS=0 disables debouncing (write on every validation) for backward compatibility.
- If the write fails, the debounce entry is evicted so the next request retries — no permanent suppression on errors.
- sync.Map.LoadOrStore is atomic, so the "first caller wins" selection is race-free.
[RHOAIENG-72575](https://redhat.atlassian.net/browse/RHOAIENG-72575)

## How Has This Been Tested?
- Local manual testing

## Merge criteria:

## Summary

* **New Features**
  * API key validation now refreshes `last_used_at` in a throttled background manner to reduce backend contention during high traffic.
  * Added `LAST_USED_DEBOUNCE_SECS` (`0` disables throttling; default is `60`) to control the minimum time between `last_used_at` updates per key.

* **Bug Fixes**
  * Improved consistency of `last_used_at` updates under concurrent validations, including recovery when background updates fail.

* **Documentation**
  * Updated API key authentication docs to explain the debounced behavior and configuration override.

* **Tests**
  * Added coverage for throttled vs non-throttled and TTL-expiry scenarios.

### DIFF
--- a/docs/content/concepts/api-key-authentication.md
+++ b/docs/content/concepts/api-key-authentication.md
@@ -97,6 +97,8 @@ The validation endpoint (`/internal/v1/api-keys/validate`) is called by Authorin
 2. Returns `valid: true` with `userId`, `groups`, and `subscription` if the key is active and not expired
 3. Returns `valid: false` with a reason if the key is invalid, revoked, or expired
 
+After a successful lookup, maas-api asynchronously updates `last_used_at` in the background. To prevent Postgres row-lock contention when many requests share a single key, these writes are **debounced**: at most one write is issued per key per `LAST_USED_DEBOUNCE_SECS` window (default 60 s). Set `LAST_USED_DEBOUNCE_SECS=0` on the maas-api Deployment to write on every validation.
+
 ---
 
 ## Subscription Binding and Priority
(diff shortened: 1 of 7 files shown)


jland-redhat (dismissed):
lgtm

jland-redhat (approved):
lgtm

PR #1075, opened 2026-06-30
fix: prevent crash-loop when Kuadrant or KServe CRDs are not installed

somya-bhatnagar:
## Summary

Prevent maas-controller from crash-looping when Kuadrant (AuthPolicy, TokenRateLimitPolicy) or KServe (LLMInferenceService) CRDs are not yet registered at startup. These are required dependencies — the controller should start cleanly and register the watches dynamically when the CRDs become available.

## Problem

Three controllers registered watches for Kuadrant and KServe CRDs unconditionally at startup:

| Controller | CRD | Dependency |
|---|---|---|
| `maassubscription_controller.go` | `TokenRateLimitPolicy` (kuadrant.io) | Kuadrant |
| `maasauthpolicy_controller.go` | `AuthPolicy` (kuadrant.io) | Kuadrant |
| `maasmodelref_controller.go` | `LLMInferenceService` (serving.kserve.io) | KServe |

When any CRD is absent, controller-runtime times out waiting for the informer cache to sync (~30s) and crashes the process — causing CrashLoopBackOff. The Tenant reconciler never runs so no status is surfaced.

### `crdExists(ctx, reader, crdName string) bool`

Uses a targeted `reader.Get` by CRD name (e.g. `authpolicies.kuadrant.io`) instead of listing all CRDs and iterating. Uses `apierrors.IsNotFound` for the expected absent-CRD case and logs unexpected errors (RBAC denials, discovery failures) rather than silently swallowing them.

Uses `mgr.GetAPIReader()` (bypasses cache) since `SetupWithManager` runs before the manager cache starts.

### `registerWatchWhenCRDAppears(c, mgr, crdName, makeSource)`

Replaces the previous `watchForCRDAndRestart` (poll every 30s + `os.Exit(0)`) with a fully dynamic approach:

- Watches `apiextensionsv1.CustomResourceDefinition` objects (always available — built into the Kubernetes API)
- Filters by CRD name; calls `c.Watch(makeSource())` via `sync.Once` the first time the target CRD appears
- `sync.Once` guarantees exactly one watch registration regardless of how many CRD update events fire
- No pod restart — other controllers keep running uninterrupted

### Dynamic LLMInferenceService watch — predicates + bug fix

The dynamic path uses `*unstructured.Unstructured` (bypasses stale REST mapper cache when CRD is installed after startup) with `handler.TypedFuncs` to mirror the static-path predicates:

- **`UpdateFunc`**: skips if generation unchanged AND ready-status unchanged — mirrors `GenerationChangedPredicate` + `llmisvcReadyChangedPredicate`
- **`CreateFunc` / `DeleteFunc`**: always enqueue (same as static path)

**Critical bug fixed:** `mapLLMISvcToMaaSModelRefs` previously type-asserted `obj.(*kservev1alpha1.LLMInferenceService)` — always nil for unstructured objects, silently dropping all events on the dynamic watch. Fixed to use `obj.GetName()`/`obj.GetNamespace()` which work for both typed and unstructured.

### SetupWithManager (all three controllers)

- If CRD present at startup: watch registered normally via builder (static path)
- If CRD absent: log message + `registerWatchWhenCRDAppears` registered; watch added dynamically when dependency becomes available
- Uses `b.Build(r)` (instead of `b.Complete(r)`) to obtain the `controller.Controller` needed for runtime `c.Watch()` calls

### Unit tests added
- `TestCrdExists_Found/NotFound/OtherError` — crdExists helper coverage
- `TestRegisterWatchWhenCRDAppears_SyncOnce` — sync.Once fires exactly once
- `TestRegisterWatchWhenCRDAppears_FiltersByName` — name filter prevents wrong CRD
- `TestRegisterWatchWhenCRDAppears_OnlyFiredForMatchingCRD` — combined filter + once
- `TestMapLLMISvcToMaaSModelRefs_UnstructuredObject` — regression test for the unstructured mapping bug

### E2E tests added (`test/e2e/tests/test_crd_watch_resilience.py`) — cluster-verified (3/3 pass)
- `test_controller_stable_without_kserve_crd`: controller stable with KServe absent; asserts unrelated Kuadrant CRD events do NOT trigger the KServe watcher (validates name filter)
- `test_dynamic_watch_fires_exactly_once_on_crd_install`: after installing KServe CRD, watch registered exactly once; 3 subsequent CRD annotation events produce 0 additional registrations (sync.Once verified)
- `test_static_watch_registered_when_crd_present_at_startup`: no dynamic watcher when CRD present at controller startup

## Risk Analysis

**Risk rating:** 2 — replaces polling+restart mechanism with dynamic watch registration. When CRDs are present at startup, behavior is identical to before. When absent, controllers start cleanly instead of crash-looping.

## Related
https://redhat.atlassian.net/browse/RHOAIENG-76525


maas-controller/pkg/controller/maas/maasmodelref_controller.go
```diff
@@ -411,13 +411,21 @@ func llmisvcReadyStatus(obj *kservev1alpha1.LLMInferenceService) string {
 }
 
 // SetupWithManager sets up the controller with the Manager.
+// crdExists checks if a CRD is registered in the cluster's REST mapper.
+// Used to make watches on optional CRDs (KServe, Kuadrant) conditional so the
+// controller can start when those operators are not installed.
+func crdExists(mapper apimeta.RESTMapper, group, version, kind string) bool {
+	_, err := mapper.RESTMapping(schema.GroupKind{Group: group, Kind: kind}, version)
+	return err == nil
+}
```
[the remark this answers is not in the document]
somya-bhatnagar: done

ishitasequeira (commented):
Could we avoid restarting the pod entirely?                                                                                                                               
                                                                                                                                                                              
`watchForCRDAndRestart` calls `os.Exit(0)` when the CRD appears, which disrupts all controllers, drops in-flight reconciliations, and churns leader election — even though only one watch needs to be added. os.Exit(0) also bypasses deferred cleanup and graceful shutdown.
                                                                                                                                                                              
controller-runtime's `controller.Controller` interface exposes a `Watch()` method that can be called at runtime, not just during SetupWithManager. Would it be possible to watch the CRD object itself (apiextensionsv1.CustomResourceDefinition — always available since it's built into the Kubernetes API) and dynamically register the real watch when the target CRD appears?                                                                                                                                                     
                                                                                                                                                                            
Something along the lines of:                                                                                                                                               
```  
  c, err := b.Build(r)                                                                                                                                                        
  if err != nil {                                                                                                                                                           
      return err                                                                                                                                                              
  }
                                                                                                                                                                              
  if !crdExists(ctx, mgr.GetAPIReader(), "kuadrant.io", "AuthPolicy") {                                                                                                       
      _ = c.Watch(source.Kind(mgr.GetCache(), &apiextensionsv1.CustomResourceDefinition{},
          handler.TypedEnqueueRequestsFromMapFunc(                                                                                                                            
              func(ctx context.Context, crd *apiextensionsv1.CustomResourceDefinition) []reconcile.Request {                                                                
                  if crd.Spec.Group == "kuadrant.io" && crd.Spec.Names.Kind == "AuthPolicy" {                                                                                 
                      authPolicy := &unstructured.Unstructured{}
                      authPolicy.SetGroupVersionKind(schema.GroupVersionKind{                                                                                                 
                          Group: "kuadrant.io", Version: "v1", Kind: "AuthPolicy",                                                                                          
                      })                                                                                                                                                      
                      if err := c.Watch(source.Kind(mgr.GetCache(), authPolicy,                                                                                             
                          handler.EnqueueRequestsFromMapFunc(r.mapGeneratedAuthPolicyToParent),                                                                               
                      )); err != nil {                                                                                                                                        
                          ctrl.Log.Error(err, "failed to register AuthPolicy watch after CRD appeared")                                                                       
                      }                                                                                                                                                       
                      return r.enqueueAllMaaSAuthPolicies(ctx)                                                                                                              
                  }                                                                                                                                                           
                  return nil                                                                                                                                                
              },                                                                                                                                                              
          ),                                                                                                                                                                
      ))
  }
```

This way no pod restart is needed — the watch registers the moment the CRD appears, other controllers keep running uninterrupted, and existing resources get reconciled immediately. The `apiextensionsv1` package is already imported for crdExists, so the scheme registration should already be in place.

A couple of smaller things on the current implementation (in case the restart approach is kept):                                                                            
  
* `_ = mgr.Add(...)` silently ignores the error: if it fails, the CRD watcher never runs and the controller never learns the CRD appeared. Worth logging.
* crdExists lists ALL CRDs every 30 seconds per missing CRD. A targeted reader.Get by CRD name (e.g., authpolicies.kuadrant.io) would be lighter, especially on OpenShift clusters with hundreds of CRDs.
* The PR description mentions `apimeta.IsNoMatchError` but the code uses direct CRD listing. Might want to update the description to match.

somya-bhatnagar:
Great suggestion from @ishitasequeira — replaced `os.Exit(0)` with dynamic `c.Watch()` registration.

**`watchForCRDAndRestart` → `registerWatchWhenCRDAppears`**

Instead of polling + pod restart, we now watch the `CustomResourceDefinition` object itself using `c.Watch(source.Kind(cache, &apiextensionsv1.CustomResourceDefinition{}, ...))`. When the target CRD appears, the handler calls `c.Watch(makeSource())` to register the real watch dynamically — no pod restart, no disruption to other controllers, no in-flight reconciliation drops.

**Smaller concerns addressed:**
- `_ = mgr.Add(...)` replaced with `c.Watch(...)` which properly returns and propagates the error
- All three controllers (`MaaSAuthPolicy`, `MaaSSubscription`, `MaaSModelRef`) now use `b.Build(r)` + `c.Watch()` pattern

**Tested on cluster:** Kuadrant CRDs installed after controller startup → watches registered dynamically with log `"CRD is now registered; watch added dynamically"` — no pod restart needed.

Thanks for the review!

maas-controller/pkg/controller/maas/maasmodelref_controller.go
```diff
@@ -411,13 +414,63 @@ func llmisvcReadyStatus(obj *kservev1alpha1.LLMInferenceService) string {
...
+// without restarting the pod — other controllers keep running uninterrupted.
+func registerWatchWhenCRDAppears(
+	c controller.Controller,
+	mgr ctrl.Manager,
+	group, kind string,
+	makeSource func() source.Source,
+) error {
+	log := ctrl.Log.WithName("crd-watcher").WithValues("group", group, "kind", kind)
+	log.Info("CRD not yet registered at startup; will register watch dynamically when it appears")
+	return c.Watch(source.Kind(
```
jland-redhat, line 449: So correct me if I am wrong but I think this may have the opposite problem. Every time the CRD changes this watch gets triggered and you could end up multiple watchers.


Not sure the best way to handle this since it is dynamic but we would want to make sure that we are only registering the watcher a single time.

maas-controller/pkg/controller/maas/maasmodelref_controller.go
```diff
@@ -411,13 +414,63 @@ func llmisvcReadyStatus(obj *kservev1alpha1.LLMInferenceService) string {
...
 
 // SetupWithManager sets up the controller with the Manager.
+// crdExists checks if a CRD is registered in the cluster via a targeted Get by CRD name.
+// Uses mgr.GetAPIReader() (not mgr.GetClient()) because SetupWithManager is called
+// before the manager's cache starts. The REST mapper is not used because scheme-registered
+// types cause false positives even when the CRD is absent.
+func crdExists(ctx context.Context, reader client.Reader, group, kind string) bool {
+	// CRD name format: <plural-lowercase-kind>.<group>
+	// We search by group+kind since we don't know the plural form.
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
```
jland-redhat: I am not sure getting every crd and looping through them is the best way. 

Can't we just get the CRD by the group/name since we already have that?

jland-redhat (changes requested):

somya-bhatnagar:
@coderabbitai review

maas-controller/pkg/controller/maas/maasmodelref_controller.go
somya-bhatnagar: Fixed. Added `sync.Once` inside the `registerWatchWhenCRDAppears` closure — `c.Watch(makeSource())` is now called exactly once regardless of how many CRD update events fire. The `once` variable is declared per call to the function, so each CRD type has its own independent guard.

maas-controller/pkg/controller/maas/maasmodelref_controller.go
somya-bhatnagar: Fixed. `crdExists` now uses a targeted `reader.Get` by CRD name (e.g. `authpolicies.kuadrant.io`) instead of listing all CRDs and iterating. Since we know the exact CRD names for our three dependencies, a direct Get is simpler and lighter — especially on OpenShift clusters with hundreds of CRDs.

somya-bhatnagar:
@jland-redhat addressed both review comments in the latest commit (`66a69694`):

1. **Multiple watchers** — added `sync.Once` inside `registerWatchWhenCRDAppears`. The `once` variable is declared per function call (one per CRD type), so `c.Watch(makeSource())` is called exactly once regardless of how many CRD update/annotate events fire.

2. **Listing all CRDs** — `crdExists` now uses a targeted `reader.Get` by CRD name (e.g. `authpolicies.kuadrant.io`) instead of `reader.List` + loop. We know the exact CRD names for all three dependencies so a direct Get is simpler and avoids scanning hundreds of CRDs on OpenShift clusters.

maas-controller/pkg/controller/maas/maasmodelref_controller.go
```diff
@@ -443,7 +506,25 @@ func (r *MaaSModelRefReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if !llmisvcExists {
+		if err := registerWatchWhenCRDAppears(c, mgr, llmisvcCRD, func() source.Source {
+			return source.Kind(mgr.GetCache(), &kservev1alpha1.LLMInferenceService{},
+				handler.TypedEnqueueRequestsFromMapFunc[*kservev1alpha1.LLMInferenceService](
+					func(ctx context.Context, obj *kservev1alpha1.LLMInferenceService) []reconcile.Request {
+						return r.mapLLMISvcToMaaSModelRefs(ctx, obj)
+					},
+				),
+			)
```
chaitanya1731, line 572: The static registration path for `LLMInferenceService` includes predicates (`GenerationChangedPredicate` + `llmisvcReadyChangedPredicate`) that gate reconciliation to generation changes and ready-status flips:

```go
b = b.Watches(&kservev1alpha1.LLMInferenceService{},
    handler.EnqueueRequestsFromMapFunc(r.mapLLMISvcToMaaSModelRefs),
    builder.WithPredicates(predicate.Or(
        predicate.GenerationChangedPredicate{},
        llmisvcReadyChangedPredicate{},
    )),
)
```

But this dynamic path has no predicates — it will trigger reconciliation on **every** `LLMInferenceService` update event, which could cause significant unnecessary reconciliation load.

`source.Kind` accepts variadic predicates, so this should be straightforward to fix:

```go
return source.Kind(mgr.GetCache(), &kservev1alpha1.LLMInferenceService{},
    handler.TypedEnqueueRequestsFromMapFunc[*kservev1alpha1.LLMInferenceService](...),
    predicate.TypedOr[*kservev1alpha1.LLMInferenceService](
        predicate.TypedGenerationChangedPredicate[*kservev1alpha1.LLMInferenceService]{},
        // llmisvcReadyChangedPredicate equivalent for typed predicate
    ),
)
```

chaitanya1731 (commented):
Added one AI comment the bug that it found. Can you check.. ?

Also wondering since we added new helper functions.. `crdExists` and `registerWatchWhenCRDAppears`  should we add test coverage for these too.. and may be for sync.once behavior too? right now no unit tests exist for these.

jland-redhat (changes requested):
There are a couple unanswer questions from folks on this PR.

Could we make sure to validate a few things? 

Firstly and most important can we make sure to throw some e2e test around deleting/creating the relevant CRDs and making sure that the controller still functions properly

And can we just make sure to do manual validation that only trigger `registerWatchWhenCRDAppears` on the creation of the appropriate CRDs?

maas-controller/pkg/controller/maas/maasmodelref_controller.go
somya-bhatnagar: Fixed. The dynamic path now uses `handler.TypedFuncs[*unstructured.Unstructured, reconcile.Request]` which gives access to both old and new objects in `UpdateFunc`, allowing the predicates to be mirrored properly:

- **`UpdateFunc`**: skips if generation unchanged AND ready-status unchanged — mirrors `GenerationChangedPredicate` + `llmisvcReadyChangedPredicate`
- **`CreateFunc` / `DeleteFunc`**: always enqueue (same as static path)
- **`unstructuredLLMIsvcReadyStatus`**: new helper that extracts the Ready condition from unstructured objects

Note: `source.Kind` in controller-runtime v0.22.5 doesn't accept variadic predicates — only `TypedEventHandler`. Using `handler.TypedFuncs` with the full event is the correct way to embed predicate logic.

Also fixed a critical bug discovered in the process: `mapLLMISvcToMaaSModelRefs` was type-asserting `obj.(*kservev1alpha1.LLMInferenceService)` which always returns nil for unstructured — meaning the watch registered correctly but never enqueued any reconciles. Fixed by using `obj.GetName()`/`obj.GetNamespace()` which work for both typed and unstructured. A regression test `TestMapLLMISvcToMaaSModelRefs_UnstructuredObject` locks this in.

somya-bhatnagar:
**Addressing @chaitanya1731 and @jland-redhat reviews — cluster-verified:**

**Unit tests added** (`maasmodelref_controller_test.go`):
- `TestCrdExists_Found/NotFound/OtherError` — crdExists helper coverage
- `TestRegisterWatchWhenCRDAppears_SyncOnce` — sync.Once fires exactly once
- `TestRegisterWatchWhenCRDAppears_FiltersByName` — name filter prevents wrong CRD
- `TestRegisterWatchWhenCRDAppears_OnlyFiredForMatchingCRD` — combined filter + once
- `TestMapLLMISvcToMaaSModelRefs_UnstructuredObject` — regression test for the critical bug where the typed `*kservev1alpha1.LLMInferenceService` assertion silently dropped all events on the dynamic watch path

**E2E tests added** (`test/e2e/tests/test_crd_watch_resilience.py`) — all 3 passed on cluster:
- `test_controller_stable_without_kserve_crd`: controller stable with KServe absent; explicitly asserts that unrelated Kuadrant CRD events do NOT trigger the KServe watcher (validates the name filter)
- `test_dynamic_watch_fires_exactly_once_on_crd_install`: after installing KServe CRD, "watch registered dynamically" appears once; 3 subsequent CRD annotation events produce 0 additional registrations — sync.Once verified
- `test_static_watch_registered_when_crd_present_at_startup`: after restart with KServe CRD present, no dynamic watcher is set up — static path used correctly

**On E2E scope (Kuadrant AuthPolicy/TRLP):** KServe is the safe representative for `registerWatchWhenCRDAppears` — all three CRDs use the same helper. AuthPolicy/TRLP delete/create E2E is blocked by a pre-existing issue (`ensureGatewayDefaultAuthPolicy` requires the CRD at startup, causing a crash unrelated to this PR). The filter validation is covered by asserting that unrelated Kuadrant CRD events don't trigger the KServe watcher. Deleting Kuadrant CRDs on a test cluster is destructive and adds little beyond what the shared helper + KServe E2E already validates.

ishitasequeira (approved):
lgtm

jland-redhat (approved):

PR #897, opened 2026-05-13
feat: configure OIDC JWKS refresh and authentication monitoring

jrhyness:
Summary
  
  Adds OIDC JWKS refresh configuration and operator alerting for authentication failures to address [RHOAIENG-60847](https://redhat.atlassian.net/browse/RHOAIENG-60847) (Define JWKS cache, IdP outage behavior, and operator alerting).

  Changes

  JWKS Refresh Configuration

  Wires Tenant.spec.externalOIDC.ttl to Authorino's JWT authentication configuration, enabling operators to control JWKS refresh frequency.

  What TTL controls:
  - How often Authorino fetches fresh JWKS from the IdP (when available)
  - Default: 300 seconds
  - Minimum: 30 seconds

  IdP outage behavior:
  - Cached JWKS persists indefinitely during IdP unavailability
  - Authentication continues using last successfully cached keys
  - Tokens signed with never-cached keys will fail validation

  Implementation:
  - Extract TTL from Tenant CR in maasauthpolicy_controller.go
  - Pass TTL to AuthPolicy → AuthConfig JWT configuration
  - Remove invalid audiences field (validation handled by authorization rules)

  OIDC Authentication Monitoring

  Adds PrometheusRules to alert operators when OIDC authentication issues occur.

  Alerts:

  1. MaaSAuthorinoOIDCAuthenticationHighFailureRate
    - Fires when: >10% auth failure rate for 5 minutes
    - Indicates: IdP down, JWKS unreachable, token validation failures
  2. MaaSAuthorinoOIDCAuthenticationHighLatency
    - Fires when: P95 auth latency >2 seconds for 5 minutes
    - Indicates: Slow IdP, network issues, excessive JWKS fetches

  Implementation:
  - Add authorino-maas-oidc-prometheusrule.yaml to deployment/base/observability/
  - Deploy via install-observability.sh to Authorino namespace
  - Fix Limitador ServiceMonitor namespace substitution for RHOAI deployments

  Testing

  Environment: RHOAI 3.4.0 on OpenShift 4.19 with Keycloak test realm

  TTL Configuration:
  - Verified TTL propagates from Tenant CR (ttl: 60) to AuthConfig (jwt.ttl: 60)
  - Configured test tenant with 60-second TTL for faster validation

  JWKS Caching During IdP Outage:
  - Scaled down Keycloak to simulate IdP outage
  - Reused valid JWT token obtained before outage
  - Authentication succeeded for 5+ minutes using cached JWKS (well beyond 60s TTL)
[...]

PR #1076, opened 2026-07-01
feat: wire up OIDC JWKS cache TTL from Tenant CR to Authorino

ishitasequeira:
## Description
Revamp https://github.com/opendatahub-io/models-as-a-service/pull/897 on the latest main branch.
https://redhat.atlassian.net/browse/RHOAIENG-60847
## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added tenant-specific External OIDC JWKS cache TTL controls (defaults/minimum) and enforced `azp`/clientId-bound OIDC JWT authorization.
  * Added two new gateway authentication Prometheus alerts (high failure rate and high P95 latency).
* **Documentation**
  * Published full External OIDC administration guidance, including monitoring and security responsibilities.
* **Bug Fixes**
  * Improved observability installation to conditionally apply the new OIDC alert rules only when applicable.
* **Tests**
  * Expanded unit and end-to-end coverage for TTL propagation, client binding, group safety, direct model access, alerting presence, and API key ownership.

### DIFF
--- a/deployment/base/observability/authorino-maas-oidc-prometheusrule.yaml
+++ b/deployment/base/observability/authorino-maas-oidc-prometheusrule.yaml
@@ -1,0 +1,52 @@
+# PrometheusRule: alert when Authorino OIDC/JWT authentication fails often or becomes slow.
+#
+# Scraped metrics come from Authorino /server-metrics (see authorino-server-metrics-servicemonitor.yaml).
+# Requires Tenant.spec.externalOIDC configured (MaaS controller creates AuthConfig with JWT auth).
+#
+# Deployed by: scripts/observability/install-observability.sh (NOT kustomize base)
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: authorino-maas-oidc-authentication-high-failure-rate
+  namespace: kuadrant-system
+  labels:
+    app.kubernetes.io/part-of: maas-observability
+    app.kubernetes.io/component: alerting
+spec:
+  groups:
+  - name: maas.authorino.oidc
+    rules:
+    - alert: MaaSAuthorinoOIDCAuthenticationHighFailureRate
+      annotations:
+        summary: Authorino OIDC/JWT authentication is failing often
+        description: >-
+          Over the last 5 minutes, more than 10% of Authorino OIDC/JWT authentication attempts
+          resulted in UNAUTHENTICATED status. Investigate IdP (Keycloak/OIDC provider) health,
+          JWKS endpoint reachability, JWKS cache TTL configuration, and token validity.
+      expr: |
+        (
+          sum(rate(auth_server_authconfig_response_status{namespace=~"rh-connectivity-link|kuadrant-system",status="UNAUTHENTICATED"}[5m]))
+          /
+          sum(rate(auth_server_authconfig_response_status{namespace=~"rh-connectivity-link|kuadrant-system"}[5m]))
+        ) > 0.10
+        and
+        sum(rate(auth_server_authconfig_response_status{namespace=~"rh-connectivity-link|kuadrant-system"}[5m])) > 0.001
+      for: 5m
+      labels:
+        severity: warning
+    - alert: MaaSAuthorinoOIDCAuthenticationHighLatency
+      annotations:
+        summary: Authorino OIDC/JWT authentication latency is high
+        description: >-
+          Over the last 5 minutes, P95 latency for Authorino OIDC/JWT authentication
+          exceeded 2 seconds. Investigate IdP performance, network latency to JWKS endpoint,
+          and consider increasing JWKS cache TTL if IdP is slow but reliable.
+      expr: |
+        histogram_quantile(0.95,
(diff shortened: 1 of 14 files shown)


maas-controller/pkg/controller/maas/maasauthpolicy_controller_test.go
EgorLu, line 1403: Does this test pass for you?
Don't we need to pass TTL (which previously was 300 but now may default to 0)?

deployment/base/observability/authorino-maas-oidc-prometheusrule.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - alert: MaaSAuthorinoOIDCAuthenticationHighFailureRate
+      annotations:
+        summary: Authorino OIDC/JWT authentication is failing often
+        description: >-
+          Over the last 5 minutes, more than 10% of Authorino OIDC/JWT authentication attempts
+          resulted in UNAUTHENTICATED status. Investigate IdP (Keycloak/OIDC provider) health,
+          JWKS endpoint reachability, JWKS cache TTL configuration, and token validity.
+      expr: |
+        (
+          sum(rate(auth_server_authconfig_response_status{namespace=~"rh-connectivity-link|kuadrant-system",status="UNAUTHENTICATED"}[5m]))
```
EgorLu, line 30: Does Authorino provide a label for OIDC that we could filter on?
Otherwise this will fire for any other failed authentication methods (API, Token, etc), but since the metric is called "...OIDC...", it will make us think that OIDC is an issue, when it might not be.

scripts/observability/install-observability.sh
```diff
@@ -229,6 +229,17 @@ else
     echo "   ⚠️  Metadata evaluator PrometheusRule not found - skipping alert"
 fi
 
+# Deploy OIDC authentication PrometheusRule (alerts on OIDC/JWT auth failures and latency)
+OIDC_RULE_FILE="$BASE_OBSERVABILITY_DIR/authorino-maas-oidc-prometheusrule.yaml"
+if ! kubectl get deploy -n "$AUTHORINO_NAMESPACE" authorino &>/dev/null 2>&1; then
```
EgorLu: Nitpick: `&>/dev/null 2>&1` can be shortened to `&>/dev/null`

scripts/observability/install-observability.sh
```diff
@@ -245,6 +256,12 @@ echo "   Istio:     istio_requests_total, istio_request_duration_milliseconds"
 echo "   vLLM:      vllm:num_requests_running, vllm:num_requests_waiting, vllm:kv_cache_usage_perc"
 echo ""
 
+echo "🚨 Alerting configured (if Authorino found):"
+echo "   MaaSAuthorinoMetadataEvaluatorHighFailureRate - maas-api metadata call failure rate > 10%"
```
EgorLu: Minor, but we are reporting hardcodded thresholds here. If someone changes the YAML these will be wrong.

While we can dynamically extract the values form the yaml with `yq`, we could also opt to print the description, but simply say something like "above set threshold".

WDYT?

deployment/base/observability/authorino-maas-oidc-prometheusrule.yaml
jland-redhat: I am not sure this is true at least on the alert itself. Maybe we should rename the file but the alert "MaaSAuthorinoAuthenticationHighFailureRate" which does not indicate OIDC and the description also does not indicate OIDC.

I think it is a fine add as is and we can rename in a followup PR

test/e2e/scripts/prow_run_smoke_test.sh
jland-redhat, line 93: We had to default OIDC to false so I am curious if these test are even running.

Fixed links so will validate. If not I don't think that blocks this PR but I do think we should have working OIDC test in some way in order to validate for GA

jland-redhat (dismissed):

test/e2e/tests/test_external_oidc.py
```diff
@@ -844,3 +896,229 @@ def test_injected_username_on_oidc_token_ignored(self, maas_api_base_url: str):
...
+    """Verify that the Authorino alerting infrastructure is present in the cluster.
+
+    The MaaSAuthorinoAuthenticationHighFailureRate PrometheusRule was introduced
+    as part of RHOAIENG-60847. If it is accidentally deleted or misconfigured,
+    the auth failure rate alerting silently stops working. This class provides
+    a lightweight smoke check that the rule exists.
+    """
+
+    def test_authorino_prometheusrule_exists(self):
+        """PrometheusRule maas-authorino-alerts is present in the MaaS subscription namespace."""
```
jrhyness: I think the name of the alert and the namespace are both wrong in this test.   It should be `authorino-maas-authentication-alerts` in the kuadrant namespace, right?

jrhyness (changes requested):

test/e2e/tests/test_external_oidc.py
ishitasequeira: Great catch!! I missed that test. Pushed the changes.

jrhyness (approved):
lgtm

PR #1081, opened 2026-07-02
feat: add tenant metrics

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-70043

## Description
This PR addresses AC1 and AC6 of the aforementioned story.
- Added `tenant_name` label to `maas_api_http_requests_total` and `maas_api_http_request_duration_seconds` metrics. The metrics middleware extracts tenant from `UserContext` in gin context after c.Next() completes. Unauthenticated routes (health, internal) use `cfg.TenantName` (e.g., "models-as-a-service") as the default rather than an empty string.
- Added `maas_api_key_validation_total` counter with `tenant_name` and `result` labels, recorded in `ValidateAPIKeyHandler` on every validation (valid/invalid).
- Added `maas_api_token_mint_total` counter with `tenant_name` and `result` labels, recorded in `CreateAPIKey` handler on success/failure.
- NewHandler now accepts an optional `MetricsRecorder` for the validation and mint counters (nil-safe for tests).

Additional changes to tests:
- metrics/prometheus_test.go: updated existing tests with tenant param; added TestRecordRequestDuration_TenantLabel, TestRecordRequestDuration_EmptyTenant, TestRecordKeyValidation, and TestRecordTokenMint
- metrics/recorder_test.go: updated mock recorder to match new interface (RecordKeyValidation, RecordTokenMint, tenant param on RecordRequestDuration)
- metrics/middleware_tenant_test.go (new): TestMiddleware_ExtractsTenantFromContext and TestMiddleware_DefaultTenantWhenNoUserContext
- metrics/server_test.go: updated expected metric format to include tenant_name="test-tenant" label
- api_keys/handler_test.go: updated all NewHandler calls to pass nil as the metrics recorder parameter

## How Has This Been Tested?
Locally

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added tenant-aware Prometheus metrics for HTTP request duration, API key validation, and token minting (with `tenant_name` and outcome/result labels).

* **Bug Fixes**
  * Request metrics now consistently use the configured default tenant label, regardless of user context.
  * API key/token events are recorded as `success`, `rejected`, or `failure` while preserving existing API responses.

* **Tests**
  * Updated and extended middleware, Prometheus recorder, and API key handler tests to verify `tenant_name` and result labels.


maas-api/internal/api_keys/handler.go
```diff
@@ -298,6 +316,14 @@ func (h *Handler) ValidateAPIKeyHandler(c *gin.Context) {
 		return
 	}
 
+	if h.metrics != nil {
```
jrhyness, line 331: Metrics are not recorded if ValidateAPIKey fails.

maas-api/internal/metrics/middleware.go
```diff
@@ -25,7 +27,14 @@ func NewMiddleware(recorder MetricsRecorder) gin.HandlerFunc {
 			route = "unmatched"
 		}
 
+		tenant := defaultTenant
+		if u, ok := c.Get("user"); ok {
```
jrhyness: Why do we need to set the tenant to anything but the one set for this maas-api?

maas-api/internal/api_keys/handler.go
yu-teo: You are right since cases when `ValidateAPIKey` returns an error, the handler returns early before reaching the metrics recording at line 319. Which would mean the validation failures due to internal errors are invisible to metrics.
Would you agree recording an "error" beofre returning on the error path? The tenant will be an unknown field when the lookup fails, so I'll use an empty string. Does that sound like a plan?

maas-api/internal/metrics/middleware.go
yu-teo: I missed the fact that each maas-api instance serves **exactly** one tenant. So, we can reply on `UserContext.Tenant` that will always match the configured tenant. Upon further look, I also see thatinternal routes (which lack a `UserContext`) will safely fall back to `defaultTenant`. I will simplify to just use `defaultTenant` directly. Thank you!

maas-api/internal/api_keys/handler.go
jrhyness: Don't we have the maas-api tenant ?

yu-teo: Looking into that. Give me some time to see if I can extract the tenant from anywhere else.

yu-teo: I think the cleanest way is to introduce a getter similar to how we already have `GetMaxExpirationDays()`. This is because our handler does not store tenant directly. And I do not like reaching to it through two layers by utilizing `service.config`. Pushing the changes in.

jrhyness (approved):
lgtm

PR #1086, opened 2026-07-03
feat: revoke api keys on tenant/subscription deletion

jrhyness:
When a tenant (AITenant) or subscription (MaaSSubscription) is deleted, all associated API keys are now automatically revoked via finalizers. This prevents orphaned keys from granting access if the tenant or subscription is recreated with the same name.

Changes:
- maas-api: Add internal endpoints for key revocation
  - POST /internal/v1/api-keys/revoke-for-tenant
  - POST /internal/v1/api-keys/revoke-for-subscription
  - Store layer methods for soft delete (status='revoked')
  - In-memory tracking (deletingTenants) to block new key creation during tenant deletion
- maas-controller: Call revocation endpoints in finalizers
  - AITenant finalizer calls revoke-for-tenant before deletion
  - MaaSSubscription finalizer calls revoke-for-subscription
  - New httpclient package with TLS support and exponential backoff retry
  - Nil-safe (skips if client not configured, for tests)
- Tests: Unit tests for all layers, E2E tests for both deletion scenarios

Security: Prevents reuse of orphaned keys ([RHOAIENG-72792](https://redhat.atlassian.net/browse/RHOAIENG-72792))

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added internal bulk API key revocation endpoints for tenant- and subscription-scoped workflows.
  * Added tenant/subscription finalizer support to revoke associated MaaS API keys.
  * Introduced an in-cluster HTTPS client with JSON handling and request retries.
* **Bug Fixes**
  * Prevents new API key issuance while a tenant is being revoked.
  * Prevents key reuse after tenant/subscription deletion by invalidating revoked keys.
* **Tests**
  * Added/extended unit, handler, HTTP client/retry, and E2E tests to cover bulk revocation and deletion scenarios.

[RHOAIENG-72792]: https://redhat.atlassian.net/browse/RHOAIENG-72792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployment/base/maas-api/core/deployment.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: CONTROLLER_NAMESPACE
+          value: "opendatahub"  # Replaced by controller at runtime based on its own namespace
```
dmytro-zaharnytskyi, line 29: This stays opendatahub but I don't see anything in the tenant render path that patches it. On RHOAI the controller runs in redhat-ods-applications, so revoke calls will get 403 because maas-api expects system:serviceaccount:opendatahub:maas-controller 🤔

maas-controller/pkg/httpclient/client.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, errors.New("failed to parse service CA certificate")
+	}
+
+	tlsConfig := &tls.Config{
+		RootCAs:    caCertPool,
+		MinVersion: tls.VersionTLS12,
+	}
+
+	// Load service account token for authentication with maas-api internal endpoints
+	token, err := os.ReadFile(ServiceAccountTokenPath)
```
dmytro-zaharnytskyi: Token is read once at client init. With projected/rotating SA tokens the file on disk gets refreshed but this cached copy won't - so seems like revoke calls can start failing 401 or 403 long after the controller starts

dmytro-zaharnytskyi (commented):
great pr, found 2 small issues

jrhyness:
No longer needed.  Tenant keys are revoked via https://github.com/opendatahub-io/models-as-a-service/pull/1112

PR #1089, opened 2026-07-06
feat: new base manifest entry point for AI Gateway

ryancham715:
## Description

As part of our move from ODH to AI Gateway, we should take back the control of image injection by generating the ConfigMap ourselves, instead of using MaaS-aware functions in the upper controller.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Added a new Kustomize base configuration for the controller component.
  * Pinned container image references for the controller, API, payload processing, and cleanup using environment-driven parameters.
  * Set the monitoring namespace to `opendatahub`.
  * Ensured stable generated ConfigMap naming and automatically wired the controller’s manager image from the generated configuration.

### DIFF
--- a/deployment/base/maas-controller/base/kustomization.yaml
+++ b/deployment/base/maas-controller/base/kustomization.yaml
@@ -1,0 +1,33 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../default
+
+labels:
+  - pairs:
+      app.kubernetes.io/name: maas-controller
+      app.kubernetes.io/component: models-as-a-service
+      app.kubernetes.io/part-of: models-as-a-service
+
+configMapGenerator:
+  # Keep a stable name because manager.yaml references this ConfigMap directly
+  # through configMapKeyRef.
+  - name: maas-parameters
+    envs:
+      - params.env
+
+generatorOptions:
+  disableNameSuffixHash: true
+
+replacements:
+  - source:
+      kind: ConfigMap
+      name: maas-parameters
+      fieldPath: data.maas-controller-image
+    targets:
+      - select:
+          kind: Deployment
+          name: maas-controller
+        fieldPaths:
+          - spec.template.spec.containers.[name=manager].image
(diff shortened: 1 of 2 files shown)


somya-bhatnagar (commented):
## Issue to Address

### 1. Duplicate Image Configuration
The images: stanza in manager/kustomization.yaml and the replacements: in base/kustomization.yaml do create two control paths. However, removing images: from manager/kustomization.yaml would break the ODH operator path,
which renders from default/ → manager/ (not through base/). The ODH operator relies on the images: stanza to set the image, then overrides it via ApplyParams with RELATED_IMAGE_* env vars.

  The two paths are intentional for different consumers:
  - manager/ images: → for ODH operator rendering via default/
  - base/ replacements: → for AGO rendering via base/ with ConfigMap

When AGO renders from base/, images: runs first, then replacements: overrides it — ConfigMap wins. So there's no
actual conflict at runtime, just visual duplication.

Can we add a comment for base/kustomization.yaml explaining why.

somya-bhatnagar (approved):
lgtm
add small comment if possible.

deployment/base/maas-controller/base/kustomization.yaml
ishitasequeira, line 1: Any reason to not re-use manifests in path `deployment/base/maas-controller/default` and require a new `base` package? Can we just add the `configMapGenerator + replacements` to `default/kustomization.yaml` and use the existing `params.env`, then let AGO override tags via an overlay if needed?

ryancham715: The reason is MaaS component handler (the old way to install MaaS) still exists in ODH, which generates another `maas-parameters` itself. I'm not sure if that will cause a conflict. I can give it a try though.

Edit: In my test, ODH will try to deploy two different `maas-parameters`, which leads to an infinite loop of reconciliation. I think it's better to have separate `/base` until we clean up the legacy installation path.

jland-redhat (approved):
lgtm

ishitasequeira (approved):
/lgtm

chaitanya1731 (approved):
lgtm

PR #1096, opened 2026-07-06
fix: disable stage-2 ext_proc on maas-api routes

### LINKED ISSUE #1095: GET /v1/models through the gateway returns a bare 500 and crashes the payload-processing pod
## Symptom

`GET /v1/models` through the MaaS gateway returns a bare `HTTP/2 500` with an empty body (no error JSON):

```
curl -i -H "x-api-key: <maas-key>" https://<gateway-host>/v1/models
→ HTTP/2 500, empty body
```

`POST /v1/messages` etc. through the same gateway work fine, so auth (Kuadrant) is not involved.

## Root cause (verified on a live cluster)

Two contributing bugs:

**1. The stage-2 ext_proc filter is not disabled on maas-api routes.**
`deployment/base/payload-processing/manager/envoy-filter.yaml` disables only `envoy.filters.http.ext_proc.ipp-pre` on `maas-api-route.0` (`/v1/models`) and `maas-api-route.1` (`/maas-api/*`). The stage-2 `envoy.filters.http.ext_proc.ipp` filter — which has `failure_mode_allow: false` — still processes these routes, even though they are answered by maas-api and are not inference traffic.

**2. The IPP framework (v0.1.0-rc.3) panics on bodyless requests.**
For a bodyless GET, Envoy sends request headers with `end_of_stream=true`; the framework returns immediately and `reqCtx.Profile` is never assigned (it is only set in `HandleRequestBody`). When the response body chunk arrives, the dispatch loop dereferences `reqCtx.Profile.NeedsResponseBuffering` → nil-pointer panic that kills the whole ext_proc process:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/llm-d/llm-d-inference-payload-processor/pkg/handlers.(*Server).Process(...)
    .../pkg/handlers/server.go:181
```

The pod accumulated 5 restarts in ~45 minutes from probing this endpoint alone. Because the process is shared, **every `GET /v1/models` also kills all in-flight inference streams** — the blast radius is much larger than the probe itself.

With the ext_proc stream gone and `failure_mode_allow: false`, Envoy aborts the request with the observed bare 500.

The framework panic is already fixed upstream (llm-d/llm-d-inference-payload-processor#207, requests now initialized with an empty profile, released in `v0.1.0-rc.4`), so the payload-processing image needs a framework bump — tracked separately.

## Why this matters

Clients capability-probe this endpoint. Claude Code calls `GET /v1/models/{id}` to learn the model's context window; when the probe fails it falls back to a 200k assumption. For 1M-context models that means the client reports ~200% context usage and triggers premature autocompaction at 20% of the real window — a silent quality degradation for every gateway user. OpenAI-compatible clients (Codex etc.) probe `GET /v1/models` the same way.

## Fix

Add `envoy.filters.http.ext_proc.ipp: disabled: true` to the `typed_per_filter_config` of both maas-api routes. Verified live: `GET /v1/models` returns 200 with the aggregated model list from maas-api, `401` without auth (Kuadrant unaffected), and the payload-processing pod no longer restarts.

## Follow-up gap (separate issue material)

`GET /v1/models/{id}` returns **404** — maas-api only implements the list endpoint. This is the endpoint Claude Code probes for context-window metadata, so returning per-model metadata there (e.g. from the ExternalModel CR) would complete the capability-probe story. 404 is a correct, non-crashing answer, but the context-window fallback described above still applies until this exists.

noyitz:
Fixes #1095

## What

Adds `envoy.filters.http.ext_proc.ipp: disabled: true` to the `typed_per_filter_config` of both maas-api routes (`/v1/models` and `/maas-api/*`) in the payload-processing EnvoyFilter. Previously only the stage-1 `ipp-pre` filter was disabled there.

## Why

`GET /v1/models` through the gateway returned a bare `HTTP/2 500` (empty body) and crashed the payload-processing pod. Root cause chain, verified on a live cluster:

1. The stage-2 `ipp` ext_proc filter (`failure_mode_allow: false`) still processed maas-api routes.
2. For a bodyless GET, the IPP framework never assigns a request profile (only set during body handling), and panics with a nil-pointer dereference when the response body arrives (`handlers/server.go:181` in framework `v0.1.0-rc.3`) — killing the whole ext_proc process and **all in-flight inference streams with it**. The pod accumulated 5 restarts in ~45 minutes from probing this endpoint.
3. With the ext_proc stream dead and `failure_mode_allow: false`, Envoy aborts with a bare 500.

The framework panic is fixed upstream (llm-d/llm-d-inference-payload-processor#207, in `v0.1.0-rc.4`) and the image bump is tracked separately, but this route-level exemption is correct regardless: maas-api traffic is not inference traffic and must not pass through the inference payload-processing chain.

**Client impact of the bug**: clients capability-probe this endpoint. Claude Code calls `GET /v1/models/{id}` to learn the model context window; on failure it assumes 200k. For 1M-context models that means ~200% reported context usage and premature autocompaction at 20% of the real window, silently degrading quality for every gateway user. OpenAI-compatible clients probe `GET /v1/models` the same way.

## Verification (live cluster)

Applied the same per-route config to a running gateway:

- `GET /v1/models` with a valid MaaS key → `200` with the aggregated model list served by maas-api (all providers)
- `GET /v1/models` without auth → `401` (Kuadrant auth unaffected by the ext_proc exemption)
- payload-processing pod: no new restarts; before the fix each probe crashed it
- `POST /v1/messages` inference traffic unaffected (still passes through both ext_proc stages)

## Risk analysis

**Risk rating**: 3

**Why**: Config-only change, but it touches the gateway EnvoyFilter, which affects all tenants' traffic on the shared gateway. The blast radius of a mistake here is high (it could disable payload processing on inference routes), mitigated by the change being additive, scoped to the two maas-api route names that already carry the identical `ipp-pre` exemption, and verified end-to-end on a live cluster. The Prow smoke path exercises the models endpoint and inference requests, which covers both the fixed path and the must-not-regress path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated gateway routing so the external processing step is disabled on the affected non-inference routes.
  * Improved consistency in request handling for those routes by preventing an extra per-route processor from running.


noyitz:
@coderabbitai review

liangwen12year (approved):
lgtm

jland-redhat (approved):
Lgtm

ishitasequeira (approved):
lgtm

chaitanya1731 (approved):
LGTM

jland-redhat (approved):
lgtm

ishitasequeira (approved):
/lgtm

PR #1108, opened 2026-07-08
docs: document dual auth modes — Dashboard path vs standalone OIDC client

liangwen12year:
## Summary
Documents both MaaS authentication paths side by side:

1. **RHOAI Dashboard path** — OpenShift OAuth → kube-auth-proxy → API key minting
2. **Standalone OIDC client** — External IdP (Keycloak/Entra ID) → JWT bearer → Authorino JWKS validation

Covers:
- How both modes coexist on the same gateway
- Which layer validates each token type (Authorino vs maas-api)
- Field alignment between Tenant CR and AITenant CR OIDC configuration
- IdP configuration requirements for standalone mode
- Explicit note that client secrets are not used (JWT bearer validation only)

## Test plan
- [ ] `mkdocs serve` renders the page without errors
- [ ] All internal links resolve correctly
- [ ] Token flow diagrams match `authentication-internals.md`


docs/content/concepts/auth-modes.md
```diff
@@ -0,0 +1,126 @@
...
+**What the admin configures:**
+
+```yaml
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: Tenant
+metadata:
+  name: default-tenant
+  namespace: models-as-a-service
+spec:
+  externalOIDC:
```
dmytro-zaharnytskyi: The default tenant is AITenant-managed, so sticking OIDC on Tenant.spec.externalOIDC here won't actually do anything, might be worth changing this section to lead with AiTenant.spec.oidc and call out that the Tenant CR path is only for unmanaged tenants

docs/content/concepts/auth-modes.md
```diff
@@ -0,0 +1,126 @@
...
+For Mode 2 (standalone OIDC), the IdP must:
+
+- Serve `/.well-known/openid-configuration` over HTTPS at the `issuerUrl`
+- Include `azp` (authorized party) claim in issued tokens matching the configured `clientId`
+- Expose a JWKS endpoint for public key retrieval
+- Issue tokens with group claims if group-based authorization is needed (claim name depends on IdP configuration)
+
+MaaS does **not** perform OAuth2 client authentication (no `client_secret`). It validates bearer tokens only. Client authentication and secret management are the IdP's responsibility.
+
+## Field Alignment Between Modes
```
dmytro-zaharnytskyi: small naming thing - this section is really about Tenant CR vs AITenant CR field mapping, not Mode 1 vs Mode 2, so something like "Field alignment (Tenant vs AITenant)" would be less confusing

dmytro-zaharnytskyi (commented):
left a few comments

chaitanya1731 (approved):
lgtm

PR #1051, opened 2026-06-25
feat: add infrastructure namespace separation support

jrhyness:
Enables optional separation of infrastructure services (maas-api, maas-db-config) from the controller namespace into a dedicated infrastructure namespace.

Background:
- Previous attempt (commit c8c084be, reverted June 11) blocked by ROSA webhook restrictions on namespace creation
- Desire to avoid opendatahub-operator changes (operator is migrating)
- Made optional due to still-unresolved ROSA webhook issue

Key features:
- Auto-derivation: INFRA_NAMESPACE=AUTO derives namespace from controller namespace (opendatahub → odh-ai-gateway-infra, redhat-ods-applications → redhat-ai-gateway-infra)
- Backward compatible: defaults to controller namespace (no separation)
- Automatic migration: when enabled, automatically cleans up legacy maas-api deployment from controller namespace
- PostgreSQL location-independent: database can be external (AWS RDS, etc.); only maas-api and connection secret move to infrastructure namespace

Implementation:
- Renamed all "split" references to "separation" for clarity
- Added --infra-namespace flag (replaces --maas-api-namespace)
- Added INFRA_NAMESPACE env var (replaces MAAS_API_NAMESPACE)
- Scripts support AUTO resolution matching Go logic
- deploy.sh automatically patches controller deployment when INFRA_NAMESPACE env var is set
- Component: deployment/components/infra-namespace-separation/

Activation:
- Via env var: export INFRA_NAMESPACE=AUTO; ./scripts/deploy.sh
- Via kustomize: add infra-namespace-separation component to overlay

Documentation:
- Added namespace architecture section to controller-architecture.md
- Added infra-namespace-migration.md with migration procedure
- TODO markers added for when separation becomes the default

## How Has This Been Tested?

Deployed using the INFRA_NAMESPACE=AUTO:

```
(opendatahub-tests) jrhyness@jrhyness-thinkpadp1gen7:~/src/models-as-a-service$ oc get pods -A | grep maas
keycloak-system                                    maas-keycloak-0                                                   1/1     Running     0              96m
odh-ai-gateway-infra                               maas-api-d5bc5fd7b-xf8hr                                          1/1     Running     0              96m
odh-ai-gateway-infra                               maas-api-jimmie-57bc67f7c8-lgbcx                                  1/1     Running     0              79m
opendatahub                                        maas-controller-7dc889b79-hq8kf                                   1/1     Running     0              96m
[...]

PR #1111, opened 2026-07-08
fix: migrate maas-db-config secret for infrastructure namespace separ…

jrhyness:
…ation

For: https://redhat.atlassian.net/browse/RHOAIENG-75591

Adds automatic migration of the maas-db-config secret from controller namespace to infrastructure namespace during upgrades when namespace separation is enabled. This fixes a backwards compatibility issue introduced in PR #1051.

Background:
- PR #1051 enabled infrastructure namespace separation with --infra-namespace defaulting to AUTO
- This derives namespace mappings (e.g., redhat-ods-applications → redhat-ai-gateway-infra)
- maas-api looks for maas-db-config in its pod's namespace (from NAMESPACE env var set via fieldRef)
- During upgrades, the secret remains in the old controller namespace
- Result: maas-api fails to start with "secret not found" error

Solution:
- Added migrateMaaSDBSecretToInfraNamespace() function that runs during controller startup after ensuring infrastructure namespace exists
- Detects upgrade scenarios vs fresh installs
- Copies secret from controller namespace to infrastructure namespace with FQDN connection strings for cross-namespace postgres access
- Idempotent: safe to run multiple times
- Coexists with setup-database.sh migration (for manual deployments)

Migration scenarios handled:
- Upgrade: secret in controller NS only → copy with FQDN
- Already migrated: secret exists in infra NS → no-op
- Fresh install: no secret anywhere → no-op
- Same namespace: no separation enabled → no-op
- External DB: secret already created → no-op

Testing:
- Added comprehensive unit tests for all migration scenarios
- Tests FQDN conversion logic
- Tests password masking for safe logging
- All tests pass

Risk analysis:
**Risk rating**: 2

**Why**:
- Low-risk defensive change: only runs when namespace separation enabled
- Idempotent and well-tested with unit tests covering all scenarios
- Mirrors existing migration logic in setup-database.sh (proven pattern)
- Fail-safe: exits with clear error if migration fails
- No changes to secret format or schema

**What could break**:
- Race condition if multiple controller replicas start simultaneously (mitigated: AlreadyExists treated as success)
- If RBAC doesn't allow secret creation in infra namespace (same RBAC as current namespace creation)

## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added automatic migration of the MaaS `maas-db-config` Secret from the controller namespace to the infrastructure namespace when namespaces are separated.
  - Updated connection URL handling to use the infrastructure DNS form and preserved existing migrated values.
  - Added namespace readiness checks during startup.
- **Documentation**
  - Refreshed setup/upgrade/development and restart instructions for infrastructure-namespace deployments.
- **RBAC / Configuration**
  - Added/updated RBAC and Kustomize/script wiring to support secret migration across controller and infrastructure namespaces.
  - Renamed the infrastructure namespace parameter to `infrastructure-namespace` (using `AUTO`).
- **Tests**
  - Added coverage for URL conversion, masking, and secret migration behavior.


deployment/base/maas-controller/rbac/kustomization.yaml
```diff
@@ -6,5 +6,6 @@ resources:
   - service_account.yaml
   - leader_election_role.yaml
   - leader_election_role_binding.yaml
+  - secret_migrate_clusterrole.yaml
```
dmytro-zaharnytskyi: I think we should add the operator-side binding here too. the Jira failure is from operator-managed nightly clusters, and this only adds the ClusterRole, so the controller may still have just the old get/list/watch secret access and fail the startup migration on Secrets().Create()

maas-controller/cmd/manager/main.go
```diff
@@ -241,6 +242,163 @@ func resolveNamespaceAfterTerminationWait(namespace string, finalNs *corev1.Name
...
+		}
+		return fmt.Errorf("failed to create secret in infrastructure namespace: %w", err)
+	}
+
+	log.Info("successfully migrated maas-db-config secret to infrastructure namespace",
+		"sourceNamespace", controllerNs,
+		"targetNamespace", infraNs,
+		"connectionURL", maskConnectionURL(fqdnURL))
+
+	// Delete the old secret from controller namespace (move, not copy)
```
dmytro-zaharnytskyi: Not entirely sure here, but I think we should make this copy-only for now. The migration is fixed once the infra namespace has maas-db-config, deleting the source secret during controller startup feels like extra cleanup before the new maas-api path is proven healthy, and it also forces delete RBAC in the controller namespace 🤔

dmytro-zaharnytskyi (commented):
🤗

jrhyness:
https://github.com/opendatahub-io/ai-gateway-operator/pull/52

deployment/base/maas-controller/base/kustomization.yaml
```diff
@@ -31,3 +31,28 @@ replacements:
...
+        fieldPaths:
+          - spec.template.spec.containers.[name=manager].env.[name=INFRA_NAMESPACE].value
+  - source:
+      kind: ConfigMap
+      name: maas-parameters
+      fieldPath: data.infrastructure-namespace
+    targets:
+      - select:
+          kind: Role
+          name: secret-migrate-role-infra
```
ishitasequeira, line 51: data.infrastructure-namespace in the generated ConfigMap is AUTO (from params.env). AUTO is a valid Go sentinel, main.go resolves it at runtime , but Kubernetes rejects it as a namespace name at apply time. This replacement will create `secret-migrate-role-infra` and `secret-migrate-rolebinding-infra` in a namespace literally named "AUTO", breaking the migration RBAC for every kustomize-based deployment path.

Also, the `INFRA_NAMESPACE` env var replacement on the Deployment is fine and should stay. Only the Role/RoleBinding namespace targets may need a fix.

scripts/deploy.sh
```diff
@@ -612,6 +616,47 @@ EOF
     }
     rm -rf "${controller_overlay_dir}"
 
+    # Create RoleBindings for secret migration in both controller and infra namespaces
+    log_info "  Creating secret migration RoleBindings..."
+    local infra_ns
+    infra_ns=$(derive_infra_namespace "$NAMESPACE")
+
+    # RoleBinding in controller namespace
+    kubectl apply -f - <<YAML || log_warn "Failed to create secret migration RoleBinding in controller namespace"
```
ishitasequeira: RBAC provisioning failures are swallowed as warnings here. The controller is then restarted a few lines down, hits Forbidden during migrateMaaSDBSecretToInfraNamespace, and calls os.Exit(1). From the operator's perspective this looks like a spontaneous controller CrashLoopBackOff with no visible link to the RBAC step.

Should we make these fatal?
```
kubectl apply -f - <<YAML || {log_error "Failed to create secret migration RoleBinding in controller namespace"
 return 1
}
```

scripts/deploy.sh
```diff
@@ -612,6 +616,47 @@ EOF
     }
     rm -rf "${controller_overlay_dir}"
 
+    # Create RoleBindings for secret migration in both controller and infra namespaces
+    log_info "  Creating secret migration RoleBindings..."
+    local infra_ns
+    infra_ns=$(derive_infra_namespace "$NAMESPACE")
```
ishitasequeira: the RBAC creation block derives the infra namespace from $NAMESPACE unconditionally:

`infra_ns=$(derive_infra_namespace "$NAMESPACE")`

But later (line ~672), the deployment patch section correctly respects an explicitly set INFRA_NAMESPACE:

```
if [ "${INFRA_NAMESPACE+x}" = "x" ]; then
  # patches deployment with $INFRA_NAMESPACE
fi
```
If a user runs `INFRA_NAMESPACE=custom-infra ./scripts/deploy.sh`, the controller is configured to migrate into custom-infra, but the RoleBinding gets applied to the auto-derived namespace (e.g. redhat-ai-gateway-infra). Migration fails with Forbidden at startup.

we could  use the same resolution logic as the INFRA_NAMESPACE patch block:
```
local infra_ns
if [ "${INFRA_NAMESPACE+x}" = "x" ]; then
  infra_ns="$INFRA_NAMESPACE"
else
  infra_ns=$(derive_infra_namespace "$NAMESPACE")
fi
```

maas-controller/cmd/manager/main.go
```diff
@@ -639,6 +797,15 @@ func main() {
...
 			os.Exit(1)
 		}
+
+		// Migrate maas-db-config secret from controller namespace to infrastructure namespace
+		// during upgrades (maintains backward compatibility for operator-managed deployments)
+		if err := migrateMaaSDBSecretToInfraNamespace(context.Background(), controllerNamespace, infraNamespace, clientset); err != nil {
+			setupLog.Error(err, "failed to migrate maas-db-config secret to infrastructure namespace",
+				"controllerNamespace", controllerNamespace,
+				"infraNamespace", infraNamespace)
+			os.Exit(1)
```
ishitasequeira: A Forbidden error from the API server (e.g. the RBAC hasn't been applied yet, or the operator companion PR isn't deployed) causes the controller to refuse to start permanently. There's no retry, no self-healing, and no way to recover without a redeployment.

This codebase already handles this situation gracefully elsewhere — `ensureManagedNamespaceWithClient` treats `Forbidden` as "assume operator-managed, continue". Should the migration follow the same pattern:

```
if err := migrateMaaSDBSecretToInfraNamespace(...); err != nil {
    if errors.IsForbidden(err) {
        setupLog.Info("insufficient RBAC to migrate maas-db-config secret — "+
            "ensure secret-migrate RoleBindings are applied; skipping migration",
            "controllerNamespace", controllerNamespace,
            "infraNamespace", infraNamespace)
    } else {
        setupLog.Error(err, "failed to migrate maas-db-config secret",
            "controllerNamespace", controllerNamespace,
            "infraNamespace", infraNamespace)
        os.Exit(1)
    }
}
```
This way a transient RBAC gap causes a warning, not a permanent crash, and the next controller restart (after RBAC is applied) will complete the migration.

scripts/deploy.sh
```diff
@@ -612,6 +616,47 @@ EOF
...
+    kubectl apply -f - <<YAML || log_warn "Failed to create secret migration RoleBinding in controller namespace"
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: maas-controller-secret-migrate-controller
+  namespace: ${NAMESPACE}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: maas-controller-secret-migrate-role
```
ishitasequeira: I could not see a ClusterRole with the name `maas-controller-secret-migrate-role` anywhere in this repository. I only see ClusterRoles named `secret-migrate-role-controller` and `secret-migrate-role-infra` in the kustomize RBAC manifests.

We could also remove the entire Creating secret migration RoleBindings block from deploy.sh. The correct bindings are already being applied by `kustomize build .` a few lines above.

jrhyness:
Closing in favour of https://github.com/opendatahub-io/models-as-a-service/pull/1149

PR #1117, opened 2026-07-08
fix: fail startup on missing gateway host and reject non-HTTPS probe URLs (FIND-010)

liangwen12year:
## Summary
Two SSRF mitigations for model discovery probes:

1. **Fail startup hard** when `gatewayInternalHost` is empty. Previously, maas-api logged a warning and continued with access checks disabled — probes would connect directly to model URLs bypassing the gateway's auth stack.

2. **Reject non-HTTPS model URLs** before probing. Models with `http://` or other schemes are skipped with a warning log. This prevents probes from being sent to arbitrary HTTP endpoints.

## Finding
**FIND-010** — Model Discovery Probe SSRF (CWE-918, CVSS 3.7 Low)

Remediation:
- ~~Fail startup hard when `gatewayInternalHost == ""`~~ — Done (`main.go`)
- ~~Reject `model.URL` whose `Scheme != "https"` before probing~~ — Done (`discovery.go`)

## Test plan
- [ ] maas-api fails to start when gateway service is not found (no silent degradation)
- [ ] Models with `http://` URLs are excluded from `/v1/models` access probes
- [ ] Models with `https://` URLs continue to work normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup now fails fast if the gateway host needed for model access checks cannot be resolved, instead of continuing with checks disabled.
  * Model discovery now skips endpoints that use unsupported URL schemes, improving safety and reliability.
* **Tests**
  * Added coverage for URL scheme validation to confirm only safe HTTP-based model endpoints are checked.

### DIFF
--- a/maas-api/internal/models/discovery.go
+++ b/maas-api/internal/models/discovery.go
@@ -206,6 +206,11 @@ func (m *Manager) FilterModelsByAccess(ctx context.Context, models []Model, auth
 			m.logger.Debug("FilterModelsByAccess: skipping model with no URL", "id", model.ID)
 			continue
 		}
+		if model.URL.Scheme != "https" && model.URL.Scheme != "http" {
+			m.logger.Warn("FilterModelsByAccess: rejecting model with unsupported URL scheme",
+				"id", model.ID, "scheme", model.URL.Scheme)
+			continue
+		}
 		modelsEndpoint, err := url.JoinPath(model.URL.String(), "v1", "models")
 		if err != nil {
 			m.logger.Debug("FilterModelsByAccess: failed to build endpoint", "id", model.ID, "error", err)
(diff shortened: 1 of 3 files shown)


maas-api/internal/models/discovery.go
```diff
@@ -206,6 +206,11 @@ func (m *Manager) FilterModelsByAccess(ctx context.Context, models []Model, auth
 			m.logger.Debug("FilterModelsByAccess: skipping model with no URL", "id", model.ID)
 			continue
 		}
+		if model.URL.Scheme != "https" && model.URL.Scheme != "http" {
+			m.logger.Warn("FilterModelsByAccess: rejecting model with unsupported URL scheme",
+				"id", model.ID, "scheme", model.URL.Scheme)
+			continue
+		}
```
ishitasequeira, line 213: Can we add some unit tests for this?

liangwen12year: Added `TestFilterModelsByAccess_URLSchemeValidation` in 60ed89e4 — covers http (allowed), file/ftp/empty (rejected), and nil URL (skipped).

jrhyness (approved):
lgtm

minor: @liangwen12year , the PR title suggests only https should be allowed but http is as well.

PR #1122, opened 2026-07-09
fix: scope Secret informer cache to infrastructure namespace

somya-bhatnagar:
## Summary

Restricts the Secret informer cache in maas-controller to the **infrastructure namespace** (where `maas-api` and `maas-db-config` live) instead of watching Secrets cluster-wide.

## Why / benefit

Without this change, `Watches(&corev1.Secret{})` causes the manager to run a **cluster-wide** Secret LIST/WATCH. The controller then caches Secrets from every namespace in memory, even though the Tenant controller predicate only reacts to `maas-db-config`.

Scoping the informer to the infra namespace:

- Reduces API-server load (no cluster-wide Secret watch)
- Shrinks memory / runtime data surface (other namespaces' Secrets never enter the controller process)
- Aligns the cache with where `maas-db-config` actually lives

Note: ClusterRole Secret RBAC is unchanged (still required for SSA escalation of `payload-processing-reader`). This PR narrows the **informer cache**, not permissions.

## Change

```go
// main.go — both standard and multi-tenant cache configs
infraNsCfg := map[string]cache.Config{infraNamespace: {}}
&corev1.Secret{}: {Namespaces: infraNsCfg},
```

The Tenant controller already filters Secret events to `maas-db-config` via predicate. `AppNamespace` is set from `infraNamespace`, and `checkDatabaseSecret` reads that Secret from the same namespace, so the cache scope matches existing read/watch paths.

## Risk analysis

**Risk rating:** 2

**Why:** Small cache-config change with no reconciler logic changes, but an incorrect namespace would silently break Secret watching (caught once in review when scoped to the subscription namespace). Current scope matches `infraNamespace` / `AppNamespace` / `maas-db-config` location.

## Related
https://redhat.atlassian.net/browse/RHOAIENG-76578

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted infrastructure Secret caching to the designated infrastructure namespace.
  * Improved namespace isolation while preserving broader access for other platform resources.


somya-bhatnagar:
@coderabbitai review

somya-bhatnagar:
## Integration Testing — PASS ✅

**Environment:**
- Fresh OpenShift 4.21 cluster (ROSA/EC2)
- ODH operator: PR #3723 branch (`aigateway.modelsAsAService` architecture)
- maas-controller: combined image with PR #1075 (dynamic CRD watches) + PR #1122 (Secret scoping)

**Test — Full integration deployment:**
1. ODH operator → ai-gateway-operator → maas-controller (all deployed via RELATED_IMAGE injection)
2. maas-controller with PR #1122 — Secret informer scoped to `odh-ai-gateway-infra`, `models-as-a-service`, `opendatahub`
3. `maas-api` Running (1/1) — successfully read `maas-db-config` from `odh-ai-gateway-infra` namespace
4. `AITenant/models-as-a-service` Ready=True — full reconcile loop completed
5. No secret access errors in controller logs

**Relates to Jira: RHOAIENG-66858**

jrhyness (changes requested):
`maas-db-config` is in the infrastructure namespace not the subscription namespace.   This would break the watcher.

maas-controller/cmd/manager/main.go
```diff
@@ -649,6 +649,9 @@ func main() {
 			&maasv1alpha1.Tenant{}:           {},
 			&maasv1alpha1.MaaSAuthPolicy{}:   {Namespaces: nsCfg},
 			&maasv1alpha1.MaaSSubscription{}: {Namespaces: nsCfg},
+			// Restrict the Secret informer to the subscription namespace so the cluster-wide
+			// secrets LIST/WATCH is not required — namespace-scoped RBAC is sufficient.
+			&corev1.Secret{}: {Namespaces: nsCfg},
```
jland-redhat: As mentioned before by @jrhyness I think this should be the infra namespace

somya-bhatnagar: Yes, agree. It will be for infra ns.

somya-bhatnagar: Fixed. The Secret informer is now scoped to `infraNamespace` (where `maas-db-config` lives) in both the standard and multi-tenant cache configurations. A separate `infraNsCfg` is used for `&corev1.Secret{}` while MaaS CRs (MaaSAuthPolicy, MaaSSubscription) continue to use `nsCfg` (subscription namespace).

maas-controller/cmd/manager/main.go
```diff
@@ -627,6 +629,9 @@ func main() {
 			&maasv1alpha1.Tenant{}:           {},
 			&maasv1alpha1.MaaSAuthPolicy{}:   {Namespaces: nsCfg},
 			&maasv1alpha1.MaaSSubscription{}: {Namespaces: nsCfg},
+			// Restrict the Secret informer to the infrastructure namespace (where maas-db-config lives)
+			// so the cluster-wide secrets LIST/WATCH is not required — namespace-scoped RBAC is sufficient.
```
ishitasequeira: Is this really true? The comment surely applies to `maas-db-config` secret, but as MaaS is still installing IPP, RBAC is still required to be cluster wide for IPP Secret informer.

somya-bhatnagar: Good catch — the comment was misleading. This PR only scopes the **informer cache** to the infra namespace; it does not change the ClusterRole (secrets `list`/`watch` remains cluster-wide in RBAC). Fixed the comment to accurately say "to avoid caching cluster-wide Secrets" and removed the incorrect claim about RBAC sufficiency.

ishitasequeira (approved):
lgtm

jrhyness (approved):
lgtm

ishitasequeira (approved):
lgtm

PR #1125, opened 2026-07-09
fix: ignore X-MaaS-Subscription header for non-API-key requests (FIND-009)

liangwen12year:
## Summary
For OpenShift token callers (non `sk-oai-` auth), Authorino does not set the `X-MaaS-Subscription` header. Any client-supplied value passes through untrusted, allowing callers to:
- Probe whether a named subscription exists vs access-denied (different error body text)
- Bypass the "return all accessible models" intent for OC-token callers

## Fix
Move `X-MaaS-Subscription` header extraction inside the `isAPIKeyRequest` branch. Non-API-key requests always get `requestedSubscription = ""`, which triggers the "return all accessible models" path.

## Finding
**FIND-009** — Unauthorized Subscription Header Passthrough (CWE-639, CVSS 4.3 Medium)

## Why not the other remediation options?
- **Auth policy approach** (set empty header for non-sk-oai-): Requires controller changes to the generated AuthPolicy, more complex, and the policy is auto-generated
- **HTTPRoute RequestHeaderModifier**: Would strip the header for ALL requests including API keys that need it
- **Application-level fix** (this PR): Simplest, no gateway/policy changes needed, enforced at the point of use

## Test plan
- [ ] OpenShift token request with `X-MaaS-Subscription: some-sub` header returns all accessible models (header ignored)
- [ ] API key request with `X-MaaS-Subscription: some-sub` continues to filter by that subscription (no behavior change)
- [ ] API key request without subscription header returns 403 (existing fail-closed behavior preserved)

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized subscription authorization failures when listing models, returning a consistent forbidden response and message.
  * Improved request validation to consistently reject API key requests without an associated subscription.
  * Prevented subscription existence from being inferred through differing error responses.

### DIFF
--- a/maas-api/internal/handlers/models.go
+++ b/maas-api/internal/handlers/models.go
@@ -185,17 +185,24 @@ func (h *ModelsHandler) extractAndValidateAuth(c *gin.Context) (string, string, 
 		return "", "", false, errors.New("missing authorization")
 	}
 
-	// Extract x-maas-subscription header.
+	isAPIKeyRequest := strings.HasPrefix(authHeader, "Bearer sk-oai-")
+
+	// Extract x-maas-subscription header ONLY for API key requests.
+	// For OpenShift token callers, Authorino does not set this header, so any
+	// client-supplied value would be untrusted. Ignoring it prevents callers
+	// from probing subscription existence or bypassing the "return all
+	// accessible models" intent. (CWE-639 / FIND-009)
 	requestedSubscription := ""
-	headerValues := c.Request.Header.Values("X-Maas-Subscription")
-	for i := len(headerValues) - 1; i >= 0; i-- {
-		trimmed := strings.TrimSpace(headerValues[i])
-		if trimmed != "" {
-			requestedSubscription = trimmed
-			break
+	if isAPIKeyRequest {
+		headerValues := c.Request.Header.Values("X-Maas-Subscription")
+		for i := len(headerValues) - 1; i >= 0; i-- {
+			trimmed := strings.TrimSpace(headerValues[i])
+			if trimmed != "" {
+				requestedSubscription = trimmed
+				break
+			}
 		}
 	}
-	isAPIKeyRequest := strings.HasPrefix(authHeader, "Bearer sk-oai-")
 
 	// Fail closed: API keys without a bound subscription must be rejected
 	if isAPIKeyRequest && requestedSubscription == "" {
--- a/maas-api/internal/handlers/models.go
+++ b/maas-api/internal/handlers/models.go
@@ -122,21 +122,13 @@ func (h *ModelsHandler) handleSubscriptionSelectionError(c *gin.Context, err err
 		return
 	}
 
-	if errors.As(err, &accessDeniedErr) {
-		h.logger.Debug("Access denied to subscription")
+	// Unify "access denied" and "not found" responses to prevent callers from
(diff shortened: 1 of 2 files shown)


maas-api/internal/handlers/models.go
```diff
@@ -185,17 +185,24 @@ func (h *ModelsHandler) extractAndValidateAuth(c *gin.Context) (string, string,
...
+		headerValues := c.Request.Header.Values("X-Maas-Subscription")
+		for i := len(headerValues) - 1; i >= 0; i-- {
+			trimmed := strings.TrimSpace(headerValues[i])
+			if trimmed != "" {
+				requestedSubscription = trimmed
+				break
+			}
 		}
 	}
-	isAPIKeyRequest := strings.HasPrefix(authHeader, "Bearer sk-oai-")
```
ishitasequeira, line 198: A few questions came up while reviewing that I wanted to raise before we go with this approach.

Does this remove a documented feature for user-token and OIDC callers?

Several sources describe X-MaaS-Subscription as an intentional opt-in filter for non-API-key users:                                                                         
  - README: "With a user token, optional X-MaaS-Subscription filters to one subscription when you have access to several"
  - OpenAPI spec: "With X-MaaS-Subscription header: Returns models from the specified subscription only"
  - MultipleSubscriptionsError tells users to send this header
  - E2E test test_user_token_with_subscription_header_filters validates this behavior

Would removing this break the expected workflow for users with multiple subscriptions?

The actual information leak is that handleSubscriptionSelectionError returns different message text for "not found" vs "access denied" — both are 403, but the message body differs. We could unify those:
```                                                                                                                                                                              
  if errors.As(err, &accessDeniedErr) || errors.As(err, &notFoundErr) {                                                                                                     
      c.JSON(http.StatusForbidden, gin.H{
          "error": gin.H{                                                                                                                                                     
              "message": "access denied to requested subscription",                                                                                                           
              "type":    "permission_error",                                                                                                                                  
          }})                                                                                                                                                                 
      return                                                                                                                                                                  
  } 
```

liangwen12year (commented):
@ishitasequeira, patch updated.

maas-api/internal/handlers/models.go
liangwen12year: Good catch — you're right that ignoring the header entirely breaks the documented opt-in filter for user-token callers. Updated in 0cf26330 with your suggested approach:

1. **Restored** `X-MaaS-Subscription` extraction for all auth types (preserves README, OpenAPI, and E2E behavior)
2. **Unified** `accessDeniedErr` and `notFoundErr` responses — both now return the same `"access denied to requested subscription"` message with 403, so callers can't probe subscription existence

Tests updated to verify both "unknown" and "inaccessible" subscriptions return identical 403 responses.

jrhyness: @liangwen12year , I don't think that commit made it to the PR.

ishitasequeira: +1, @liangwen12year not seeing the changes on the PR. Did you missing pushing the change?

liangwen12year: I push it now. Thanks for reminding me. Can you re-review the PR?

ishitasequeira (approved):
/lgtm

PR #1129, opened 2026-07-09
fix: exempt /v1/subscriptions and /v1/api-keys from gateway deny-all policies

ishitasequeira:
### Problem
Requests to `/v1/subscriptions` and `/v1/api-keys` through the gateway fail with 429 or 500, making these management endpoints unusable in operator-managed deployments.                                  

Two independent issues cause this:
1. **TRLP deny-all blocks management paths** — `gateway-default-deny` exempts only `/maas-api/*` and `/v1/models` from its zero-token rate limit. Requests to `/v1/subscriptions` and `/v1/api-keys` get 0 tokens → 429 Too Many Requests.                                             

2. **ext_proc fails on unhandled routes** — The payload-processing EnvoyFilter disables ext_proc only on HTTPRoute rules 0 (`/v1/models`) and 1 (`/v1/subscriptions`), but not rules 2 (`/v1/api-keys`) and 3 (`/maas-api/*`). When the payload-processing pod is absent, `failure_mode_allow: false` causes those routes to return 500.                                 

Both issues affect any auth method (OpenShift tokens, API keys, OIDC) since the policies operate at the gateway level before authentication is evaluated.

## Changes                                                     
- Widen `gateway-default-deny` TRLP predicate from `/v1/models` to `/v1/` so all management endpoints are exempt from the zero-token deny-all rate limit                                                   
- Add ext_proc disable entries for HTTPRoute rules 2 and 3 in the payload-processing EnvoyFilter template
- Update `patchPayloadProcessingEnvoyFilter` loop and test assertions to cover all four route rules

## How Has This Been Tested?

Manual testing on the cluster

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Expanded routing and request-processing coverage for additional MAAS API paths.
  * Added clearer routing rules so more management endpoints are handled consistently.

* **Bug Fixes**
  * Updated default-deny behavior so key management endpoints are no longer unintentionally blocked.
  * Disabled extra request-processing on more non-inference routes to reduce unwanted interference.

* **Documentation**
  * Added clearer inline guidance about route ordering and internal-only endpoints.

### DIFF
--- a/deployment/base/maas-controller/policies/gateway-default-deny.yaml
+++ b/deployment/base/maas-controller/policies/gateway-default-deny.yaml
@@ -2,8 +2,9 @@
 # Scoped to model inference paths only — does NOT affect non-model routes
 # like /maas-api/ or /v1/models (the MaaS API management endpoints).
 #
-# The when predicate below excludes /maas-api and /v1/models so this deny
-# applies only to model inference paths (e.g. /llm/..., /production/...).
+# The when predicate below excludes /maas-api and all /v1/* management
+# endpoints so this deny applies only to model inference paths (e.g.
+# /llm/..., /production/...).
 #
 # The PRIMARY default deny is gateway-default-auth.yaml (AuthPolicy), which
 # returns 401/403 for unconfigured models. This TRLP is a defense-in-depth layer:
@@ -33,7 +34,7 @@ spec:
     limits:
       deny-all-by-default:
         when:
-          - predicate: '!request.path.startsWith("/maas-api") && !request.path.startsWith("/v1/models")'
+          - predicate: '!request.path.startsWith("/maas-api") && !request.path.startsWith("/v1/")'
         rates:
           - limit: 0
             window: 1m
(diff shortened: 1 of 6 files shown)


deployment/base/maas-controller/policies/gateway-default-deny.yaml
```diff
@@ -33,7 +34,7 @@ spec:
     limits:
       deny-all-by-default:
         when:
-          - predicate: '!request.path.startsWith("/maas-api") && !request.path.startsWith("/v1/models")'
+          - predicate: '!request.path.startsWith("/maas-api") && !request.path.startsWith("/v1/")'
```
jland-redhat: Does this not also include `/v1/chat/completions` endpoint?

ishitasequeira: good catch, fixed it 👍

jland-redhat (approved):
lgtm

chaitanya1731 (approved):
lgtm

PR #1133, opened 2026-07-09
fix: hash/redact username

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-69257

## Description
This PR addresses the aforementioned CWE with a hashing of sensitive value like `username`. Worth noting that hashing was selected over masking because we still prefer to have some sort of a visible identifier in logs in order to be able to trace the logs to a particular user. On that note, the HMAC key is **per-process** (regenerated on every restart), so the hash for the same username will differ across pod **restarts**.

## How Has This Been Tested?
Updated unit-tests pass successfully.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive usernames are now redacted in operational logs across API key, subscription, and token-related workflows.
  * Success, warning, and error messages now use consistent masked values, reducing the chance of exposing private account details in logs.
  * Added a reusable redaction format for sensitive values so masked entries are stable and easier to correlate without revealing raw data.


jland-redhat (approved):
lgtm

PR #1138, opened 2026-07-10
fix: add container hardening — readOnlyRootFilesystem, seccompProfile (FIND-007)

liangwen12year:
## Summary

Add missing container hardening fields across all deployment manifests:

| Workload | readOnlyRootFilesystem | seccompProfile: RuntimeDefault |
|---|---|---|
| **maas-controller** | **added** | **added** |
| **maas-api** | already set | **added** |
| **cleanup CronJob** | already set | **added** |
| **payload-processing** | already set | **added** |

**Finding:** FIND-007 — Missing Container Hardening (CWE-1188)  
**CVSS:** 4.0 (AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:L/A:L)

### Notes

- maas-controller's only writable path (`/tmp/k8s-webhook-server/serving-certs`) is already a `secret` volume mount — `readOnlyRootFilesystem: true` is safe
- OpenShift `restricted-v2` SCC injects seccomp in most installs, but explicit declaration satisfies CIS 5.7.2 and prevents drift on permissive SCCs
- No code changes — manifest-only

## Test plan

- [x] `kustomize build` succeeds for all bases (controller, api, payload-processing)
- [ ] Cluster validation: pods start and remain healthy with hardened security context

### DIFF
--- a/deployment/base/maas-api/core/cronjob-cleanup.yaml
+++ b/deployment/base/maas-api/core/cronjob-cleanup.yaml
@@ -43,3 +43,5 @@ spec:
                 - ALL
               readOnlyRootFilesystem: true
               runAsNonRoot: true
+              seccompProfile:
+                type: RuntimeDefault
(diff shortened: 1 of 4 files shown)


chaitanya1731 (approved):
lgtm

PR #1139, opened 2026-07-10
fix: deprecated flag forces insecure mode

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-69242

## Description
`handleDeprecatedFlags()` in `maas-api/internal/config/config.go:237-246` unconditionally sets `c.Secure = false` whenever the deprecated PORT env var is set. This creates a dangerous scenario:

 Operator sets `SECURE=true` + `PORT=8443` (leftover) without TLS certs:
 1. `handleDeprecatedFlags()` silently clears `Secure` to `false`
 2. TLS is not configured, so `TLS.Enabled()` doesn't re-enable it
 3. The Secure && !TLS.Enabled() guard doesn't catch (Secure is already false)
 4. Server starts plaintext HTTP on :8443 — a port that conventionally implies HTTPS

The proposed **solution** is to have `handleDeprecatedFlags()` now returning an error if the deprecated `PORT/--port` flag is used alongside `SECURE=true` or any TLS configuration (TLS_CERT/TLS_KEY/TLS_SELF_SIGNED). This prevents the silent downgrade to plaintext that the CWE describes. Backward compatibility is preserved — PORT still works when neither `Secure` nor `TLS` is configured.

## How Has This Been Tested?
Unit tests updated and successfully ran.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conflicting deprecated port and secure/TLS settings from silently downgrading connections to plaintext.
  * Configuration validation now reports errors for incompatible options instead of continuing with side effects.
  * Deprecated port settings now set the listen address only when none is already configured.

### DIFF
--- a/maas-api/internal/config/config.go
+++ b/maas-api/internal/config/config.go
@@ -235,14 +235,22 @@ func (c *Config) Validate() error {
 }
 
 // handleDeprecatedFlags maps deprecated flags to new configuration.
-func (c *Config) handleDeprecatedFlags() {
-	// If deprecated --port flag is used, map to new model (HTTP mode)
-	if c.deprecatedHTTPPort != "" {
-		c.Secure = false
-		if c.Address == "" {
-			c.Address = ":" + c.deprecatedHTTPPort
-		}
+// Returns an error if the deprecated --port flag conflicts with TLS settings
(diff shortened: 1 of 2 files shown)


yu-teo:
@coderabbitai review

chaitanya1731 (approved):
lgtm

PR #1140, opened 2026-07-10
ci: make govulncheck non-blocking until go-toolset is updated

chaitanya1731:
## Summary

Makes govulncheck CI jobs non-blocking and adds self-trigger paths so workflow changes are validated.

## Description

- Set `continue-on-error: true` on govulncheck jobs in both `maas-api-ci.yml` and `maas-controller-ci.yml` so stdlib findings are reported but do not block PR merges.
- Add each workflow file to its own `paths` trigger so changes to CI configuration are validated by the affected pipeline.
- No go.mod, dependency, or application code changes.

**Why non-blocking:** Go stdlib CVEs (`crypto/tls`, `crypto/x509`, `mime`, `net/textproto`, `html/template`, `net`) require Go >= 1.25.12, but the Red Hat `ubi9/go-toolset:1.25` builder image currently ships Go 1.25.9. These failures are pre-existing on `main` and cannot be fixed until Red Hat updates the image. govulncheck will be re-enforced once the builder image catches up.

## Risk analysis

**Risk rating**: 0

**Why**: CI-only change — adds `continue-on-error: true` to two workflow jobs and self-trigger paths. No runtime, build, or application impact.

## How it was tested

- This PR itself triggers both workflows (via the self-trigger paths) and validates that govulncheck runs but does not block the overall check status.
- No go.mod or application changes to validate.

### DIFF
--- a/.github/workflows/maas-api-ci.yml
+++ b/.github/workflows/maas-api-ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - 'maas-api/**'
+      - '.github/workflows/maas-api-ci.yml'
 
 permissions:
   actions: read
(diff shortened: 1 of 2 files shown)


chaitanya1731:
@coderabbitai review

dmytro-zaharnytskyi (dismissed):
lgtm

somya-bhatnagar (dismissed):

jrhyness (dismissed):
lgtm

ishitasequeira (dismissed):

somya-bhatnagar (approved):
lgtm

dmytro-zaharnytskyi (approved):
lgtm

PR #1159, opened 2026-07-13
feat: maas-controller self teardown

ryancham715:
https://redhat.atlassian.net/browse/RHOAIENG-69775

## Description

**Major**

Since maas-controller has a self deployment process, it should have a self teardown process too, instead of relying on upper operator in this matter, causing strong coupling. This PR implements a workflow for maas-controller to clean up everything it generates or reconciles. Specifically,
1. an external party (ai-gateway) annotates maas-controller with `teardown-requested` to start the teardown process.
2. maas-controller will stop ensuring namespace/ai-tenants, default tenant and Config/default.
3. maas-controller deletes MaaS CRs (including AITenants, whose finalizer will delete all tenant-specific resources), Config/default, and namespace/ai-tenants.
4. after self-teardown is done, maas-controller annotates itself with `teardown-completed` to inform the external party.

**Minor**
1. Remove legacy Tenant CRs' finalizer on migration, to present AITenant removal process from being stuck.
2. Stop Config/default from owning maas-controller. maas-controller should control the lifecycle of Config/default, not the way around.

## How Has This Been Tested?

Deployed MaaS on live cluster, manually annotated `teardown-requested`, observed expected cleanup behavior and `teardown-completed` after it's done. If annotation `teardown-requested` is taken down, maas-controller's self-heal will bring back everything.

## Merge criteria:

## Summary by CodeRabbit

- **New Features**
  - Added self-teardown support for the MAAS controller, including teardown requested/completed signals.
  - Self-teardown now cleans up remaining managed tenant resources and deletes the generated namespace when eligible.
  - Prevents AITenant namespace maintenance from recreating resources while teardown is requested.

- **Bug Fixes**
  - Improved finalizer handling and cleanup sequencing so legacy and tenant resources are removed reliably.
  - Updated controller RBAC for safer permissions (reduced delete capability).
  - Enhanced lifecycle E2E coverage by re-enabling previously skipped tests.

### DIFF
--- a/maas-controller/pkg/controller/maas/aitenant_controller.go
+++ b/maas-controller/pkg/controller/maas/aitenant_controller.go
@@ -422,6 +422,7 @@ func (r *AITenantReconciler) markLegacyTenantDeprecated(ctx context.Context, ten
 	annotations["maas.opendatahub.io/deprecated-by"] = maasv1alpha1.MaasTenantConfigKind
 	annotations["maas.opendatahub.io/migrated-to"] = maasv1alpha1.MaasTenantConfigInstanceName
 	legacy.SetAnnotations(annotations)
+	controllerutil.RemoveFinalizer(&legacy, tenantFinalizer)
 	if equality.Semantic.DeepEqual(base, &legacy) {
 		return nil
 	}
--- a/maas-controller/pkg/controller/maas/aitenant_controller.go
+++ b/maas-controller/pkg/controller/maas/aitenant_controller.go
@@ -116,16 +116,14 @@ func (r *AITenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return r.reconcileAITenantDelete(ctx, &aitenant)
 	}
 
-	// Unblocking UI
-	// TODO: Include adding the finalizer back as part of https://github.com/opendatahub-io/models-as-a-service/pull/1159
-	// if !controllerutil.ContainsFinalizer(&aitenant, aitenantFinalizer) {
-	// 	base := aitenant.DeepCopy()
-	// 	controllerutil.AddFinalizer(&aitenant, aitenantFinalizer)
-	// 	if err := r.Patch(ctx, &aitenant, client.MergeFrom(base)); err != nil {
-	// 		return ctrl.Result{}, err
-	// 	}
-	// 	return ctrl.Result{RequeueAfter: time.Second}, nil
-	// }
+	if !controllerutil.ContainsFinalizer(&aitenant, aitenantFinalizer) {
+		base := aitenant.DeepCopy()
+		controllerutil.AddFinalizer(&aitenant, aitenantFinalizer)
+		if err := r.Patch(ctx, &aitenant, client.MergeFrom(base)); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
 
 	statusSnapshot := aitenant.Status.DeepCopy()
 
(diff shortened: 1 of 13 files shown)


maas-controller/pkg/controller/maas/self_teardown.go
```diff
@@ -0,0 +1,220 @@
...
+// use this as the single source of truth for "is teardown in progress".
+func TeardownRequestedOnDeployment(dep *appsv1.Deployment) bool {
+	if dep == nil {
+		return false
+	}
+	return dep.GetAnnotations()[TeardownRequestedAnnotation] == "true"
+}
+
+var configCleanupResourceTypes = []schema.GroupVersionKind{
+	{Group: "maas.opendatahub.io", Version: "v1alpha1", Kind: "AITenant"},
```
jland-redhat, line 52: For this PR lets just remove AITenant and nothing else.

I want to keep this as simple as possible

jland-redhat (changes requested):

maas-controller/pkg/controller/maas/self_teardown.go
```diff
@@ -0,0 +1,220 @@
...
+
+	if cfg != nil {
+		if err := r.Delete(ctx, cfg); client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, fmt.Errorf("delete Config/default during teardown: %w", err)
+		}
+	}
+
+	if err := r.markTeardownCompleted(ctx, dep); err != nil {
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
ryancham715: Not applying — `Config` never carries a finalizer here (see comment above `handleRequestedTeardown`: "a plain, unguarded step (no finalizer)"), so `Delete` removes it synchronously with no terminating window. Its dependents (ServiceMonitor, EnvoyFilter, Perses dashboards) are finalizer-free too and cascade-delete via GC independent of this controller. Resources with finalizers (AITenant, MaasTenantConfig, etc.) are owned by AITenant, not Config, and are already deleted earlier in `cleanupTeardownResources`. Covered by `TestLifecycleReconciler_TeardownRequestedDeletesConfigAndMarksCompleted`.

maas-controller/pkg/controller/maas/self_teardown.go
ryancham715: Addressed. AI Gateway might have to change accordingly, as CRDs can't be deleted any more.

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -422,6 +422,7 @@ func (r *AITenantReconciler) markLegacyTenantDeprecated(ctx context.Context, ten
 	annotations["maas.opendatahub.io/deprecated-by"] = maasv1alpha1.MaasTenantConfigKind
 	annotations["maas.opendatahub.io/migrated-to"] = maasv1alpha1.MaasTenantConfigInstanceName
 	legacy.SetAnnotations(annotations)
+	controllerutil.RemoveFinalizer(&legacy, tenantFinalizer)
```
ishitasequeira, line 423: We mentioned that we need to add the finalizer addition logic back as part of this PR. Can we address that? https://github.com/ryancham715/models-as-a-service/blob/2654493b03348ba875be7db1958ddeb0847eb3bc/maas-controller/pkg/controller/maas/aitenant_controller.go#L119-L128

jrhyness (changes requested):
The failing test, test_maas_controller_deployment_lists_config_owner_reference, needs to be adjusted for the new approach.

jrhyness (approved):
lgtm

ishitasequeira (approved):
/lgtm

jland-redhat (approved):

PR #1160, opened 2026-07-13
feat: add body-based routing (BBR) support for LLMInferenceService

ishitasequeira:
## Summary

- Add `resolvedModelAlias` status field to MaaSModelRef, populated by the controller from the backend handler. For LLMISvc this is the publisher ID (`publishers/{ns}/models/{name}`), for ExternalModel it's the `targetModel` from the spec.
- Update Authorino auth policy generation to pass the raw `x-gateway-model-name` header as the model identity (CEL + Rego), and add alternate keys from `resolvedModelAlias` so auth matches both path-based and BBR identities.
- Add reverse-lookup in maas-api subscription selector so BBR requests (keyed by publisher ID) resolve to the correct subscription.
- Switch ipp-post `request_body_mode` to `BUFFERED` in the EnvoyFilter so the ext_proc plugin can rewrite the request body before it reaches the backend.

https://redhat.atlassian.net/browse/RHOAIENG-78214
## Motivation

BBR sends `POST /v1/chat/completions` with the model in the request body (no path prefix). The model name in the body uses publisher ID format (`publishers/llm/models/facebook/opt-125m`), which didn't match the canonical identity used by auth, subscription lookup, or the backend. This change ensures all three layers handle the BBR identity correctly.

## Multi-tenancy

Publisher ID includes the namespace (`publishers/{namespace}/models/{name}`), so models across different tenant namespaces produce distinct identities at every layer.

## Test plan

- [x] BBR request returns 200 with correct model response
- [x] Path-based request still returns 200 (no regression)
- [x] Auth passes for BBR identity (dual-key model_access)
- [x] Subscription resolves correctly for BBR identity
- [x] Unit tests pass for auth policy controller and model ref controller

## Test evidence

**BBR (body-based routing) — 200 OK:**
```
curl -X POST "https://maas.apps.isequeir-test-1.eh5f.s1.devshift.org/v1/chat/completions" \
  -H "Authorization: Bearer sk-oai-..." \
  -H "Content-Type: application/json" \
  -d '{"model": "publishers/llm/models/facebook/opt-125m", "messages": [{"role": "user", "content": "hello"}]}'

HTTP/1.1 200 OK
{"id":"...","object":"chat.completion","model":"facebook/opt-125m","choices":[...]}
```

**Path-based routing — 200 OK (no regression):**
```
curl -X POST "https://maas.apps.isequeir-test-1.eh5f.s1.devshift.org/llm/facebook-opt-125m-simulated/v1/chat/completions" \
  -H "Authorization: Bearer sk-oai-..." \
  -H "Content-Type: application/json" \
  -d '{"model": "facebook/opt-125m", "messages": [{"role": "user", "content": "hello"}]}'

HTTP/1.1 200 OK
{"id":"...","object":"chat.completion","model":"facebook/opt-125m","choices":[...]}
```

## Merge criteria:

### DIFF
--- a/deployment/base/payload-processing/manager/envoy-filter.yaml
+++ b/deployment/base/payload-processing/manager/envoy-filter.yaml
@@ -37,8 +37,6 @@ spec:
               envoy_grpc:
                 cluster_name: outbound|9004||payload-pre-processing.openshift-ingress.svc.cluster.local
     # Stage 2: must run AFTER the WasmPlugin (needs auth to have completed).
-    # request_body_mode BUFFERED: Envoy holds the full request body until ext_proc responds,
-    # allowing model-provider-resolver to rewrite the body model field before vLLM receives it.
     - applyTo: HTTP_FILTER
       match:
         context: GATEWAY
@@ -59,7 +57,7 @@ spec:
             processing_mode:
               request_header_mode: "SEND"
               response_header_mode: "SEND"
-              request_body_mode: "BUFFERED"
+              request_body_mode: "FULL_DUPLEX_STREAMED"
               response_body_mode: "FULL_DUPLEX_STREAMED"
               request_trailer_mode: "SEND"
               response_trailer_mode: "SEND"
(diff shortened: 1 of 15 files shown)


deployment/base/payload-processing/manager/envoy-filter.yaml
```diff
@@ -57,7 +59,7 @@ spec:
             processing_mode:
               request_header_mode: "SEND"
               response_header_mode: "SEND"
-              request_body_mode: "FULL_DUPLEX_STREAMED"
+              request_body_mode: "BUFFERED"
```
jland-redhat: I don't believe we require buffered anymore

jland-redhat (approved):
lgtm

PR #1193, opened 2026-07-15
feat: surface canonical BBR model ID in GET /v1/models and MaaSModelRef status

israel-hdez:
## Description

Implements [RHOAIENG-76236](https://issues.redhat.com/browse/RHOAIENG-76236): expose the canonical BBR model identifier (`publishers/{namespace}/models/{name}`) as the authoritative model ID in `GET /v1/models` responses and `MaaSModelRef.status`.

**maas-controller** — KServe API `v1alpha1` → `v1alpha2` + dependency bump:
- Bump `github.com/kserve/kserve` to upstream HEAD (`v0.19.0-rc0`, commit `85065ab`) which includes `ModelSourcedAddressStatus` in `SourcedAddress`.
- `ResolveModelAlias` now reads `LLMInferenceService.status.addresses[*].models[0].name` (KServe's authoritative canonical ID) instead of recomputing it from spec/metadata.
- All imports and typed references updated from `v1alpha1` to `v1alpha2`; the dynamic CRD watch GVK updated to `serving.kserve.io/v1alpha2`.

**maas-api** (`GET /v1/models`):
- Model `id`: reads `status.resolvedModelAlias` (`publishers/{ns}/models/{name}`) for LLMInferenceService-backed models instead of `metadata.name`.
- Model `url`: derived from `status.httpRouteHostnames[0]` (shared gateway base URL) instead of per-model path URLs.
- Backend probing removed for `llmisvc` models — they are included directly when Ready, matching the ExternalModel pattern. This also fixes the empty list returned on BBR clusters ([RHOAIENG-76220](https://issues.redhat.com/browse/RHOAIENG-76220)).

**golangci-lint** bumped `v2.6.2` → `v2.12.2` to fix a panic on go1.26 system toolchains caused by `//go:build go1.26` files in transitive dependencies.

## How Has This Been Tested?

- `make -C maas-controller lint test` passes.
- `make -C maas-api lint test` passes.
- Manual verification on a BBR cluster is required to confirm end-to-end routing with the canonical model ID.

## Merge criteria:

[RHOAIENG-76236]: https://redhat.atlassian.net/browse/RHOAIENG-76236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHOAIENG-76220]: https://redhat.atlassian.net/browse/RHOAIENG-76220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Improved model discovery by filtering available models based on service readiness (instead of active probing/access validation).
* **Improvements**
  * Resolved MaaS model IDs and URLs from inference service status.
  * Updated controller and integrations to KServe **LLMInferenceService v1alpha2**.
  * Refined endpoint selection to prefer **model-routing** addresses when available.
  * Added a safety check to skip usage-log rendering/apply when the manifest path isn’t configured.
* **Maintenance**
  * Upgraded key dependencies and tooling.
* **Tests**
  * Updated model-listing and endpoint-selection expectations for readiness/scoping and v1alpha2 behavior.
  * Added lint suppressions for test-only credential-like strings.


maas-api/internal/models/maasmodelref.go
```diff
@@ -94,10 +110,27 @@ func maasModelRefToModel(u *unstructured.Unstructured) *Model {
...
+			if parsed, err := url.Parse(endpoint); err == nil {
+				urlPtr = (*apis.URL)(parsed)
+			}
+		}
+	default:
+		// LLMInferenceService-backed models on BBR clusters share the gateway base URL.
+		// Derive it from status.httpRouteHostnames[0] so all models point at the same
+		// gateway entry-point instead of per-model path URLs.
+		if hostnames, _, _ := unstructured.NestedStringSlice(u.Object, "status", "httpRouteHostnames"); len(hostnames) > 0 {
+			if parsed, err := url.Parse("https://" + hostnames[0]); err == nil {
```
israel-hdez, line 132: Is there a way to check if HTTPS is enabled?

@ishitasequeira @jland-redhat

maas-api/internal/models/discovery.go
```diff
@@ -202,6 +202,27 @@ func (m *Manager) FilterModelsByAccess(ctx context.Context, models []Model, auth
...
+		// Probing the base URL at /v1/models does not yield per-model access information
+		// (the gateway routes by model field in the request body, not by URL path).
+		// Include directly if Ready; gateway authpolicy enforces access at inference time.
+		// Subscription pre-filtering (upstream in ListLLMs) scopes the listing to models
+		// the user's subscriptions cover, providing the primary access boundary.
+		kind := model.Kind
+		if kind == "" {
+			kind = "llmisvc"
+		}
+		if kind == "llmisvc" {
```
ishitasequeira: I think there might be a kind mismatch in maas-api/internal/models/discovery.go:215 that would prevent the probe bypass from firing for BBR models. The CRD enum enforces LLMInferenceService (not llmisvc), so model.Kind will always be "LLMInferenceService" for BBR models. But the check on line 215 compares against "llmisvc", which means these models still fall through to the backend probe path — the one that doesn't work for BBR's body-based routing.

```suggestion
        if kind == "LLMInferenceService" || kind == "llmisvc" {
```

Could you double-check? I might be missing something about how the kind flows through, but from tracing `maasModelRefToModel → FilterModelsByAccess`, it looks like the bypass never triggers for current CRs.

maas-controller/go.mod
```diff
@@ -123,20 +77,18 @@ require (
...
-	sigs.k8s.io/gateway-api-inference-extension v1.3.1 // indirect
+	sigs.k8s.io/gateway-api-inference-extension v1.5.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 )
+
+// kserve requires controller-runtime v0.23.3 but pins via replace to v0.22.5 for
+// KEDA/webhook API compatibility; mirror that pin here so maas-controller compiles.
+replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.22.5
```
ishitasequeira, line 94: Could we keep the version of controller-runtime to be tagged at `v0.22.5` and avoid the replace?

maas-controller/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.25.8
 
 require (
 	github.com/go-logr/logr v1.4.3
-	github.com/kserve/kserve v0.19.0
+	github.com/kserve/kserve v0.19.0-rc0.0.20260715171408-85065ab905e4
```
ishitasequeira, line 7: Looks like we are tagging on the RC release. Do we have an estimate of if the change will be getting added to a more stable release any time soon?

maas-api/internal/models/discovery.go
israel-hdez: It should be fixed now.

maas-controller/go.mod
israel-hdez: I can rollback, but I think we anyway will need the replacement. It affects the dependencies of the imported modules which is the main purpose here: to pin the kserve one.

maas-controller/go.mod
israel-hdez: I don't know the an exact date. But it should be in about 1-2 months.

maas-api/internal/models/maasmodelref.go
bartoszmajsak: I think hard assumption about protocol might be risky, even if 90% of use cases are going through HTTPS.

maas-controller/go.mod
bartoszmajsak: FWIW we only need APIs, so perhaps this is good enough?

maas-controller/pkg/controller/maas/providers_llmisvc.go
```diff
@@ -290,17 +285,24 @@ func (h *llmisvcHandler) selectAddress(llmisvc *kservev1alpha1.LLMInferenceServi
...
-	if llmisvc.Spec.Model.Name != nil && *llmisvc.Spec.Model.Name != "" {
-		modelName = *llmisvc.Spec.Model.Name
+	for _, addr := range llmisvc.Status.Addresses {
+		if len(addr.Models) > 0 && addr.Models[0].Name != "" {
+			return addr.Models[0].Name
+		}
 	}
-	return fmt.Sprintf("publishers/%s/models/%s", model.Namespace, modelName)
+	return ""
 }
```
bartoszmajsak, line 312: `ResolveModelAlias` returns `""` when KServe hasn't reconciled status yet. The caller unconditionally writes this to `status.resolvedModelAlias`, clearing a previously-populated value. `maasauthpolicy_controller` then skips BBR model header injection - fail open window might be a bit problematic here.

Maybe guard the write:
```go
if alias := handler.ResolveModelAlias(...); alias != "" {
    model.Status.ResolvedModelAlias = alias
}
```

maas-controller/pkg/controller/maas/providers_llmisvc.go
```diff
@@ -290,17 +285,24 @@ func (h *llmisvcHandler) selectAddress(llmisvc *kservev1alpha1.LLMInferenceServi
...
 
+// ResolveModelAlias returns the canonical BBR model ID for the referenced LLMInferenceService.
+//
+// It reads from LLMInferenceService.status.addresses[*].models[0].name, which KServe populates
+// as the authoritative canonical ID in the format publishers/{namespace}/models/{model-name}.
+// KServe is the source of truth for this value; MaaS reads and mirrors it.
+// Returns an empty string when the LLMISVC cannot be fetched or has no populated addresses.
 func (h *llmisvcHandler) ResolveModelAlias(ctx context.Context, log logr.Logger, model *maasv1alpha1.MaaSModelRef) string {
-	llmisvc := &kservev1alpha1.LLMInferenceService{}
```
bartoszmajsak, line 294: Identity resolution changed from local construction to reading `status.addresses[*].models[0].name`. This value drives auth policy generation - maybe we need more coverage for: populated models, empty addresses, empty model name, Get failure.

maas-controller/cmd/manager/migrate_db_secret_test.go
```diff
@@ -17,31 +17,31 @@ func TestConvertToFQDNConnectionURL(t *testing.T) {
...
 		},
-		{
+		{ //nolint:gosec // test data only, not real credentials
 			name:      "hostname with query params",
 			url:       "postgresql://user:pass@postgres:5432/db?sslmode=require",
 			namespace: "opendatahub",
 			want:      "postgresql://user:pass@postgres.opendatahub.svc.cluster.local:5432/db?sslmode=require",
 		},
-		{
+		{ //nolint:gosec // test data only, not real credentials
```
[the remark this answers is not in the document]
israel-hdez: I'm keeping as is. This is enough for me.

maas-controller/pkg/controller/maas/providers_llmisvc.go
israel-hdez: Hopefully, fixed.

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -528,6 +528,11 @@ func (r *LifecycleReconciler) ensureUsageDashboard(ctx context.Context, log logr
 // ensureUsageLogs deploys or removes OTel collector and RBAC for usage logging based on
 // the Config's usageLogging feature gate.
 func (r *LifecycleReconciler) ensureUsageLogs(ctx context.Context, log logr.Logger) error {
+	if r.UsageLogsManifestPath == "" {
+		log.Info("WARNING: Usage logs manifest path not configured; skipping usage logs")
+		return nil
+	}
+
```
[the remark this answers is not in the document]
israel-hdez: This is the same pattern as in other places like `ensureUsageDashboard`. Keeping as is.

maas-controller/pkg/controller/maas/providers_llmisvc.go
bartoszmajsak: Can be a follow-up

ishitasequeira (dismissed):
lgtm

ishitasequeira:
Closing in favor of https://github.com/opendatahub-io/models-as-a-service/pull/1208 - it includes test fixes and token rate limit fixes

PR #1214, opened 2026-07-20
fix: use OC bearer token for model discovery and inference in verify …

tgitelman:
…script

The verify-models-and-limits.sh script used API key ($TOKEN) for model discovery, inference, and rate limit testing. This fails because:
- /maas-api/v1/models only accepts OC bearer tokens (returns 401 for API keys)
- API key is bound to a single subscription, causing 403 on models outside that subscription

Switch to $OC_TOKEN for these calls so the gateway uses the user's full identity to select the appropriate subscription per model.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated verification requests to authenticate using the OpenShift identity token instead of the temporary MaaS API key, improving model discovery, chat completion, and rate-limit checks.
  * Preserved automatic cleanup of temporary MaaS API keys after verification runs.


jland-redhat (approved):
lgtm

PR #1215, opened 2026-07-20
fix: skip OpenShift-only watchers on xKS to prevent cache-sync timeout

tsisodia10:
## Summary

On xKS (vanilla Kubernetes), the `TenantReconciler` unconditionally watches `DSCInitialization` and `config.openshift.io/Authentication` — both OpenShift-only CRDs that don't exist on xKS. This causes a cache-sync timeout after ~2 minutes and the controller shuts down.

**Fix:** Use `crdExists()` + `registerWatchWhenCRDAppears()` — the same pattern already used for LLMInferenceService, AuthPolicy, and TokenRateLimitPolicy watchers in this codebase. No env var dependency needed.

## Problem

```
error: "failed to wait for maastenantconfig caches to sync"
```

The controller-runtime manager tries to sync all registered informers. The DSCInitialization watcher can't sync because the CRD doesn't exist → 2-minute timeout → `os.Exit(1)`.

## Changes

- `tenant_controller.go`:
  - **Authentication:** Static watch when CRD exists at boot; no dynamic fallback (CRD ships with OpenShift — present at install or never)
  - **DSCInitialization:** Static watch when CRD exists; dynamic `registerWatchWhenCRDAppears` fallback with `ResourceVersionChangedPredicate` for parity with static path

## Context

Part of xKS MaaS enablement. E2E validated on AKS — maas-controller stays running without cache-sync timeout.

Related PRs:
- [ai-gateway-operator#65](https://github.com/opendatahub-io/ai-gateway-operator/pull/65) — gateway-namespace fix (merged)
- [ai-gateway-operator#69](https://github.com/opendatahub-io/ai-gateway-operator/pull/69) — RBAC escalation fix (merged)

## Test plan

- [x] `go build ./...` compiles
- [x] `go test ./pkg/controller/maas/ -short` passes
- [x] AKS E2E: maas-controller starts clean, no cache-sync timeout (Jul 22 run with custom image containing this fix)

### DIFF
--- a/maas-controller/pkg/controller/maas/tenant_controller.go
+++ b/maas-controller/pkg/controller/maas/tenant_controller.go
@@ -290,26 +290,8 @@ func (r *TenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
-	if !authExists {
-		if err := registerWatchWhenCRDAppears(c, mgr, authCRD, func() source.Source {
-			auth := &unstructured.Unstructured{}
-			auth.SetGroupVersionKind(schema.GroupVersionKind{
-				Group: "config.openshift.io", Version: "v1", Kind: "Authentication",
-			})
-			return source.Kind(mgr.GetCache(), auth,
-				handler.TypedEnqueueRequestsFromMapFunc[*unstructured.Unstructured](
-					func(ctx context.Context, obj *unstructured.Unstructured) []reconcile.Request {
-						if obj.GetName() != openshiftAuthenticationClusterName {
-							return nil
-						}
-						return r.enqueueDefaultTenant(ctx, obj)
-					},
-				),
-			)
-		}); err != nil {
-			return fmt.Errorf("failed to register CRD watcher for Authentication: %w", err)
-		}
-	}
+	// No dynamic watch for Authentication CRD: it ships with OpenShift itself,
+	// so it is either present at boot (OCP) or will never appear (xKS).
 
 	if !dsciExists {
 		if err := registerWatchWhenCRDAppears(c, mgr, dsciCRD, func() source.Source {
@@ -323,6 +305,7 @@ func (r *TenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
 						return r.enqueueDefaultTenant(ctx, obj)
 					},
 				),
+				predicate.TypedResourceVersionChangedPredicate[*unstructured.Unstructured]{},
 			)
 		}); err != nil {
 			return fmt.Errorf("failed to register CRD watcher for DSCInitialization: %w", err)
--- a/maas-controller/pkg/controller/maas/tenant_controller.go
+++ b/maas-controller/pkg/controller/maas/tenant_controller.go
@@ -19,6 +19,7 @@ package maas
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
@@ -35,6 +36,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	maasv1alpha1 "github.com/opendatahub-io/models-as-a-service/maas-controller/api/maas/v1alpha1"
 	"github.com/opendatahub-io/models-as-a-service/maas-controller/pkg/platform/tenantreconcile"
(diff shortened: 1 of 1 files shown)


maas-controller/pkg/controller/maas/tenant_controller.go
```diff
@@ -257,16 +247,35 @@ func (r *TenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			Kind:    "DSCInitialization",
+		})
+
+		b = b.WatchesMetadata(
 			authMeta,
 			handler.EnqueueRequestsFromMapFunc(r.enqueueDefaultTenant),
 			builder.WithPredicates(authenticationClusterSingleton()),
-		).
-		Watches(
+		).Watches(
```
zdtsw: why even on non-xks, DSCI need to be watched? is it about the monitoring part

tsisodia10: @zdtsw Updated -- switched from platform-type check to CRD existence check using crdExists() + registerWatchWhenCRDAppears(), same pattern already used for LLMInferenceService/AuthPolicy/TokenRateLimitPolicy watchers in this codebase. No platform-type dependency needed. When CRDs are absent at startup, the static watch is skipped and a dynamic watcher activates if the CRD appears later.

maas-controller/pkg/controller/maas/tenant_controller.go
```diff
@@ -257,16 +247,87 @@ func (r *TenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+						return r.enqueueDefaultTenant(ctx, obj)
+					},
+				),
+			)
+		}); err != nil {
+			return fmt.Errorf("failed to register CRD watcher for DSCInitialization: %w", err)
+		}
+	}
+
+	return nil
```
zdtsw: try to understand this logic
so first 2 CRD does not exists, the it does static,
but this block of code is, later one if these 2CRD appear, i can image if someone install DSCI CRD to a cluster(probbly a xKS one) but the Auth one is a pure OCP thing, either it is there when OCP is installed or never be there.
so not sure in real, if both !exist() are needed? 
maybe I missed some case that we need a dynamic watch for these?

somya-bhatnagar:
@tsisodia10 Quick question on the test plan — have you been able to confirm on AKS/xKS that `maas-controller` stays up without the `failed to wait for maastenantconfig caches to sync` timeout after this change?

somya-bhatnagar (commented):
Thanks for switching to the `crdExists` + `registerWatchWhenCRDAppears` pattern — that matches the existing Kuadrant/KServe watchers and is the right fix for the cache-sync timeout on clusters without these CRDs.

A few notes (none are merge blockers on their own):

1. **PR description is stale** — it still describes the earlier `ODH_MODULE_OPERATOR_PLATFORM_TYPE` / `IsXKS` approach + `main.go` changes. The final diff only touches `tenant_controller.go`; worth updating so reviewers aren’t chasing removed code.

2. **Dynamic-path parity (nit)** — static DSCI watch uses `ResourceVersionChangedPredicate`; the dynamic fallback doesn’t. Same for Auth (`WatchesMetadata` vs full `unstructured`). Enqueue-only so not a correctness bug, but easy to align.

3. **Dynamic Auth registration** — agree with @zdtsw that `config.openshift.io/Authentication` is effectively present-at-install or never. Keeping the dynamic path is fine for pattern consistency / low cost; DSCI is the more realistic “appears later” case. Either keep both or drop Auth’s dynamic path with a short comment.

Holding `/lgtm` until we hear back on AKS/xKS startup confirmation.

maas-controller/pkg/controller/maas/tenant_controller.go
tsisodia10: @zdtsw @somya-bhatnagar removed the dynamic watch for it entirely. Kept DSCI's dynamic watch for the edge case where ODH/RHOAI is installed after maas-controller starts, consistent with the existing pattern for LLMInferenceService/AuthPolicy/TokenRateLimitPolicy watchers. Let me know what you think.

tsisodia10:
@somya-bhatnagar Thanks for the review — all three addressed in the latest push:

1. **PR description** — updated to reflect the current `crdExists()` approach (no more `ODH_MODULE_OPERATOR_PLATFORM_TYPE` / `main.go` references).

2. **Dynamic-path parity** — added `TypedResourceVersionChangedPredicate[*unstructured.Unstructured]{}` to the DSCI dynamic source. Matches the static branch now.

3. **Auth dynamic watch** — removed entirely with a short comment. Agreed with you and @zdtsw that it's present-at-install-or-never on OCP, so the dynamic path is dead code.

zdtsw (commented):
i am good with the current code change,  good to update PR description to algin the final solution
if anything else need to address e.g test etc, i leave it to @somya-bhatnagar for the final approval

ishitasequeira (commented):
Changes look good. @tsisodia10, will wait for confirmation on successful test run on xKS.

tsisodia10:
@zdtsw @ishitasequeira can someone add /ok-to-test

zdtsw:
not sure mine counts, not owner of this git repo :hehe:

somya-bhatnagar (approved):
Re-reviewed the latest push — looks good.

- Auth: static-only when CRD present; no dynamic fallback (correct for OCP-or-never)
- DSCI: static + `registerWatchWhenCRDAppears` with `TypedResourceVersionChangedPredicate` parity
- Matches existing Kuadrant/KServe optional-CRD pattern
- PR description / AKS validation updated

/lgtm

tsisodia10:
@somya-bhatnagar yes, validated on a clean AKS cluster (Jul 22 E2E run). maas-controller starts clean, no failed to wait for maastenantconfig caches to sync timeout. Full E2E passed through to model inference.

zdtsw (approved):

chaitanya1731 (approved):
lgtm

PR #1146, opened 2026-07-10
fix: rhoaieng 76228 payload processing rhcl

jland-redhat:
## Summary

Fixes [RHOAIENG-76228](https://redhat.atlassian.net/browse/RHOAIENG-76228): body-based inference was broken on **RHCL 1.4** because the payload-processing `EnvoyFilter` only anchored ext_proc filters on the Kuadrant `WasmPlugin` sub-filter.

On RHCL 1.4, auth is injected via `envoy.filters.http.wasm` (no `WasmPlugin` CR), so the existing patches never matched and `payload-pre-processing` / `payload-processing` were never wired into the gateway filter chain.

This PR adds a second anchor pair for RHCL 1.4 and updates tenant reconcile patching so both deployment paths work:

| Platform | Auth anchor |
|----------|-------------|
| ODH / community Kuadrant | `extensions.istio.io/wasmplugin/...` |
| RHCL 1.4 | `envoy.filters.http.wasm` |

Only one pair matches per cluster, so there is no double-insertion.

## Changes

- `deployment/base/payload-processing/manager/envoy-filter.yaml` — add RHCL 1.4 ext_proc patches (8 configPatches total: 4 filter inserts + 4 route-disable MERGEs)
- `maas-controller/pkg/platform/tenantreconcile/params.go` — patch all 4 filter inserts at reconcile time
- `maas-controller/pkg/platform/tenantreconcile/params_test.go` — update expectations for dual-anchor layout

## Test plan

- [x] `go test ./pkg/platform/tenantreconcile/...`
- [x] `./scripts/ci/validate-manifests.sh`
- [x] On RHCL 1.4: confirm `ext_proc` appears in gateway config dump
- [x] On RHCL 1.4: `POST /v1/<endpoint>` with model in JSON body returns 200 (not 404)
- [ ] On ODH/Kuadrant: confirm path-based and body-based inference still work (WasmPlugin anchor unchanged)

## Risk analysis

| | |
|---|---|
| **Risk rating** | 4 |
| **Why** | This changes gateway filter-chain wiring for payload processing. The default Prow smoke path may not cover RHCL 1.4 specifically, and a bad anchor could leave ext_proc unapplied (404 on body-routed inference) or mis-ordered relative to auth. ODH/Kuadrant behavior should be unchanged since the original WasmPlugin patches are untouched, but gateway ext_proc wiring is inherently high blast radius. |

---

[...]

PR #1239, opened 2026-07-22
fix: apply payload-processing envoyfilter after kuadrant

jland-redhat:
## Summary
- Set `EnvoyFilter/payload-processing` `spec.priority: 10` so RHCL 1.4 Kuadrant `envoy.filters.http.wasm` exists before our `INSERT_BEFORE`/`INSERT_AFTER` patches run (follow-up to #1146 / [RHOAIENG-76228](https://redhat.atlassian.net/browse/RHOAIENG-76228)).
- Without this, istiod applies our EF at default priority 0 before Kuadrant; wasm anchors miss; `ipp-pre` never enters the gateway chain; body-routed `/v1/*` returns `404 NR`.
- Surface misconfiguration via Tenant `DeploymentsAvailable` (priority/targetRefs check), e2e priority assertion, and `scripts/check-payload-ext-proc-filters.sh` (live config_dump order: `ipp-pre → wasm → ipp → router`).

## Test plan
- [x] `go test ./pkg/platform/tenantreconcile/...`
- [x] `./scripts/ci/validate-manifests.sh`
- [x] On RHCL 1.4 cluster: `./scripts/check-payload-ext-proc-filters.sh` passes with both ext_proc filters present
- [x] On RHCL 1.4: `POST /v1/chat/completions` with `model` in JSON body returns 200 (not 404 NR)
- [ ] On ODH/community Kuadrant: path- and body-based inference still work (WasmPlugin anchors unchanged; priority 10 is after default)

## Risk analysis
- **Risk rating**: 4
- **Why**: Gateway HTTP filter-chain wiring for payload-pre-processing and payload-processing. Wrong priority or ordering can leave body routing broken (404 NR) or mis-order ext_proc relative to auth. Default Prow smoke may not exercise RHCL 1.4 specifically; ODH WasmPlugin path should be unchanged but still depends on EnvoyFilter apply order.

[RHOAIENG-76228]: https://redhat.atlassian.net/browse/RHOAIENG-76228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Set the payload-processing gateway EnvoyFilter to the required priority (>= 10) and ensured it targets the correct gateway, reducing missed filter-chain behavior.
  * Updated platform readiness to also verify the payload-processing EnvoyFilter exists with an adequate priority and correct target.
  * Improved tenant readiness messaging to reflect payload-processing availability.

* **Tests**
  * Strengthened unit and e2e checks to validate payload-processing EnvoyFilter `spec.priority` (>= 10) and correct targeting.
  * Enhanced model-inference test errors with clearer guidance when payload-processing filters are missing.

* **Chores**
  * Added a script to verify live gateway config contains required `ext_proc` filters and correct ordering.


jland-redhat:
@coderabbitai review

liangwen12year:
Nice fix — priority 10 is the right approach and the root cause analysis is clear.

One question on the readiness gate: `PayloadProcessingEnvoyFilterReady` validates the K8s resource shape (priority, targetRefs) but not whether the `configPatches` actually matched in the live Envoy config — as the code itself notes: "This is a config-shape check (not a live config_dump)."

The failure mode this PR fixes is exactly "inserts silently no-op" — the EnvoyFilter YAML looks correct but the `subFilter` match on `envoy.filters.http.wasm` finds nothing and patches are dropped. Priority 10 fixes the ordering cause, but the wasm anchor could also disappear for other reasons (Kuadrant operator restart, AuthPolicy deleted, etc.). In that case the readiness gate still reports ready while inference returns 404 NR.

The diagnostic script handles the live check, so the tooling coverage is there. But is the `DeploymentsAvailable` condition name accurate if it doesn't verify the filters are actually in the proxy? Not suggesting the controller should query the Envoy admin API (that would be a significant complexity increase), just flagging that the condition could be "True" while body-routed `/v1/*` is broken if the wasm anchor is absent for a non-priority reason.

PR #1244, opened 2026-07-23
fix: propagate MAAS_CONTROLLER_NAMESPACE for RHOAI deploys and default sslmode=disable

jharan1:
https://redhat.atlassian.net/browse/RHOAIENG-78753

## Summary

- **Namespace propagation**: When `--operator-type rhoai` is used, `NAMESPACE` is set to `redhat-ods-applications` but was never exported to `MAAS_CONTROLLER_NAMESPACE`. This caused `deploy_postgresql()`, `setup-database.sh`, and the maas-api wait loop to fall back to `opendatahub`, deploying the POC postgres and its secret to the wrong namespace and polling indefinitely.
- **sslmode default**: The POC postgres container has no TLS configured, so `sslmode=require` (the previous default) causes maas-api to fail with "server refused TLS connection". Changed the default to `sslmode=disable` for both fresh installs and the upgrade path (existing secret migration). Production users who pass `--postgres-connection` are unaffected — they control sslmode in their own connection string.

## Risk analysis

- **Risk rating**: 3
- **Why**: Script-only changes to deploy/setup helpers with no Go or manifest changes. The sslmode default change aligns with what CI already does (`DB_SSLMODE=disable` in `prow_run_smoke_test.sh`). The namespace fix is narrowly scoped to the RHOAI operator-type path. However, the deploy scripts affect all installation flows, and the sslmode default change could theoretically affect edge-case users who relied on the implicit `require` default without configuring TLS on their POC postgres (unlikely but possible).

## Test plan

- [ ] Deploy with `--operator-type rhoai` and verify postgres + secret land in `redhat-ods-applications`
- [ ] Deploy with `--operator-type odh` and verify behavior is unchanged (namespace = `opendatahub`)
- [ ] Verify maas-api connects successfully to the POC postgres with `sslmode=disable`
- [ ] Verify a user-supplied `--postgres-connection` with `sslmode=require` is not overridden
- [ ] Run existing E2E smoke tests


jharan1:
No longer confident in this fix as the file has been refactored since I tested it a few weeks back.

Issue #1246, opened 2026-07-23
feat: per-gateway payload-processing EnvoyFilter copies for multi-tenancy (shared IPP)


jland-redhat:
## Summary

Under **multi-tenancy**, each AITenant has its own Gateway, but `EnvoyFilter/payload-processing` is a singleton whose `targetRefs` is last-writer-wins. Tenant gateways other than the last reconciled one never get `ipp-pre` / `ipp`, so body-based routing / ExternalModel BBR fails (404 NR).

## Freeze-friendly fix (this work)

Create **one EnvoyFilter copy per Gateway** (`payload-processing-<gatewayName>`), still pointing at the **shared** `payload-pre-processing` / `payload-processing` Deployments/Services (request-stateless IPP with a shared CR informer cache).

- Set `priority: 10` so RHCL Kuadrant wasm anchors apply
- Delete legacy singleton `EnvoyFilter/payload-processing` to avoid double ext_proc
- Do **not** create per-tenant IPP pods yet (known follow-up)

## Acceptance criteria

- [ ] Each AITenant Gateway has `EnvoyFilter/payload-processing-<gatewayName>` targeting that Gateway
- [ ] Shared IPP pods remain single Deployment pair in the gateway namespace
- [ ] Body-routed `/v1/*` works on non-default tenant gateways
- [ ] Default gateway still works after singleton cleanup
- [ ] Unit/e2e coverage for per-gateway EnvoyFilter naming

## Follow-up (out of scope / post-freeze)

Unique IPP Deployments per tenant/gateway when we have time.

## Implementation branch

`fix/per-gateway-payload-envoyfilter`

## Related

- EnvoyFilter priority / RHCL: #1239
- Multi-tenancy docs: `docs/content/concepts/multi-tenancy.md`, `docs/content/install/multi-tenant-setup.md`

## Target

**3.5 / multi-tenancy**

PR #1249, opened 2026-07-23
fix: apply an envoy filter for pre/post processing to every gateway not just default

jland-redhat:
## Summary

- Create one `EnvoyFilter` copy per Gateway (`payload-processing-<gatewayName>`), still pointing at the **shared** `payload-pre-processing` / `payload-processing` Deployments/Services, so multi-tenant gateways are not last-writer-wins on a singleton `targetRefs`.
- Set `spec.priority: 10` so RHCL Kuadrant wasm anchors exist before our `INSERT_*` patches.
- Delete the legacy singleton `EnvoyFilter/payload-processing` after apply to avoid double `ext_proc`.
- Truncate long EnvoyFilter names to 60 chars (with hash) instead of failing the 63-char DNS1123 limit.

## Hold / decision needed

**Putting this PR on hold.** Ishita is investigating how hard it would be to run **separate payload-processing (IPP) per unique Gateway** instead of shared IPP pods with per-gateway EnvoyFilter copies.

We should decide between:

1. **This PR (freeze-friendly):** per-gateway EnvoyFilter → shared IPP  
2. **Ishita’s direction:** unique IPP Deployment/Service (and EnvoyFilter) per Gateway  

…before merging, so we don’t ship the shared-IPP approach and then rip it out.

Tracking: https://github.com/opendatahub-io/models-as-a-service/issues/1246

## Test plan

- [ ] Unit: `go test ./pkg/platform/tenantreconcile/…` (naming, priority, legacy EF delete)
- [ ] E2E tenant assertions for per-gateway EnvoyFilter name
- [ ] On a multi-tenant cluster: each AITenant Gateway has `payload-processing-<gateway>`; body-routed `/v1/*` works on non-default gateways; default gateway still works after singleton cleanup
- [ ] Confirm long gateway names produce a ≤60-char EnvoyFilter name

## Risk analysis

- **Risk rating:** 3  
- **Why:** Touches gateway dataplane wiring (EnvoyFilter naming + priority + singleton cleanup). Unit/e2e cover the naming path, but multi-tenant gateway behavior is easy to miss in default smoke; shared IPP is intentional for now and may be superseded if we choose per-gateway IPP.

## Summary by CodeRabbit

* **New Features**
  * Payload processing now creates a dedicated EnvoyFilter for each gateway.
  * Long gateway names are safely shortened while remaining unique.
  * EnvoyFilter ordering is explicitly configured to ensure consistent processing.
* **Bug Fixes**
  * Legacy shared payload-processing resources are cleaned up to prevent duplicate configuration.
* **Tests**
  * Added coverage for gateway-specific naming, cleanup behavior, priority settings, and end-to-end gateway targeting.

### DIFF
--- a/deployment/base/payload-processing/manager/envoy-filter.yaml
+++ b/deployment/base/payload-processing/manager/envoy-filter.yaml
@@ -1,9 +1,14 @@
 apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
+  # Base name; tenant reconcile renames to payload-processing-<gatewayName> so each
+  # AITenant Gateway gets its own copy while IPP Deployments stay shared.
   name: payload-processing
   namespace: openshift-ingress
 spec:
+  # Must run AFTER Kuadrant's EnvoyFilter (default priority 0). Also set by
+  # patchPayloadProcessingEnvoyFilter at reconcile time.
+  priority: 10
   targetRefs:
     - group: gateway.networking.k8s.io
       kind: Gateway
--- a/maas-controller/pkg/platform/tenantreconcile/constants.go
+++ b/maas-controller/pkg/platform/tenantreconcile/constants.go
@@ -5,8 +5,11 @@
 package tenantreconcile
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strings"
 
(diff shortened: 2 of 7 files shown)


jland-redhat:
Replaced by
https://github.com/opendatahub-io/models-as-a-service/pull/1252

PR #1251, opened 2026-07-23
feat: move EnvoyFilter lifecycle to TenantReconciler for per-tenant scoping

tgitelman:
## Summary

- Move per-tenant usage-logs EnvoyFilter lifecycle from `LifecycleReconciler` to `TenantReconciler`, alongside other gateway resources.
- Each `MaasTenantConfig` gets its own EnvoyFilter scoped via `spec.workloadSelector` to that tenant's gateway; collector address and manifest patches live in `tenantreconcile`.
- `Config.spec.usageLogging` changes fan out to all `MaasTenantConfig` CRs via `mapConfigToMaasTenantConfigs`.
- When usage logging is enabled but the filter cannot be applied (empty `MonitoringNamespace`, missing manifest, CRD unavailable), surface `Degraded` with reason `UsageLoggingNotProvided` instead of a dedicated `UsageLogsReady` condition.
- `usageLogging: false` deletes existing filters even when `MonitoringNamespace` is unset.

## Risk analysis

- **Risk rating**: 3
- **Why**: Changes controller reconcile path and gateway EnvoyFilter lifecycle for all tenants when usage logging is enabled. Unit tests cover create/delete/warning paths and `Degraded` aggregation, but full cluster validation (EnvoyFilter CRD present, collector reachable, multi-tenant gateways) is not exercised in default CI smoke.

## Summary by CodeRabbit

- **New Features**
  - Added tenant-specific usage-log routing for default and named tenants.
  - Usage-log configuration now responds to tenant discovery setting changes.
  - Usage logging can be enabled, disabled, or skipped when monitoring configuration is unavailable.
  - Added clearer status reporting for usage-log deployment failures and unavailable logging.

- **Bug Fixes**
  - Usage-log resources are cleaned up when a tenant is deleted or logging is disabled.

- **Tests**
  - Added coverage for routing, configuration changes, cleanup, and status reporting.


maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -290,6 +314,169 @@ func (r *AITenantReconciler) enqueueAITenantForTenantConfig(_ context.Context, o
...
+			return r.deleteEnvoyFilterIfExists(ctx, log, efName)
+		}
+		return err
+	}
+
+	if !ptr.Deref(cfg.Spec.UsageLogging, false) {
+		return r.deleteEnvoyFilterIfExists(ctx, log, efName)
+	}
+
+	// usageLogging is enabled — MonitoringNamespace is required to build the collector address.
```
ahadas: yeah, that could not have happened before https://github.com/opendatahub-io/models-as-a-service/pull/1262 since we assumed MonitoringNamespace is always set but now that it may be empty, we need to think about how to handle this
I tend to think that if MonitoringNamespace is not set, and especially if the EnvoyFilter won't be created in this case, we can add a condition for observability features and rather check that condition here and possibly in other places - wdyt?

maas-controller/pkg/controller/maas/self_deployment_controller.go
```diff
@@ -743,86 +731,6 @@ func patchTenancyProxyImage(res *unstructured.Unstructured) error {
 	return errors.New("proxy container not found in usage-logs-tenancy-proxy deployment")
 }
 
-// ensureUsageLogsEnvoyFilter deploys or removes the OTel usage logs EnvoyFilter based on
-// the Config's usageLogging feature gate. The EnvoyFilter emits structured per-request
-// usage logs (token counts, identity, model) to an OTel Collector via gRPC Access Log Service.
-func (r *LifecycleReconciler) ensureUsageLogsEnvoyFilter(ctx context.Context, log logr.Logger) error {
```
ahadas, line 759: as I see it, we can split the issues we see with the current state to:
1. currently the envoy filter targets only the default gateway that is hard coded in the definition of the envoy filter
2. the envoy filter doesn't respect the telemetry settings for the aitenant

by moving the creation/deletion of envoy filters from this controller to the aitenants controller, we can address both issues but:
1. issue number 2 above is not covered here (intentionally?)
2. I wonder if it wouldn't make sense to keep having a broad handling for a default configuration - currently, each aitenant is configured with telemetry settings, that has some default but there's no "global default". what if we'll introduce default telemetry settings in, e.g., the global Config that can later be overridden by aitenants, if needed. this would allow users to change the default once globally. and if we do that, would we want the aitenant to respect that global defaults and apply them on their aitenant-specific envoy filter, or shall we have a global envoy filter that applies to all gateways of aitenants for which the telemetry settings were not overridden? the problem I see with the latter is that we don't control the gateways and we'll need to label them with certain labels.

@bro-adm @tgitelman we talked about point number 2 above last week offline, wdyt?

maas-controller/pkg/controller/maas/aitenant_controller.go
tgitelman: My code already handles the monitoring namespace empty with silent skip:
 if r.MonitoringNamespace == "" {
		log.V(1).Info("MonitoringNamespace is empty, skipping usage-logs EnvoyFilter")
		return nil
	}
I thought about this, we want to fail on empty field?

ahadas: > I thought about this, we want to fail on empty field?

no, we certainly don't want to fail - empty monitoring namespace is expected on xls

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -290,6 +314,201 @@ func (r *AITenantReconciler) enqueueAITenantForTenantConfig(_ context.Context, o
...
+	gatewayRef := r.gatewayRefFor(aitenant)
+	efName := tenantreconcile.UsageLogsEnvoyFilterName(envoyFilterTenantID(aitenant))
+
+	if r.MonitoringNamespace == "" {
+		apimeta.SetStatusCondition(&aitenant.Status.Conditions, metav1.Condition{
+			Type:               maasv1alpha1.AITenantConditionObservabilityReady,
+			Status:             metav1.ConditionFalse,
+			Reason:             "MonitoringNamespaceNotConfigured",
+			Message:            "Usage-logs EnvoyFilter not deployed: MonitoringNamespace is empty",
+			ObservedGeneration: aitenant.Generation,
```
ahadas: message

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -290,6 +314,201 @@ func (r *AITenantReconciler) enqueueAITenantForTenantConfig(_ context.Context, o
...
+	// Config is cluster-scoped; EnvoyFilter is namespaced. Kubernetes GC does not
+	// enforce cross-namespace owner refs, so this won't cascade-delete the EF when
+	// Config is removed. That's acceptable: Config is managed by a higher-level
+	// operator and outlives tenants; reconcileAITenantDelete explicitly cleans up
+	// the EF, and ensureUsageLogsEnvoyFilter deletes it when usageLogging is off.
+	//
+	// Note: the API server does not currently reject cross-namespace ownerRefs on
+	// namespaced resources. If a future Kubernetes version tightens this constraint,
+	// switch to a label-based ownership model (e.g., maas.opendatahub.io/config-owner).
+	if err := controllerutil.SetOwnerReference(cfg, ef, r.Scheme); err != nil {
```
ahadas: wouldn't it make more sense now to set the AITenant as owner?

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -290,6 +314,201 @@ func (r *AITenantReconciler) enqueueAITenantForTenantConfig(_ context.Context, o
...
+// additional workloadSelector labels, merge them here rather than relying on the manifest value.
+func patchEnvoyFilterWorkloadSelector(ef *unstructured.Unstructured, gatewayName string) error {
+	if err := unstructured.SetNestedStringMap(ef.Object,
+		map[string]string{"gateway.networking.k8s.io/gateway-name": gatewayName},
+		"spec", "workloadSelector", "labels"); err != nil {
+		return fmt.Errorf("write workloadSelector: %w", err)
+	}
+	// targetRefs and workloadSelector are mutually exclusive (Istio 1.26+). Drop any
+	// leftover targetRefs from older manifests so SSA/admission never sees both.
+	unstructured.RemoveNestedField(ef.Object, "spec", "targetRefs")
```
ahadas: need to update the comment to reflect that targetRefs is not supposed to be set (https://github.com/opendatahub-io/models-as-a-service/pull/1313 was merged before 3.5 was released) - it's just a defensive approach

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -647,6 +866,14 @@ func (r *AITenantReconciler) reconcileAITenantDelete(ctx context.Context, aitena
 		return ctrl.Result{}, err
 	}
 
+	// Clean up the per-tenant EnvoyFilter. Transient errors are returned so the
+	// reconciler retries; only NotFound / CRD-not-installed are swallowed by
+	// deleteEnvoyFilterIfExists itself.
+	efName := tenantreconcile.UsageLogsEnvoyFilterName(envoyFilterTenantID(aitenant))
+	if err := r.deleteEnvoyFilterIfExists(ctx, ctrl.LoggerFrom(ctx), efName); err != nil {
```
ahadas: I guess we can then remove this part if we set the AITenant as the owner of the EnvoyFilter

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -715,6 +942,9 @@ func (r *AITenantReconciler) forceRemoveAITenantFinalizer(ctx context.Context, a
 	if err := r.deleteTenantGatewayAuthPolicy(ctx, aitenant); err != nil {
 		log.Error(err, "best-effort deleteTenantGatewayAuthPolicy failed during forced finalizer removal")
 	}
+	if err := r.deleteEnvoyFilterIfExists(ctx, log, tenantreconcile.UsageLogsEnvoyFilterName(envoyFilterTenantID(aitenant))); err != nil {
```
ahadas: and probably this as well

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -290,6 +314,201 @@ func (r *AITenantReconciler) enqueueAITenantForTenantConfig(_ context.Context, o
...
+	if err := r.Get(ctx, client.ObjectKey{Name: maasv1alpha1.ConfigInstanceName}, &cfg); err != nil {
+		if apierrors.IsNotFound(err) {
+			apimeta.SetStatusCondition(&aitenant.Status.Conditions, metav1.Condition{
+				Type:               maasv1alpha1.AITenantConditionObservabilityReady,
+				Status:             metav1.ConditionFalse,
+				Reason:             "ConfigNotFound",
+				Message:            "Usage-logs EnvoyFilter not deployed: Config CR not found",
+				ObservedGeneration: aitenant.Generation,
+			})
+			return r.deleteEnvoyFilterIfExists(ctx, log, efName)
```
ahadas: not sure that's the right call - why removing the envoy-filter when Config is missing?

maas-controller/pkg/controller/maas/aitenant_controller_test.go
```diff
@@ -3293,3 +3296,473 @@ func TestAITenantReconcile_DeletionTimeoutDisabledWhenZero(t *testing.T) {
...
+		g.Expect(err).NotTo(HaveOccurred(), "missing manifest is non-fatal")
+
+		// ObservabilityReady condition should be set to False with EnvoyFilterSkipped reason
+		cond := apimeta.FindStatusCondition(aitenant.Status.Conditions, maasv1alpha1.AITenantConditionObservabilityReady)
+		g.Expect(cond).NotTo(BeNil())
+		g.Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+		g.Expect(cond.Reason).To(Equal("EnvoyFilterSkipped"))
+	})
+
+	t.Run("sets UsageLoggingDisabled condition when usageLogging off and MonitoringNamespace empty", func(t *testing.T) {
```
ahadas: what's UsageLoggingDisabled condition? this description doesn't seem to reflect what's happening in this case

maas-controller/api/maas/v1alpha1/aitenant_types.go
```diff
@@ -26,6 +26,10 @@ const (
 
 	// AITenantConditionReady indicates whether the tenant bootstrap resources are reconciled.
 	AITenantConditionReady = "Ready"
+
+	// AITenantConditionObservabilityReady indicates whether usage-logs observability
+	// resources (EnvoyFilter) are deployed for this tenant.
+	AITenantConditionObservabilityReady = "ObservabilityReady"
```
ahadas: Observability is a broad name that also includes metrics - but here we'll set `ObservabilityReady` to True also when Telemetry is disabled and EnvoyFilter is defined, right?

maas-controller/pkg/controller/maas/self_deployment_controller.go
ahadas: we'll consider this as a follow-up

tgitelman:
@ahadas Thanks for the review — addressed everything in the latest push.

Ownership: Agreed the EnvoyFilter should belong to the AITenant, not Config. I tried SetOwnerReference(aitenant, ef), but it fails because the AITenant lives in ai-tenants while the EnvoyFilter must be in the gateway namespace — controller-runtime rejects cross-namespace owner refs. Instead I'm using applyAITenantMetadata (the same labels/annotations pattern we use for tenant namespaces, roles, and jobs). I kept the explicit delete in reconcileAITenantDelete and forceRemoveAITenantFinalizer for the same reason — garbage collection can't cascade across namespaces.

Config missing: Fixed. We no longer delete the EnvoyFilter when Config is temporarily absent. We set UsageLogsReady=False, keep the existing filter, and requeue until Config comes back.

usageLogging: false: Still deletes the per-tenant EnvoyFilter — that's intentional.

Condition name: Renamed ObservabilityReady → UsageLogsReady since this only tracks the usage-logs EnvoyFilter, not metrics or other observability.

targetRefs comment: Updated — manifests since #1313 already use workloadSelector only; stripping targetRefs is just defensive.

Per-tenant telemetry (your Aug 4 comment on self_deployment_controller.go): Agreed as follow-up per your Sep 1 note — this PR only handles per-gateway EnvoyFilter scoping.

ahadas (approved):

maas-controller/pkg/controller/maas/aitenant_controller.go
```diff
@@ -269,6 +282,14 @@ func (r *AITenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	if err := r.ensureUsageLogsEnvoyFilter(ctx, &aitenant); err != nil {
+		setAITenantPhase(&aitenant, "Failed", "EnvoyFilterReconcileFailed", err.Error())
+		if err2 := r.updateAITenantStatus(ctx, &aitenant, statusSnapshot); err2 != nil {
+			return ctrl.Result{}, err2
+		}
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
```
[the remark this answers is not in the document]
ahadas: yeah. that's a good point

ahadas (commented):
need to rebase and address coderabbitai's comment(s)

maas-controller/cmd/manager/main.go
```diff
@@ -1211,15 +1211,17 @@ func main() {
...
+		Scheme:                  mgr.GetScheme(),
+		APIReader:               mgr.GetAPIReader(),
+		AppNamespace:            infraNamespace,
+		TenantNamespace:         maasSubscriptionNamespace,
+		AITenantNamespace:       aitenantNamespace,
+		GatewayName:             gatewayName,
+		GatewayNamespace:        gatewayNamespace,
+		DeletionTimeout:         aitenantDeletionTimeout,
+		MonitoringNamespace:     monitoringNamespace,
+		EnvoyFilterManifestPath: filepath.Join(usageLogsManifestPath, "envoy-otel-access-log.yaml"),
```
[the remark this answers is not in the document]
tgitelman: The flag defaults to /deployment/components/observability/usage-logs and isn't overridden in our deployment manifests. An empty value would require explicitly passing --usage-logs-manifest-path="", which isn't a supported configuration. In normal operation filepath.Join produces the correct absolute path under the embedded manifests. No change needed.

maas-controller/pkg/controller/maas/aitenant_controller.go
ahadas, line 1: I think these changes need to move to the tenant (rather than aitenant) level, why did you choose this place?

tgitelman:
@coderabbitai review

tgitelman:
@coderabbitai review

maas-controller/api/maas/v1alpha1/aitenant_types.go
```diff
@@ -26,6 +26,10 @@ const (
 
 	// AITenantConditionReady indicates whether the tenant bootstrap resources are reconciled.
 	AITenantConditionReady = "Ready"
+
+	// AITenantConditionUsageLogsReady is deprecated; usage-logs EnvoyFilter readiness is
+	// reported on MaasTenantConfig.status.conditions (type UsageLogsReady).
+	AITenantConditionUsageLogsReady = "UsageLogsReady"
```
ahadas: this one can be dropped

maas-controller/cmd/manager/main.go
ahadas, line 1: why not to simply set UsageLogsEnvoyFilterManifestPath as the other properties are set?

maas-controller/pkg/platform/tenantreconcile/constants.go
```diff
@@ -124,6 +125,7 @@ const (
 	ConditionMaaSPrerequisitesAvailable = "MaaSPrerequisitesAvailable"
 	ConditionDeploymentsAvailable       = "DeploymentsAvailable"
 	ConditionTypeDegraded               = "Degraded"
+	ConditionTypeUsageLogsReady         = "UsageLogsReady"
```
ahadas: I think we don't really need this condition - it would be enough the set Degraded mode when usage-logging is requested but not provided

tgitelman:
@coderabbitai resume

maas-controller/pkg/controller/maas/tenant_controller.go
```diff
@@ -185,6 +188,33 @@ func (r *TenantReconciler) enqueueTenantForAITenant(_ context.Context, obj clien
...
+
+	var tenantList maasv1alpha1.MaasTenantConfigList
+	if err := r.List(ctx, &tenantList); err != nil {
+		ctrl.LoggerFrom(ctx).Error(err, "failed to list MaasTenantConfigs for Config change mapping")
+		return nil
+	}
+	requests := make([]reconcile.Request, 0, len(tenantList.Items))
+	for i := range tenantList.Items {
+		if tenantList.Items[i].Name != maasv1alpha1.MaasTenantConfigInstanceName {
+			continue
```
ahadas, line 209: doesn't it mean that we'll get a list with a single maastenantconfig in it?..
is it intentional?

maas-controller/pkg/controller/maas/tenant_reconcile_test.go
```diff
@@ -906,6 +907,21 @@ func TestAggregateWarningsAndSetDegraded(t *testing.T) {
 			wantStatus:      metav1.ConditionTrue,
 			wantMessage:     "DSCI monitoring stack not available; Perses not available; invalid replica annotation on maas-api; invalid replica annotation on payload-processing",
 		},
+		{
+			name:             "usage logging warning only",
+			usageLogsWarning: "Usage-logs EnvoyFilter not deployed: MonitoringNamespace is empty",
```
ahadas: so on xKS we'd have an empty MonitoringNamespace, intentionally, and we'll get a Degraded mode? I think that in this case, it should rather be just a log. Degraded mode needs to be set when usage-logging can be enabled but it's not - and until we enable that by default, I'd also not set Degraded mode when usageLogging is disabled

maas-controller/pkg/controller/maas/tenant_reconcile.go
```diff
@@ -169,8 +174,19 @@ func (r *TenantReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	usageLogsWarning, err := r.ensureUsageLogsEnvoyFilter(ctx, log, &tenant, platformContext, mcfg)
+	if err != nil {
+		log.Error(err, "usage-logs EnvoyFilter reconciliation failed")
+		setTenantCondition(&tenant, tenantreconcile.ConditionTypeDegraded, metav1.ConditionTrue,
+			"UsageLogsEnvoyFilterReconcileFailed", err.Error())
+		if err := r.Status().Update(ctx, &tenant); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
```
ahadas, line 188: that's a bit weird - it seems we'll get to aggregateWarningsAndSetDegraded only in two cases:
1. MonitoringNamespace is not set (commented on that separately)
2. we failed to apply the EnvoyFilter

in all other cases, including when we fail to read or parse the envoy filter, we're going to set the Degraded condition here and request to reconcile - in some of these cases, we better just continue to aggregateWarningsAndSetDegraded (e.g., if we failed to parse EnvoyFilter, there's no point in keep trying) - we need to get to the code above only in cases there's a change something would change the next time we reconcile

maas-controller/pkg/controller/maas/tenant_controller.go
tgitelman: Yes, intentional — same pattern as AITenant fan-out. Name is always default-tenant; namespace is what differs.

tgitelman: Yes, intentional. Each tenant namespace has its own MaasTenantConfig (default-tenant), and when usageLogging changes on Config we need to reconcile all of them so each tenant’s EnvoyFilter is created or removed. The loop enqueues one reconcile per namespace. Test: TestTenantReconcile_ConfigChangeEnqueuesAllMaasTenantConfigs.

maas-controller/pkg/controller/maas/tenant_reconcile_test.go
tgitelman: Sorry what is xKS?

ahadas: "The term xKS is a shorthand notation or collective acronym used in cloud computing and container orchestration to refer to major managed Kubernetes services across different cloud providers, typically substituting "x" with specific provider letters (such as EKS for AWS, AKS for Azure, and GKS/GKE for Google Cloud"

maas-controller/pkg/controller/maas/tenant_controller.go
ahadas: so is this check redundant then? (or in other words: when would we get to the `continue` statement)

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
```diff
@@ -0,0 +1,223 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	maasv1alpha1 "github.com/opendatahub-io/models-as-a-service/maas-controller/api/maas/v1alpha1"
+	"github.com/opendatahub-io/models-as-a-service/maas-controller/pkg/platform/tenantreconcile"
+)
+
+const usageLogsEnvoyFilterManifestPath = "/deployment/components/observability/usage-logs/envoy-otel-access-log.yaml"
+
+// deleteUsageLogsEnvoyFilterIfDisabled removes the per-tenant filter when Config has usageLogging
+// off. Called before reconcile gates so a blocked tenant still stops exporting usage logs.
```
ahadas: do we have something like this for the envoy-filter for IPP? not sure I understand why we need this - seems that `ensureUsageLogsEnvoyFilter` is sufficient...

maas-controller/pkg/controller/maas/tenant_reconcile.go
ahadas: how about aligning ensureUsageLogsEnvoyFilter with other similar functions here (e.g., checkDependenciesAndPrerequisites) and let it return a result? I think it would help us simplify this part a bit

maas-controller/pkg/controller/maas/tenant_controller.go
ahadas: ok, so can we drop the name check then?

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
ahadas, line 40: do we have something like this for the envoy-filter for IPP? not sure I understand why we need this, seems that `ensureUsageLogsEnvoyFilter` is sufficient...

tgitelman: CodeRabbit flagged and issue that if reconcile returns early (gateway not ready, prereqs, etc.) and usageLogging is false, we never reach ensureUsageLogsEnvoyFilter at the end, so the filter can stay until a full reconcile succeeds. Early delete covers that gap. I will drop this part if you suggest so.

ahadas: I see, you refer to [this comment](https://github.com/opendatahub-io/models-as-a-service/pull/1251#discussion_r3949022562) - I guess it would make more sense when usage logging is enabled by default, so we can keep it

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
```diff
@@ -0,0 +1,241 @@
...
+func (r *TenantReconciler) ensureUsageLogsEnvoyFilter(
+	ctx context.Context,
+	log logr.Logger,
+	tenant *maasv1alpha1.MaasTenantConfig,
+	platformContext tenantreconcile.PlatformContext,
+	mcfg *maasv1alpha1.Config,
+) (string, *ctrl.Result, error) {
+	tenantID, err := tenantreconcile.TenantIdentifierFor(tenant)
+	if err != nil {
+		return usageLogsEnvoyFilterOutcome(log, mcfg, err)
```
ahadas: here we don't need that filter but we can just return the error, right?

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	efName := tenantreconcile.UsageLogsEnvoyFilterName(tenantID)
+	gatewayName := platformContext.GatewayRef.Name
+
+	if !ptr.Deref(mcfg.Spec.UsageLogging, false) {
+		if err := r.deleteUsageLogsEnvoyFilterIfExists(ctx, log, efName); err != nil {
+			return usageLogsEnvoyFilterOutcome(log, mcfg, err)
```
ahadas: I think it's the same case here, right?

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
```diff
@@ -0,0 +1,241 @@
...
+		return false, fmt.Errorf("patch workloadSelector gateway in EnvoyFilter: %w", err)
+	}
+
+	ef.SetName(efName)
+	ef.SetNamespace(r.GatewayNamespace)
+	applyUsageLogsEnvoyFilterMetadata(ef, tenant)
+
+	if err := r.Patch(ctx, ef, client.Apply, client.ForceOwnership, client.FieldOwner("maas-controller")); err != nil {
+		if apimeta.IsNoMatchError(err) {
+			log.Info("EnvoyFilter CRD not available, skipping usage-logs EnvoyFilter")
```
ahadas, line 158: I think this would be the only case where we may want to set Degraded mode as if "the user asked to enable usage logging, but something is missing in order to enable it so let's reflect it to the user" - am I right?

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
tgitelman: Done

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
tgitelman: Yes — in the latest push Degraded is only set when usageLogging=true and we soft-skip because the manifest or EnvoyFilter CRD isn’t available. Everything else returns error through the normal reconcile path.

ahadas (approved):

jland-redhat (approved):
lgtm

PR #1255, opened 2026-07-24
feat: secure allowedRoutes defaults for Gateway setup scripts

somya-bhatnagar:
## Summary

Jira: [RHOAIENG-72987](https://redhat.atlassian.net/browse/RHOAIENG-72987)

Replaces the insecure `allowedRoutes: namespaces: from: All` default in all customer-facing Gateway API scripts with configurable, secure options.

**Changes:**
- `setup-gateway.sh` — adds `ALLOWED_ROUTE_NAMESPACES` and `NAMESPACE_SELECTOR_LABELS` env vars; clusterip inline YAML now defaults to `from: Same`; warns when no config is provided
- `create-ai-tenant.sh` — same env vars; sources shared `deployment-helpers.sh` instead of an inline duplicate; warns when no config is provided
- `deployment-helpers.sh` — adds `build_allowed_routes_yaml`, `build_allowed_routes_json`, `patch_gateway_allowed_routes` helpers with full input validation
- `deploy.sh` — defaults Gateway `allowedRoutes` to the app namespace (`$NAMESPACE`) and includes `$MODEL_NAMESPACE` when set, so maas-api and model HTTPRoutes both attach correctly
- `maas-gateway-api.yaml` — route-mode kustomize manifest changed to `from: Same`
- `test/e2e/scripts/prow_run_smoke_test.sh` — exports `MODEL_NAMESPACE` so `deploy.sh` includes the model namespace in Gateway allowedRoutes
- `scripts/README.md` — documents new env vars for both scripts

**Behaviour:**

| Scenario | `allowedRoutes` result |
|---|---|
| `setup-gateway.sh` standalone, no env vars | `from: Same` (secure default) + warning |
| `deploy.sh` ODH (no `MODEL_NAMESPACE`) | `from: Selector → opendatahub` |
| `deploy.sh` ODH + `MODEL_NAMESPACE=llm` | `from: Selector → opendatahub, llm` |
| `deploy.sh` RHOAI + `MODEL_NAMESPACE=llm` | `from: Selector → redhat-ods-applications, llm` |
| `ALLOWED_ROUTE_NAMESPACES=ns1,ns2` | `from: Selector → [ns1, ns2]` |
| `NAMESPACE_SELECTOR_LABELS=key=val` | `from: Selector → matchLabels` |
| Existing gateway with `from: All` (re-run) | Auto-upgraded to `from: Same` or configured Selector |
| `create-ai-tenant.sh`, no env vars | `from: Same` + warning to set app/model ns |
| `DRY_RUN=true` | No cluster changes, correct log output |

**Why `from: All` is kept in the route-mode manifest as the starting point but immediately overridden:**
The route-mode manifest (`maas-gateway-api.yaml`) is now `from: Same`. `deploy.sh` always computes and passes `ALLOWED_ROUTE_NAMESPACES` (app ns + model ns), so `patch_gateway_allowed_routes` patches to `from: Selector` on every deploy. Standalone use of `setup-gateway.sh` without env vars keeps `from: Same`. The auto-upgrade path also promotes any pre-existing `from: All` gateway to the secure default on re-run.

## Live cluster testing (OCP 4.21 / ROSA)

All tests run on `api.nh46a-qqkvu-8yf.hlsm.p3.openshiftapps.com`.

**1. Standalone `setup-gateway.sh` — default `from: Same` + warning**
```
[INFO]   AllowedRoutes: from: Same (secure default)
[WARN]   MaaS HTTPRoutes attach from the app/model namespaces, not openshift-ingress.
[WARN]   Set ALLOWED_ROUTE_NAMESPACES (e.g. opendatahub,llm) or NAMESPACE_SELECTOR_LABELS for MaaS.
```
```json
{ "namespaces": { "from": "Same" } }
```

**2. `ALLOWED_ROUTE_NAMESPACES=opendatahub,llm` (deploy.sh ODH + MODEL_NAMESPACE=llm)**
```json
{
  "namespaces": {
    "from": "Selector",
    "selector": {
      "matchExpressions": [
        { "key": "kubernetes.io/metadata.name", "operator": "In", "values": ["opendatahub","llm"] }
      ]
    }
  }
}
```

**3. Auto-upgrade `from: All` → `from: Same` on re-run (no env vars)**
```
[INFO]   Updating Gateway allowedRoutes (current: All)...
```
```json
{ "namespaces": { "from": "Same" } }
```

**4. `NAMESPACE_SELECTOR_LABELS=gateway.opendatahub.io/maas-access=true`**
```json
{
  "namespaces": {
    "from": "Selector",
    "selector": { "matchLabels": { "gateway.opendatahub.io/maas-access": "true" } }
  }
}
```

**5. ClusterIP mode — `ALLOWED_ROUTE_NAMESPACES=opendatahub,llm`**
Gateway + Route created, `from: Selector → [opendatahub, llm]` ✅

**6. `create-ai-tenant.sh` — default `from: Same` + warning**
```
[WARN] No ALLOWED_ROUTE_NAMESPACES or NAMESPACE_SELECTOR_LABELS set; using from: Same.
[WARN] MaaS HTTPRoutes attach from the app/model namespaces — set ALLOWED_ROUTE_NAMESPACES (e.g. opendatahub,llm)
```
```json
{ "namespaces": { "from": "Same" } }
```

**7. `DRY_RUN=true` — no cluster changes**
```
[INFO]   [DRY RUN] Would create/update Gateway maas-default-gateway
[INFO] DRY RUN MODE - no changes were applied
```
Gateway confirmed absent after dry run ✅

## Risk analysis

- **Risk rating**: 2
- **Why**: Changes are additive — new env vars with backward-compatible defaults. `deploy.sh` behaviour changes from `from: All` to `from: Selector` scoped to app + model namespace, which is strictly more secure and functionally equivalent for standard MaaS deployments. The `patch_gateway_allowed_routes` helper uses `op: add` (RFC 6902) which is safe for both present and absent `allowedRoutes` fields. The auto-upgrade path for existing `from: All` gateways is tested on cluster. The e2e smoke test exports `MODEL_NAMESPACE` so the CI gateway gets the correct allowedRoutes. Risk is low; no controller or CRD changes.

[RHOAIENG-72987]: https://redhat.atlassian.net/browse/RHOAIENG-72987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the model namespace setting is exported and applied during MaaS deployments so Gateway allowed-routes restrictions work consistently.
* **New Features**
  * Added configurable controls to restrict which namespaces can attach HTTPRoutes to the MaaS Gateway using `ALLOWED_ROUTE_NAMESPACES` or `NAMESPACE_SELECTOR_LABELS`, with a secure default when unset.
  * Updates/patches allowed-routes on existing Gateways.
* **Documentation**
  * Updated multi-tenant and scripting guides with new environment variables, precedence/default behavior, and required namespace labeling steps.


somya-bhatnagar:
@coderabbitai review

somya-bhatnagar:
**Root cause of CI failure identified and fixed.**

The `gateway_allowed_namespaces` computation in `deploy.sh` only included `$NAMESPACE` (opendatahub) and `$MODEL_NAMESPACE` (llm) — it was missing the infra namespace (`odh-ai-gateway-infra`) where `maas-api-route` lives.

This caused:
- `maas-api-route` in `odh-ai-gateway-infra` → not accepted by gateway (`from: Selector → [opendatahub, llm]`)
- Gateway returned `404 route_not_found` for all `/maas-api/v1/...` calls
- E2E test couldn't create API keys/subscriptions via the gateway
- All 4 e2e MaaSModelRefs stuck at `GovernanceAttached: False`

**Fix:** always include the infra namespace derived from `derive_infra_namespace "$NAMESPACE"`.

Result for ODH: `ALLOWED_ROUTE_NAMESPACES="opendatahub,odh-ai-gateway-infra,llm"`
Result for RHOAI: `ALLOWED_ROUTE_NAMESPACES="redhat-ods-applications,redhat-ai-gateway-infra,llm"`

somya-bhatnagar:
`/retest`

Second run: 234/235 tests pass. Remaining 1 failure (`test_rate_limit_enforced_per_tenant`, 500 AUTH_FAILURE from per-tenant maas-api) is unrelated to this PR's allowedRoutes changes — per-tenant gateways in the e2e framework use label-based `from: Selector` (not `create-ai-tenant.sh`), so PR #1255 doesn't affect them. The error is an internal maas-api token-generation failure, likely intermittent.

deployment/base/networking/maas/maas-gateway-api.yaml
```diff
@@ -21,7 +21,7 @@ spec:
      protocol: HTTPS
      allowedRoutes:
        namespaces:
-         from: All
+         from: Same
```
ishitasequeira, line 24: Multi-tenant concern: per-tenant Gateways need routes from `ai-tenant-<name>`, infra ns, and model ns not just `opendatahub,llm`. Default `from: Same` on `create-ai-tenant.sh` will break tenant bootstrap unless configured. For MT, please document/recommend `NAMESPACE_SELECTOR_LABELS` (as e2e does) rather than static namespace lists, and update `multi-tenant-setup.md` accordingly.

somya-bhatnagar: @ishitasequeira:
This has been fixed — per-tenant Gateways need HTTPRoutes from the infra namespace, tenant namespace, and model namespaces, none of which are predictable at script invocation time, so a static namespace list won't work for multi-tenant deployments.

Fixed in https://github.com/opendatahub-io/models-as-a-service/commit/83b115bf404ca93a3408e41d3874b3a2bf8ba693:

multi-tenant-setup.md — replaced both from: All instances with the per-gateway label selector pattern already established in e2e (maas.opendatahub.io/gateway-access-<tenant>=true), with explicit instructions to label the infra and model namespaces after Gateway creation.

create-ai-tenant.sh header — added a dedicated multi-tenant section recommending NAMESPACE_SELECTOR_LABELS with this label pattern and the labeling workflow for infra/model namespaces.

Tested on a live ROSA cluster:

NAMESPACE_SELECTOR_LABELS="maas.opendatahub.io/gateway-access-red-team=true" → Gateway correctly gets from: Selector → matchLabels ✅
Labeled odh-ai-gateway-infra and llm → only those namespaces match the selector ✅
Default (no env vars) → from: Same + WARN pointing to the label approach ✅

docs/content/install/multi-tenant-setup.md
```diff
@@ -101,9 +120,16 @@ oc get gateway ${TENANT_NAME} -n ${GATEWAY_NAMESPACE}
...
+    For multi-tenant deployments use the per-gateway label selector and label namespaces manually:
     ```bash
-    ./scripts/create-ai-tenant.sh red-team
+    NAMESPACE_SELECTOR_LABELS="maas.opendatahub.io/gateway-access-red-team=true" \
+      ./scripts/create-ai-tenant.sh red-team
+
+    # Label the infra namespace and any model namespaces after the Gateway is created
+    oc label namespace odh-ai-gateway-infra \
+      maas.opendatahub.io/gateway-access-red-team=true --overwrite
+    # oc label namespace llm maas.opendatahub.io/gateway-access-red-team=true --overwrite
```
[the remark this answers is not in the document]
somya-bhatnagar: Fixed in 7ed7d408 — split the single fenced block (which had a blank line inside it causing MD046) into two separate fenced blocks with a prose line between them.

scripts/deployment-helpers.sh
```diff
@@ -1783,3 +1783,232 @@ dump_llmis_diagnostics() {
...
+#   desired configuration. Patches when:
+#     - Current from is "All" (upgrades insecure default to secure Same, or to
+#       the configured selector when env vars are set), OR
+#     - ALLOWED_ROUTE_NAMESPACES or NAMESPACE_SELECTOR_LABELS is set (applies user config)
+#   Skips when already at a secure non-All state and no custom config is requested.
+#
+# Arguments:
+#   gateway_name      - Name of the Gateway resource
+#   gateway_namespace - Namespace of the Gateway
+patch_gateway_allowed_routes() {
```
liangwen12year, line 1982: `patch_gateway_allowed_routes` only patches `listeners[0]` — The jsonpath read and JSON patch both hardcode index 0. This is fine for the current single-listener scripts, but the `multi-tenant-setup.md` doc shows customers creating two-listener Gateways (HTTP + HTTPS). If a customer follows the docs and later re-runs a script that calls `patch_gateway_allowed_routes`, only the first listener gets patched. More importantly, any future script change that adds a second listener would silently leave it unsecured. Should at minimum document this limitation, or loop over all listeners.

somya-bhatnagar: Good catch. Fixed in d4015c8c — `patch_gateway_allowed_routes` now counts listeners dynamically and builds a single JSON patch array covering every index, so all listeners are patched consistently.

Tested on a live ROSA cluster against a two-listener Gateway (HTTP + HTTPS, both initially `from: All`):

- No env vars → both listeners upgraded to `from: Same` ✅
- `ALLOWED_ROUTE_NAMESPACES=opendatahub,llm` → both listeners get `from: Selector → [opendatahub, llm]` ✅
- No env vars, already `from: Selector` → skip (no patch) ✅
- Re-run same custom config → `patched (no change)` ✅

jland-redhat (approved):
lgtm

ishitasequeira (approved):
lgtm

PR #1262, opened 2026-07-24
fix(RHOAIENG-78942): gracefully handle empty monitoring-namespace

tsisodia10:
## Summary

On xKS clusters, the `monitoring-namespace` parameter in `params.env` may be empty (the ai-gateway-operator sets it from the platform config, which on xKS has no monitoring namespace). This causes the maas-controller to crash immediately on startup with exit code 1 and **no log output**, making it very hard to debug.

**Root cause:** `validation.IsDNS1123Label("")` rejects empty strings, and `os.Exit(1)` runs before `ctrl.SetLogger()` is called — so there's no log output at all.

**Fix:**
- Allow empty `--monitoring-namespace` to gracefully disable observability (matching `--infra-namespace` semantics which already allows empty)
- Add namespace existence check in `ensureObservability` to skip when the namespace doesn't exist on the cluster
- Unit tests for both scenarios

## Test plan

- [x] `go build ./cmd/manager/` compiles cleanly
- [x] `go test ./pkg/controller/maas/` — all existing tests pass
- [x] New test `TestEnsureObservability_EmptyMonitoringNamespace` passes (both "empty" and "nonexistent" cases)
- [x] Manually verified on AKS cluster: controller starts and runs with empty monitoring-namespace instead of crashing

Fixes: https://redhat.atlassian.net/browse/RHOAIENG-78942

## Summary by CodeRabbit

* **New Features**
  * Monitoring and observability are now optional when no monitoring namespace is configured.
  * Observability setup is skipped when the configured namespace does not exist.

* **Bug Fixes**
  * Invalid configured monitoring namespaces now produce clear validation errors.
  * Missing monitoring namespaces no longer cause unnecessary reconciliation failures.

### DIFF
--- a/maas-controller/cmd/manager/main.go
+++ b/maas-controller/cmd/manager/main.go
@@ -935,11 +935,15 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	if errs := validation.IsDNS1123Label(monitoringNamespace); len(errs) > 0 {
-		setupLog.Error(stderrors.New("invalid monitoring namespace"),
-			"--monitoring-namespace must be a valid Kubernetes namespace name",
-			"namespace", monitoringNamespace, "errors", errs)
-		os.Exit(1)
+	// Allow empty monitoring-namespace to disable observability features (e.g. on xKS
+	// where the monitoring namespace may not exist). Non-empty values must be valid.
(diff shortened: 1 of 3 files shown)


jland-redhat (dismissed):
lgtm

ishitasequeira (approved):
lgtm

Issue #79, opened 2025-09-15
Fix hardcoded user references preventing multi-user shared environment


yossiovadia:
## Problem

The MaaS platform contains hardcoded references to user 'noyitz' throughout the codebase, preventing it from working in a truly shared multi-user environment. This causes issues where:

1. **Request Simulator fails** for users other than 'noyitz' with DNS resolution errors
2. **Token creation** always assigns tokens to 'noyitz' instead of the authenticated user  
3. **Environment inconsistency** between different developers using create-my-env.sh

### Issue 1: DNS Resolution Failures
- **Before**: `qwen3-llm.summit-gpu.octo-emerging.redhataicoe.com` → DNS failure
- **After**: `qwen3-llm.apps.summit-gpu.octo-emerging.redhataicoe.com` → HTTP 401 (reachable)

### Issue 2: Hardcoded User References
Multiple hardcoded 'noyitz' references found in:
- `apps/frontend/src/components/TokenManagement.tsx:251` - hardcoded user_id in token creation
- `apps/backend/src/routes/tokens.ts:131,235` - hardcoded fallback user ID  
- `apps/backend/server.py:37,1120,1130` - hardcoded DEFAULT_USER_ID and team references

## Solution Implemented

✅ **Fixed Model Endpoint URLs**: Added missing `.apps.` subdomain to simulator endpoints  
✅ **Dynamic User Detection**: Replace hardcoded 'noyitz' with authenticated user from `oc whoami`  
✅ **Frontend Integration**: TokenManagement component now fetches current user via cluster status API  
✅ **Backend Authentication**: TypeScript routes use OpenShift authentication to get current user  
✅ **Legacy Compatibility**: Updated Python server.py to use configurable DEFAULT_USER_ID

## Files Changed

- `apps/backend/src/routes/simulator.ts` - Fixed model endpoint URLs with `.apps.` subdomain
- `apps/frontend/src/components/TokenManagement.tsx` - Added dynamic user detection  
- `apps/backend/src/routes/tokens.ts` - Enhanced to use `oc whoami` for user ID
- `apps/backend/server.py` - Made DEFAULT_USER_ID configurable

## Testing Results

- ✅ Request Simulator now works for user 'yossiovadia' (previously only worked for 'noyitz')
- ✅ Token creation assigns tokens to authenticated user instead of hardcoded 'noyitz'  
- ✅ DNS resolution fixed: endpoints now resolve correctly with `.apps.` subdomain
[...]

PR #1263, opened 2026-07-24
fix(RHOAIENG-78602): auto-select xKS manifest overlay for maas-api

tsisodia10:
## Summary

On non-OpenShift Kubernetes clusters, the `maas-api` Deployment references OCP-specific resources:
- `openshift-service-ca.crt` ConfigMap (auto-injected by OCP's service-ca operator)
- `maas-api-serving-cert` Secret (auto-created by OCP's serving-cert-signer)

These don't exist on vanilla Kubernetes, causing the pod to be stuck in `ContainerCreating` indefinitely.

**Fix:**
- Adds a new `maas-api/deploy/overlays/xks/` kustomize overlay that sources TLS from cert-manager (`opendatahub-ca` Secret) instead of OCP service-ca
- Introduces `ManifestPathForPlatform(isOCP bool)` that auto-selects the correct overlay based on whether `config.openshift.io` APIs are available (same detection already used for TLS profile loading)
- `MAAS_PLATFORM_MANIFESTS` env var still takes precedence for manual override
- Existing Dockerfile already copies `maas-api/deploy/` into the image — no Dockerfile changes needed

## Test plan

- [x] `go build ./cmd/manager/` compiles cleanly
- [x] `go test ./pkg/platform/tenantreconcile/` — new `TestManifestPathForPlatform` passes
- [x] `go test ./pkg/controller/maas/` — all existing tests pass
- [x] Validated on AKS cluster: maas-api starts with cert-manager certs instead of OCP service-ca

Fixes: https://redhat.atlassian.net/browse/RHOAIENG-78602

## Summary by CodeRabbit

* **New Features**
  * Added platform-aware default selection of deployment manifests (OpenShift vs Kubernetes) when no custom manifest path is provided.
* **Bug Fixes**
  * Improved XKS certificate and CA handling by updating how the service CA is sourced and preventing automatic OpenShift certificate/CA bundle injection so cert-manager can manage it.
* **Tests**
  * Added unit tests covering platform-specific manifest path selection and verifying environment-based overrides.

### DIFF
--- a/maas-controller/pkg/platform/tenantreconcile/kustomize.go
+++ b/maas-controller/pkg/platform/tenantreconcile/kustomize.go
@@ -182,3 +182,17 @@ func DefaultManifestPath() string {
 	}
 	return "../maas-api/deploy/overlays/odh"
 }
+
+// ManifestPathForPlatform returns the appropriate kustomize overlay path based on
+// whether the cluster is OpenShift (isOCP=true) or vanilla Kubernetes (isOCP=false).
+// The xKS overlay avoids OCP-specific resources like service-serving-certs and
+// service-ca ConfigMap injection that don't exist on non-OpenShift clusters.
+func ManifestPathForPlatform(isOCP bool) string {
+	if v := os.Getenv("MAAS_PLATFORM_MANIFESTS"); v != "" {
+		return v
+	}
+	if isOCP {
+		return "/maas-api/deploy/overlays/odh"
+	}
+	return "/maas-api/deploy/overlays/xks"
+}
--- a/deployment/base/maas-controller/rbac/clusterrole.yaml
+++ b/deployment/base/maas-controller/rbac/clusterrole.yaml
@@ -119,6 +119,17 @@ rules:
   - delete
   - get
 - apiGroups:
+  - cert-manager.io
+  resources:
+  - certificates
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
   - config.openshift.io
   resources:
   - apiservers
--- a/maas-api/deploy/overlays/xks/certificate.yaml
+++ b/maas-api/deploy/overlays/xks/certificate.yaml
@@ -1,0 +1,15 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: maas-api-serving-cert
+spec:
+  secretName: maas-api-serving-cert
+  issuerRef:
+    name: rhai-ca-issuer
+    kind: ClusterIssuer
+    group: cert-manager.io
+  duration: 2160h # 90 days
+  renewBefore: 360h # 15 days
+  dnsNames:
+    - "maas-api.opendatahub.svc"
+    - "maas-api.opendatahub.svc.cluster.local"
--- a/maas-api/deploy/overlays/xks/kustomization.yaml
+++ b/maas-api/deploy/overlays/xks/kustomization.yaml
@@ -11,6 +11,8 @@ resources:
   - ../../../../deployment/base/maas-controller/policies
   - ../../../../deployment/base/payload-processing/default
 
+namespace: opendatahub
+
 patches:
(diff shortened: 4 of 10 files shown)


maas-controller/pkg/platform/tenantreconcile/kustomize.go
```diff
@@ -182,3 +182,17 @@ func DefaultManifestPath() string {
 	}
 	return "../maas-api/deploy/overlays/odh"
 }
+
+// ManifestPathForPlatform returns the appropriate kustomize overlay path based on
+// whether the cluster is OpenShift (isOCP=true) or vanilla Kubernetes (isOCP=false).
+// The xKS overlay avoids OCP-specific resources like service-serving-certs and
+// service-ca ConfigMap injection that don't exist on non-OpenShift clusters.
+func ManifestPathForPlatform(isOCP bool) string {
```
jland-redhat, line 190: The logic here is a little confusing to me, we are passing in `tlsConfig.available` and using that to determine if the platform is OCP?

The description said this was based on `config.openshift.io` being available but not totally following that logic

jland-redhat (changes requested):
Left question above

maas-controller/pkg/platform/tenantreconcile/kustomize.go
tsisodia10: @Jaland Good catch... you're right, it was unclear. tlsConfig.available indicates whether config.openshift.io API exists (set by fetchTLSProfileWithRetry), which is the authoritative signal for OCP vs xKS. But passing it directly into ManifestPathForPlatform(isOCP) hid that connection.

Fixed in the latest push.. extracted it into a named isOCP variable with an explaining comment at the call site.
Note: this is purely a fallback, in production, MAAS_PLATFORM_MANIFESTS is injected by ai-gateway-operator (PR # 79 in that repo), so this code path isn't hit.

tsisodia10: Sorry this is the right #79 https://github.com/opendatahub-io/ai-gateway-operator/pull/79

zdtsw (commented):
LGTM

ishitasequeira:
For xKS, please add `namespace: opendatahub` to the new overlay (same as odh) so maas-api lands in `redhat-ai-gateway-infra`, not `maas-api`

jland-redhat (dismissed):
lgtm

ishitasequeira (dismissed):
lgtm

tsisodia10:
@ishitasequeira @Jaland 
**fix [RHOAIENG-78602](https://redhat.atlassian.net/browse/RHOAIENG-78602): controller-managed serving cert for xKS**
During E2E validation, maas-api pods were stuck in ContainerCreating because the TLS secret (maas-api-serving-cert) was being created in `redhat-ods-applications` by the Helm chart hook, but maas-api actually runs in the infrastructure namespace `(redhat-ai-gateway-infra)`. Kubernetes Secrets are namespace-scoped, so the pod couldn't find the secret.

Root cause: The Helm chart hook uses applicationsNamespace for the Certificate, but the maas-controller deploys maas-api to the derived infra namespace. The hook doesn't know (and shouldn't need to know) where the controller places workloads.

Solution: Make the controller own its own serving Certificate. WAdded a cert-manager Certificate to the xKS overlay, so it goes through the same render pipeline as the Deployment and Service. The controller's postBuildTransform places it in the infra namespace automatically, and patchMaaSAPIServingCert rewrites the dnsNames to use the actual runtime namespace, no hardcoded values, works for multi-tenant too. This removes the dependency on the Helm hook for cert creation and ensures the secret always lands in the correct namespace.



[RHOAIENG-78602]: https://redhat.atlassian.net/browse/RHOAIENG-78602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ishitasequeira (approved):
lgtm

PR #1265, opened 2026-07-24
fix: wait for gateway AuthPolicy Enforced before e2e API calls

jland-redhat:
## Summary
- Fix Konflux e2e flakes where API key creates fail with empty HTTP 403 right after tests create/delete `MaaSAuthPolicy` resources (gateway `maas-gateway-auth` still reconciling).
- Add `_wait_for_gateway_auth_enforced()` and centralize empty-403 / `AUTH_FAILURE` retries in `test_helper`; wait after AuthPolicy cleanup and before minting keys in the flaky cases.
- Improve failure messages so a never-Enforced AuthPolicy is obvious instead of a bare empty 403.

Observed on [PR #1260](https://github.com/opendatahub-io/models-as-a-service/pull/1260) group-test (`maas-group-test-ngjk6`): `test_search_without_subscription_returns_all` and `test_injected_identity_headers_ignored` — unrelated to that PR's controller change.

## Test plan
- [ ] Konflux `maas-group-test` / e2e-maas-openshift passes
- [ ] Re-run or watch for recurrence of empty-403 failures on API key create after AuthPolicy churn

## Risk analysis
- **Risk rating**: 1
- **Why**: E2E test harness only (waits + retries). No product/runtime code changes; may add a bit of poll time after AuthPolicy mutations but failure mode becomes clearer if Enforced never arrives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end API key, subscription, and search reliability during gateway authorization propagation and Kuadrant auth policy transitions.
  * Reduced intermittent 403/empty-response flakes by synchronizing requests with gateway authorization readiness.
  * Added retries for transient gateway authentication failures during API key creation.
  * Improved failure messages to better explain when gateway authorization is not yet enforced.

* **Tests**
  * Updated negative security and multitenancy E2E flows to wait for stable gateway authorization before running assertions.

### DIFF
--- a/test/e2e/tests/test_api_keys.py
+++ b/test/e2e/tests/test_api_keys.py
@@ -57,57 +57,27 @@ from test_helper import (
     _get_cr,
     _maas_api_url,
     _ns,
+    _request_with_gateway_retry,
     _sa_to_user,
     _scale_controller_down,
     _scale_controller_up,
+    _wait_for_gateway_auth_enforced,
     _wait_for_maas_subscription_phase,
     _wait_reconcile,
 )
 
 log = logging.getLogger(__name__)
 
-# ---------------------------------------------------------------------------
-# Gateway propagation retry helper
-# ---------------------------------------------------------------------------
-# Kuadrant gateway propagation can lag behind MaaS CR readiness.
-# MaaSAuthPolicy "Active" means the controller created the Kuadrant AuthPolicy,
-# but Envoy may not have loaded it yet.  Retry on empty 403 (gateway rejection).
-GATEWAY_PROPAGATION_RETRIES = 6
-GATEWAY_PROPAGATION_DELAY = 5  # seconds
-
-
-def _request_with_gateway_retry(method, url, retries=GATEWAY_PROPAGATION_RETRIES, **kwargs):
-    """Make an HTTP request, retrying on transient gateway/auth errors.
-
-    Retries on:
-    - Empty 403: Envoy hasn't loaded the AuthPolicy yet.
-    - 500 with AUTH_FAILURE: Authorino race condition during AuthConfig updates
-      (metadata evaluators not yet linked after policy reconciliation).
-
-    Returns the last response — the caller's assertion will surface the
-    failure clearly if the gateway never becomes ready.
-    """
-    for attempt in range(1, retries + 1):
-        r = method(url, timeout=kwargs.pop("timeout", 30), verify=kwargs.pop("verify", TLS_VERIFY), **kwargs)
-        retryable = (r.status_code == 403 and not r.text.strip()) or (
-            r.status_code == 500 and "AUTH_FAILURE" in r.text
-        )
-        if retryable and attempt < retries:
-            log.info("Gateway returned %d (attempt %d/%d), retrying in %ds...",
-                     r.status_code, attempt, retries, GATEWAY_PROPAGATION_DELAY)
-            time.sleep(GATEWAY_PROPAGATION_DELAY)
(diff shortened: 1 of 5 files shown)


jland-redhat:
@coderabbitai review

test/e2e/tests/test_api_keys.py
```diff
@@ -57,57 +57,27 @@
...
-    """Wait for the gateway to start forwarding requests before running tests.
+    """Wait for gateway AuthPolicy enforcement before running API key tests.
 
-    Makes a single request with gateway retry to absorb the propagation delay.
-    All subsequent tests can then make requests directly without retry logic.
+    MaaSAuthPolicy churn in earlier modules (or this session) can leave
+    maas-gateway-auth reconciling; empty 403s follow until Enforced=True and
+    Envoy loads the config.
     """
+    _wait_for_gateway_auth_enforced(timeout=120)
```
yu-teo: ` _wait_for_gateway_auth_enforced` calls `_authpolicy_condition(cr, "Accepted")` then `_authpolicy_condition(cr, "Enforced")`, so each iterates the full conditions list. May I suggest a small single-pass helper that would be cleaner like so?
```
  def _authpolicy_conditions(cr, *types):
      """Return {type: (status, reason, message)} for requested condition types."""
      wanted = set(types)
      result = {t: (None, None, None) for t in wanted}
      for c in (cr or {}).get("status", {}).get("conditions", []):
          if c.get("type") in wanted:
              result[c["type"]] = (c.get("status"), c.get("reason"), c.get("message"))
      return result
```

yu-teo: Also, we did not plan for an early way out in `_wait_for_gateway_auth_enforced` when CR is not found. In cases like the AuthPolicy CR doesn't exist (e.g. wrong name/namespace), the function polls for the full **120s** before timing out. Should we early check-and-fail or at least log a warning (not just setting `last_snapshot`) on the first "not found" would save time during misconfiguration?

yu-teo:
Also, we have hard-coded the timeout to 120 seconds. Every `_wait_for_gateway_auth_enforced()` call hardcodes this timeout while we could have it configurable for different tests like a module-level constant alongside the other gateway propagation settings. What do you think? It will make it easy to tune. It is also consistent with the env-var pattern used for `GATEWAY_PROPAGATION_RETRIES`  and `GATEWAY_PROPAGATION_DELAY`.
Something like:
```
GATEWAY_ENFORCED_TIMEOUT = int(os.environ.get("E2E_GATEWAY_ENFORCED_TIMEOUT", "120"))
```

test/e2e/tests/test_helper.py
```diff
@@ -926,8 +1037,13 @@ def _wait_for_maas_auth_policy_phase(name, expected_phase="Active", namespace=No
...
 
             if phase == expected_phase:
-                # No auth policies required — phase match is sufficient
+                # No per-model auth policies required — phase match is sufficient for the CR,
+                # but gateway-only mode still needs Kuadrant Enforced before HTTP calls.
                 if not require_auth_policies:
+                    if require_enforced and expected_phase == "Active":
+                        # Keep a floor so a slow phase wait does not starve Kuadrant Enforced.
+                        remaining = max(60, int(deadline - time.time()))
+                        _wait_for_gateway_auth_enforced(timeout=remaining)
```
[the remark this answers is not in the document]
liangwen12year: This seems to be valid suggested changes.

ishitasequeira (approved):
lgtm

PR #1278, opened 2026-07-28
feat: serve controller metrics over HTTPS with authn/authz on OpenShift

sukumars321:
## Description
- Switch default metrics endpoint to :8443 and enable secure serving using controller-runtime's WithAuthenticationAndAuthorization filter.
- Mount OpenShift service-ca TLS secret via a new metrics Service and Deployment volume, and replace PodMonitor with ServiceMonitor using the service-ca CA bundle.
- Add kustomize replacements to derive ServiceMonitor tlsConfig.serverName from Service metadata.
- Keep xKS overlay on insecure HTTP :8080 by patching Deployment and Service.
- Update README defaults and add unit tests for metrics server options.

Addresses - https://redhat.atlassian.net/browse/RHOAIENG-76363.

## How Has This Been Tested?
Tested the changes with this custom build image - https://quay.io/repository/rh-ee-suksubra/maas-controller/manifest/sha256:cbe8111575e836c260152e6d9b54eb12279f5e86953bb7e50c2e19010aa84465 deloyed using deploy.sh in kustomize mode. and validated the changes with
[verify-maas-controller-secure-metrics.sh](https://github.com/user-attachments/files/30463577/verify-maas-controller-secure-metrics.sh)

## Merge criteria:

## New Features
- Metrics are now served securely over HTTPS by default on port **8443**, with authentication and authorization enabled.
- Added a dedicated metrics service and ServiceMonitor for HTTPS scraping.
- Added TLS certificate support and automatic service-name configuration.

## Compatibility
- The xKS overlay continues to support unsecured HTTP metrics on port **8080**.

## Bug Fixes
- Updated monitoring and network access rules for the HTTPS metrics endpoint.

## Documentation
- Updated metrics flag defaults and documented secure metrics behavior.


sukumars321:
@coderabbitai review

deployment/overlays/xks/kustomization.yaml
```diff
@@ -19,6 +19,23 @@ patches:
...
+        path: /metadata/annotations/service.beta.openshift.io~1serving-cert-secret-name
+      - op: replace
+        path: /spec/ports/0/name
+        value: metrics
+      - op: replace
+        path: /spec/ports/0/port
+        value: 8080
+      - op: replace
+        path: /spec/ports/0/targetPort
+        value: metrics
```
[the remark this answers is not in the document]
sukumars321: This is expected, we don't need to enable secure metrics serving for xks.

maas-controller/go.mod
```diff
@@ -80,13 +100,18 @@ require (
...
+	google.golang.org/grpc v1.80.0 // indirect
 	google.golang.org/protobuf v1.36.12-0.20260120151049-f2248ac996af // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/apiserver v0.35.3 // indirect
+	k8s.io/component-base v0.35.3 // indirect
 	k8s.io/klog/v2 v2.140.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20260427204847-8949caaa1199 // indirect
 	knative.dev/serving v0.48.1 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
```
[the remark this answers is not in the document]
sukumars321: @coderabbitai please elaborate the issue, I didn't introduce the replace line.

deployment/base/maas-controller/manager/manager.yaml
```diff
@@ -90,8 +92,8 @@ spec:
         name: manager
         imagePullPolicy: Always
         ports:
-        - containerPort: 8080
-          name: metrics
+        - containerPort: 8443
+          name: https
```
ahadas: do I get this right that you differentiate the port name since with insecure connectivity we'd still have the `maas-controller-metrics` service defined but it would then point to non-exposed port in maas-controller? if that's the case, I wonder if it wouldn't be better to keep the port name `metrics` in both cases, and avoid creating the service when we're deploying insecure connectivity for the metrics endpoint

deployment/base/maas-controller/default/kustomization.yaml
```diff
@@ -8,3 +8,32 @@ resources:
   - ../webhook
   # Gateway policies (gateway-default-auth, gateway-default-deny) are applied
   # separately by deploy.sh so they stay in openshift-ingress; see deploy_via_kustomize.
+
+# Derive ServiceMonitor tlsConfig.serverName as <service-name>.<namespace>.svc
```
ahadas: nice! I think I'll use this technique in other places :)

ahadas: ok, just tried that and I think it won't always work since the `opendatahub` namespace is replaced after kustomize build, in [postBuildTransform](https://github.com/opendatahub-io/models-as-a-service/blob/main/maas-controller/pkg/platform/tenantreconcile/kustomize.go#L63) - did you check this?

deployment/base/maas-controller/manager/manager.yaml
```diff
@@ -123,10 +125,19 @@ spec:
...
       - name: webhook-cert
         secret:
           defaultMode: 420
           secretName: maas-controller-webhook-cert
+      - name: metrics-certs
+        secret:
+          defaultMode: 420
+          secretName: maas-controller-metrics-tls
+          # optional until service-ca provisions the secret after the Service exists
+          optional: true
```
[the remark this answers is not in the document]
sukumars321: Removed `optional: true`.

deployment/base/maas-controller/default/kustomization.yaml
sukumars321: yeah I came across the issue when securing maas-api metrics #1293, Added https://github.com/opendatahub-io/models-as-a-service/pull/1293/changes#diff-28e1a9dd05f7f5a723cee8bad9c975c66d7b2ef465ef852e843d6d2bcd6c2e40R86 to address the namespace change.

deployment/overlays/xks/kustomization.yaml
```diff
@@ -28,11 +45,25 @@ patches:
...
+        value: --metrics-bind-address=:8080
+      - op: replace
+        path: /spec/template/spec/containers/0/args/3
+        value: --metrics-secure=false
+      - op: replace
+        path: /spec/template/spec/containers/0/ports/0/containerPort
+        value: 8080
+      - op: replace
+        path: /spec/template/spec/containers/0/ports/0/name
+        value: metrics
```
ishitasequeira: The comment says "Replace (not merge) base volumes/volumeMounts so OpenShift metrics-certs from manager.yaml are dropped." This means the xKS overlay silently swallows any new volume added to the base `manager.yaml` in the future.

For example, if someone later adds an `emptyDir` for temp files or a ConfigMap volume, it will just disappear on xKS with no build error. This is the kind of bug that only surfaces in production on a different platform.

At minimum, add a comment in the base `manager.yaml` near the `volumes:` block:          
```yaml                                                                                  
# NOTE: xKS overlay replaces the entire volumes/volumeMounts arrays.                     
# Any new volume here must also be added to deployment/overlays/xks/kustomization.yaml.  
```

ishitasequeira (commented):
Just one comment but overall the change looks good.

deployment/base/maas-controller/default/kustomization.yaml
ahadas: ok, so that needs to get in after #1293 then
and similar question to what I asked there - why do we need to replace the namespace also here?

deployment/overlays/xks/kustomization.yaml
ahadas: or can't we set that volume as optional as done for maas-api?

deployment/base/maas-controller/default/kustomization.yaml
sukumars321: `postBuildTransform` does not apply to the controller path — that remap is only used when rendering tenant/`maas-api` overlays via `tenantreconcile`. The controller ServiceMonitor is applied from `deployment/base/maas-controller/default` (deploy.sh / operator), so this change does not depend on #1293.

We still need the namespace replacement here: deploy.sh wraps default with `namespace: ${NAMESPACE}`, which can be `opendatahub` or `redhat-ods-applications`. Without pulling `Service.metadata.namespace` into `tlsConfig.serverName`, TLS hostname verification would keep the baked-in `opendatahub` label and fail on RHOAI.

deployment/overlays/xks/kustomization.yaml
sukumars321: Good call — went with the optional-secret approach instead of replacing the whole volume arrays.

xKS now keeps the base `webhook-cert` / `metrics-certs` mounts and only sets `metrics-certs.secret.optional: true` (with `--metrics-secure=false`). OpenShift base still requires the metrics TLS secret. That way future volumes added to `manager.yaml` are not silently dropped on xKS.

deployment/base/maas-controller/default/kustomization.yaml
ahadas: > `postBuildTransform` does not apply to the controller path — that remap is only used when rendering tenant/`maas-api` overlays via `tenantreconcile`. The controller ServiceMonitor is applied from `deployment/base/maas-controller/default` (deploy.sh / operator), so this change does not depend on #1293.

yeah, makes sense

> We still need the namespace replacement here: deploy.sh wraps default with `namespace: ${NAMESPACE}`, which can be `opendatahub` or `redhat-ods-applications`. Without pulling `Service.metadata.namespace` into `tlsConfig.serverName`, TLS hostname verification would keep the baked-in `opendatahub` label and fail on RHOAI.

yeah well, I doubt someone would actually use `deploy.sh` to deploy MaaS on RHOAI (as opposed to using the operator) but it is currently possible and changing that is out of scope for this PR - how about adding a comment to reflect it's needed only for this flow?

liangwen12year (commented):
## xKS overlay: hardcoded args indices are fragile

The xKS overlay patches args by positional index:

```yaml
- op: replace
  path: /spec/template/spec/containers/0/args/2
  value: --metrics-bind-address=:8080
- op: replace
  path: /spec/template/spec/containers/0/args/3
  value: --metrics-secure=false
```

If anyone adds, removes, or reorders args before position 2 in `deployment/base/maas-controller/manager/manager.yaml`, these silently overwrite the wrong flags at deploy time — hard to diagnose.

Same issue with `path: /spec/template/spec/volumes/1/secret/optional` — depends on `metrics-certs` staying at volume index 1.

**Suggestion:** Add a comment in the xKS kustomization mapping the indices to the base args they target, so the next person who touches the base args list knows to update the overlay:

```yaml
  # Indices match base manager.yaml args ordering:
  #   0: --leader-elect
  #   1: --health-probe-bind-address=:8081
  #   2: --metrics-bind-address=:8443  → patched to :8080
  #   3: --metrics-secure=true         → patched to false
```

This doesn't eliminate the fragility (kustomize limitation), but makes it visible.

liangwen12year (approved):
lgtm

ishitasequeira (commented):
Overall code looks good to me. @ahadas do you have any open concerns?

deployment/overlays/xks/kustomization.yaml
```diff
@@ -27,21 +39,27 @@ patches:
...
+      - op: replace
+        path: /spec/template/spec/containers/0/ports/0/containerPort
+        value: 8080
       - op: add
-        path: /spec/template/spec/volumes
-        value:
-          - name: webhook-certs
-            secret:
-              secretName: maas-controller-webhook-cert
-      - op: add
```
ahadas, line 42: I wouldn't drop those webhook related thingies as part of this PR but if you're confident with these changes then ok

ahadas (approved):

deployment/base/maas-controller/manager/manager.yaml
sukumars321: keeping the port name as metrics in both cases.

deployment/overlays/xks/kustomization.yaml
sukumars321: the webhook mount was always in base; the xKS overlay only re-declared it under a different volume name when it replaced the whole volume array for metrics reasons; the new approach keeps base webhook-cert and only patches metrics-specific settings.

ishitasequeira (approved):
/lgtm

PR #1284, opened 2026-07-28
fix(RHOAIENG-79535): disable IPP response api-translation by default

jland-redhat:
## Summary
- Turn off response-path `api-translation` in `payload-processing-plugins` so SSE streaming works for internal / OpenAI-compatible models by default (request-path translation stays enabled).
- After bootstrap/migrate, stamp `opendatahub.io/managed=false` on the plugins ConfigMap so the tenant reconciler leaves operator edits alone; set `managed=true` (or remove the annotation) to opt back into reconciler management / reset defaults.
- Document how to re-enable response translation for providers that need it (e.g. Anthropic).

Fixes: https://redhat.atlassian.net/browse/RHOAIENG-79535

## Test plan
- [ ] Fresh / upgraded deploy: streaming (`"stream": true`) works for an internal model without manual ConfigMap edits
- [ ] `payload-processing-plugins` ends up with `opendatahub.io/managed=false` after reconcile
- [ ] Editing the ConfigMap to add response `api-translation` + restarting `payload-processing` sticks across reconciles
- [ ] Removing `opendatahub.io/managed` triggers a one-shot reset to product defaults
- [ ] Existing external-model e2e (`provider: openai`) still passes

## Risk analysis
- **Risk rating**: 3
- **Why**: Changes default IPP response plugin behavior and reconciler ownership of the plugins ConfigMap. Blast radius is gateway streaming / external-provider response rewrite; Konflux smoke covers OpenAI external models (response translator already a no-op) but not Anthropic response translation. Operators who need response rewrite must follow the opt-in docs.

## Summary by CodeRabbit

- **New Features**
  - Response-side API translation is now disabled by default to preserve SSE streaming.
  - Optional response translation can be enabled for providers that rewrite between different API formats.

- **Documentation**
  - Added “IPP response translation (opt-in)” setup steps, including restart guidance.
  - Documented how to reset payload-processing defaults and how controller management ownership is controlled.

- **Bug Fixes**
  - Improved reconciliation of payload-processing plugin configuration to preserve intentional management settings and user edits.

### DIFF
--- a/deployment/base/payload-processing/manager/plugins-configmap.yaml
+++ b/deployment/base/payload-processing/manager/plugins-configmap.yaml
@@ -3,6 +3,11 @@ kind: ConfigMap
 metadata:
   name: payload-processing-plugins
   namespace: openshift-ingress
+  # Do NOT set opendatahub.io/managed=false here: PostRender drops any resource that
+  # already carries that annotation, which would prevent first-time creation.
+  # The tenant reconciler stamps managed=false after bootstrap/migrate so later
+  # SSA leaves this ConfigMap alone (see ApplyRendered). Set managed=true on the
+  # live ConfigMap to opt back into continuous reconciler management.
 data:
   custom-pre-processing-ipp-config.yaml: |
     apiVersion: llm-d.ai/v1alpha1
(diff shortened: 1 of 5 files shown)


ishitasequeira (approved):
lgtm

PR #1287, opened 2026-07-29
fix: advertise gateway base url for externalmodel catalog

jland-redhat:
## Summary
- Make ExternalModel `status.endpoint` the shared gateway base URL (`https://{host}`), matching LLMInferenceService BBR catalog entries.
- Have GET `/v1/models` derive ExternalModel URLs from `status.httpRouteHostnames[0]` the same way as internal models.
- Leave path-based HTTPRoute rules in place for backward-compatible clients.

## Test plan
- [x] Unit tests for `GetModelEndpoint` / ExternalModel Status updated and passing
- [ ] Deploy controller; ExternalModel MaaSModelRef `status.endpoint` is `https://{gateway}` (no `/ns/name`)
- [ ] GET `/v1/models` returns the same base URL for ExternalModel and LLMInferenceService entries
- [ ] Existing path-based ExternalModel inference (`/{ns}/{name}/v1/chat/completions`) still works

## Risk analysis
- **Risk rating**: 2
- **Why**: Catalog and status URL shape change only; path HTTPRoute rules and auth identity are unchanged. Clients that treated the catalog URL as a path-prefixed base may need to send `body.model` on the unified gateway path. Covered by unit tests; smoke still exercises path-built ExternalModel URLs in e2e.

### DIFF
--- a/maas-api/internal/models/maasmodelref.go
+++ b/maas-api/internal/models/maasmodelref.go
@@ -57,9 +57,10 @@ func GVR() schema.GroupVersionResource {
 // maasModelRefToModel converts a MaaSModelRef unstructured to a Model for the API.
 //
 // For LLMInferenceService-backed models (BBR clusters), the model ID is read from
-// status.resolvedModelAlias (the canonical publishers/{ns}/models/{name} form), and
-// the URL is derived from status.httpRouteHostnames[0] (the shared gateway base URL).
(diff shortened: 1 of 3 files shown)


jland-redhat:
@coderabbitai review

ishitasequeira (approved):
lgtm

Tested the change locally on cluster

PR #1288, opened 2026-07-29
fix(RHOAIENG-79568): always verify gateway AuthPolicy enforcement bef…

jrhyness:
…ore proceeding

The MaaSAuthPolicy controller previously only checked gateway AuthPolicy enforcement status during legacy 3.4-to-3.5 upgrades. This left a transient window after any gateway AuthPolicy spec change where Kuadrant had not yet re-enforced the updated policy, causing Authorino to stop injecting X-MaaS-Username/X-MaaS-Group headers and API key creation to fail with AUTH_FAILURE.

Remove the legacy-only gate and check enforcement on every reconcile. The check is cheap (single GET + condition check) and returns immediately when the policy is already enforced.

- Change reconcileGatewayAuthPolicy return type to (bool, error)
- Always call gatewayAuthPolicyReady after reconciling the gateway AuthPolicy
- Add unit test for requeue-until-enforced behavior
- Add e2e test that scales Kuadrant down to deterministically reproduce the enforcement gap and verify the controller holds Pending
- Fix 9 existing tests that were missing proper gateway setup (previously passed only because the enforcement check was gated behind legacy flag)

For https://redhat.atlassian.net/browse/RHOAIENG-79568

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MaaSAuthPolicy reconciliation to keep policies **Pending** until the gateway-scoped AuthPolicy is **Accepted** and **Enforced**.
  * Avoids unnecessary reconciliations when the gateway-scoped AuthPolicy already matches the desired enforced state.
  * Refined gateway-related status messaging during legacy cutover and reduced spurious gateway spec updates.

* **Tests**
  * Updated unit tests to include gateway “ready” AuthPolicy prerequisites and explicit gateway targeting.
  * Added unit tests for requeue-on-enforcement and no-requeue-when-unchanged.
  * Added e2e coverage for enforcement-gap behavior and adjusted multi-tenant/discovery flows and step sequencing.

### DIFF
--- a/maas-controller/pkg/controller/maas/conflict_detection_test.go
+++ b/maas-controller/pkg/controller/maas/conflict_detection_test.go
@@ -120,6 +120,7 @@ func TestDetectConflictingAuthPolicies_RogueDetected(t *testing.T) {
 		httpRouteName  = "maas-" + modelName
 		maasPolicyName = "policy-a"
 		rogueName      = "kserve-route-authn"
+		gatewayNS      = "openshift-ingress"
 	)
 
 	model := newMaaSModelRef(modelName, namespace, "ExternalModel", modelName)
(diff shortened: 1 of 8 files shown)


ishitasequeira (dismissed):
lgtm

jland-redhat (dismissed):
lgtm

ishitasequeira (approved):
lgtm

ishitasequeira (approved):
lgtm

PR #1292, opened 2026-07-29
fix: close default tenant bootstrap race

dmytro-zaharnytskyi:
## Description

Fixes [RHOAIENG-79720](https://redhat.atlassian.net/browse/RHOAIENG-79720).

The default MaaS tenant bootstrap previously waited for the full `AITenant` gateway validation path before creating `MaasTenantConfig/default-tenant`. During that window, the UI could attempt to create a `MaaSAuthPolicy` or `MaaSSubscription` and be rejected by the tenant-namespace admission webhook.

This change:

- reduces the default `AITenant` bootstrap retry interval from 30 seconds to 2 seconds;
- creates or adopts the default tenant namespace and `MaasTenantConfig` before gateway readiness, with the expected AITenant ownership metadata;
- retries after one second when the API server temporarily reports that a newly created namespace is not yet available for namespaced resources;
- preserves gateway-first provisioning for non-default AITenants.

This PR does not make the default tenant mandatory. The existing bootstrap annotation and intentional default-tenant deletion behavior are unchanged: deleting the default `AITenant` still removes its tenant configuration without automatically recreating it.

## How Has This Been Tested?

Run from `maas-controller/`:

```bash
go test ./...
go test -race ./pkg/controller/maas ./cmd/manager
./bin/tools/golangci-lint run --new-from-rev=HEAD
```

The tests cover:

- creation of the default `MaasTenantConfig` before gateway readiness;
- preservation of gateway-first behavior for custom AITenants;
- retry and recovery after a transient namespace-not-found response;
- existing default-tenant deletion behavior.

No manual OpenShift cluster testing has been performed yet.

## Risk analysis

- **Risk rating:** 2
- **Why:** This changes controller bootstrap ordering and retry timing, but the behavior is limited to default-tenant initialization and transient namespace handling. Unit and race tests cover the changed paths, while the Prow smoke flow exercises full deployment plus subscription and authorization-policy creation. There are no API, CRD, RBAC, or manifest changes.

## Merge criteria:

[RHOAIENG-79720]: https://redhat.atlassian.net/browse/RHOAIENG-79720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Default tenant resources can now be created before the gateway is fully ready.
  * Tenant setup reports a distinct pending status when the namespace is not yet available.
* **Bug Fixes**
  * Improved tenant provisioning retries when namespaces or configuration resources are temporarily unavailable.
  * Tenant configuration readiness now verifies that the latest configuration was observed.
  * Cluster bootstrap retries more frequently, reducing delays when creating the default tenant.
  * Custom tenants continue waiting for gateway readiness before resource creation.

### DIFF
--- a/maas-controller/cmd/manager/main.go
+++ b/maas-controller/cmd/manager/main.go
@@ -638,6 +638,10 @@ func markDefaultAITenantBootstrapped(ctx context.Context, c client.Client, ct *m
 	return nil
 }
 
+// clusterBootstrapPollInterval is how often ensureClusterBootstrapRunnable retries
+// creating the default AITenant after Config/default exists.
+const clusterBootstrapPollInterval = 2 * time.Second
+
 // ensureClusterBootstrapRunnable bootstraps the default AITenant once
 // Config/default is present (created by LifecycleReconciler). It does not set
 // owner references; LifecycleReconciler links Config to the default AITenant,
@@ -648,7 +652,7 @@ func ensureClusterBootstrapRunnable(mgr ctrl.Manager, tenantNamespace, aitenantN
 		log := ctrl.Log.WithName("setup").WithName("ensureClusterBootstrap")
 		c := mgr.GetClient()
 
-		ticker := time.NewTicker(30 * time.Second)
+		ticker := time.NewTicker(clusterBootstrapPollInterval)
 		defer ticker.Stop()
 
 		ensure := func() {
--- a/maas-controller/cmd/manager/main.go
+++ b/maas-controller/cmd/manager/main.go
@@ -638,10 +638,6 @@ func markDefaultAITenantBootstrapped(ctx context.Context, c client.Client, ct *m
 	return nil
 }
 
-// clusterBootstrapPollInterval is how often ensureClusterBootstrapRunnable retries
-// creating the default AITenant after Config/default exists.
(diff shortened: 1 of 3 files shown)


maas-controller/cmd/manager/main.go
```diff
@@ -638,6 +638,10 @@ func markDefaultAITenantBootstrapped(ctx context.Context, c client.Client, ct *m
 	return nil
 }
 
+// clusterBootstrapPollInterval is how often ensureClusterBootstrapRunnable retries
+// creating the default AITenant after Config/default exists.
+const clusterBootstrapPollInterval = 2 * time.Second
```
jrhyness: _Is_ it only after Config/default exists, and does it continue indefinitely?    2 seconds might be too often, or maybe okay if it stops after successful bootstrap?

dmytro-zaharnytskyi: fair enough, removed it (the retry remains in the AITenant reconciler)

ishitasequeira (approved):
/lgtm

PR #1293, opened 2026-07-29
feat: add secure metrics serving and ServiceMonitor for maas-api

sukumars321:
## Description

Serve maas-api /metrics over HTTPS with Kubernetes token authentication and SAR authorization when METRICS_SECURE=true (OpenShift default). Mount TLS certificates provisioned by OpenShift service-ca and expose them through a dedicated maas-api-metrics Service. Replace the PodMonitor with a ServiceMonitor configured for HTTPS scraping. Add a dev overlay patch to disable secure metrics for local development. Update observability docs and add controller-side remapping of the ServiceMonitor tlsConfig.serverName for tenant namespaces.

## How Has This Been Tested?
Tested the code in a live cluster using https://quay.io/repository/rh-ee-suksubra/maas-api/manifest/sha256:e3833ebd3483dd9fea8adf0856592bf0cef2ba52a21f69709afb435850f387e7 and https://quay.io/repository/rh-ee-suksubra/maas-controller/manifest/sha256:f51e445f29f3c938bdd5a4ce76aeaa75c69d70d17982afb8b47a02e2f45bc6f6

## Merge criteria:

Jira: https://redhat.atlassian.net/browse/RHOAIENG-76363

## Summary by CodeRabbit

- **New Features**
  - Added HTTPS-protected metrics on port 9090 with authentication and authorization.
  - Added a dedicated metrics service and ServiceMonitor for `/metrics`.
  - Added TLS support for metrics in secure deployments, with insecure mode available for development.
  - Added automatic service certificate provisioning and configuration options.
  - Added support for request rates, latency, and key validation metrics in monitoring dashboards.

- **Documentation**
  - Updated observability guidance with metrics availability, dashboards, and scraping requirements.

- **Bug Fixes**
  - Corrected cross-namespace metrics monitoring by generating appropriate service DNS names.


sukumars321:
@ahadas Can you also review this PR as well.

deployment/base/maas-api/core/metrics_service.yaml
```diff
@@ -0,0 +1,22 @@
+# Dedicated metrics Service for SecureServing.
+# OpenShift service-ca provisions maas-api-metrics-tls from the annotation.
+# Separate from the main maas-api Service (which already has API serving certs).
+apiVersion: v1
+kind: Service
```
ahadas, line 5: so with this we'll have two endpoints to poll the same metrics, `maas-api-metrics.ns.svc:https` and `maas-api.ns.svc:metrics`, right?
can't we set the `service.beta.openshift.io/serving-cert-secret-name` annotation on the `maas-api` service instead, when deploying it on non-dev environments?

deployment/base/maas-api/default/kustomization.yaml
```diff
@@ -16,3 +16,32 @@ labels:
...
+  - source:
+      kind: Service
+      name: maas-api-metrics
+      fieldPath: metadata.namespace
+    targets:
+      - select:
+          kind: ServiceMonitor
+          name: maas-api-metrics
+        fieldPaths:
+          - spec.endpoints.0.tlsConfig.serverName
```
ahadas: do I get this correctly that this meant for deployments of maas-api that are not done through maas-controller (since then we override the namespace after the kustomize build)? if so, I wonder if there really is such a case

deployment/base/maas-api/core/deployment.yaml
```diff
@@ -31,6 +31,10 @@ spec:
           value: "models-as-a-service"
         - name: METRICS_PORT
           value: "9090"
+        - name: METRICS_SECURE
+          value: "false"
+        - name: METRICS_CERT_DIR
```
ahadas: I think we can drop this since it's set when `METRICS_SECURE` is turned on, right?

deployment/base/maas-api/core/metrics_service.yaml
sukumars321: Good catch. We can't put the metrics serving-cert annotation on the main `maas-api` Service: TLS already uses `service.beta.openshift.io/serving-cert-secret-name: maas-api-serving-cert` for the API on `:8443`, and OpenShift service-ca allows only one serving-cert secret per Service. Metrics need a separate cert (`maas-api-metrics-tls`) with SANs for `maas-api-metrics.<ns>.svc`.

On the TLS/ODH path, `overlays/tls/service-patch.yaml` already replaces the main Service ports to only `https:8443`, so Prometheus scrapes only via `maas-api-metrics`. I've also dropped the redundant `metrics:9090` port from base `service.yaml` so even non-TLS builds don't expose metrics on both Services.

deployment/base/maas-api/default/kustomization.yaml
sukumars321: Partly — the controller path also uses these replacements.

`maas-api/deploy/overlays/odh` includes TLS → `default/`, then sets `namespace: opendatahub` and re-runs the same replacements. After that, Go `RenderKustomize` remaps again when the live infra namespace differs (e.g. `odh-ai-gateway-infra`).

Keeping the `default/` replacements also keeps standalone `kustomize build` of `default/` / TLS overlays correct without going through the controller. So I'd keep them here.

deployment/base/maas-api/core/deployment.yaml
sukumars321: Agreed for base. Go already defaults `METRICS_CERT_DIR` to `/tmp/k8s-metrics-server/metrics-certs`, and base has `METRICS_SECURE=false`, so the env var was redundant there.

Removed it from base `deployment.yaml`. Left it explicit in the TLS overlay next to the cert volume mount.

maas-api/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.25.8
 require (
 	github.com/gin-contrib/cors v1.7.6
 	github.com/gin-gonic/gin v1.10.1
+	github.com/go-logr/logr v1.4.3
```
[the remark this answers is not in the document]
sukumars321: Acknowledged. The `controller-runtime v0.22.1` vs `k8s.io/* v0.35.1` skew was already present on main (CRT was an indirect dependency). This PR only promotes CRT to a direct require for `filters.WithAuthenticationAndAuthorization`.

I'd prefer to leave versions as-is here and handle a CRT `v0.23.x` bump in a dedicated dependency PR — maas-controller already shows kserve/`replace` complexity around CRT upgrades.

maas-controller/pkg/platform/tenantreconcile/kustomize.go
```diff
@@ -144,6 +151,64 @@ func remapSubjectNamespaces(res *resource.Resource, appNamespace string) error {
...
+		}
+		serverName, ok := tlsCfg["serverName"].(string)
+		if !ok || serverName == "" {
+			continue
+		}
+		rewritten := replaceHostNamespace(serverName, appNamespace)
+		// Only rewrite when the host was using the overlay default namespace.
+		if rewritten != serverName && strings.Contains(serverName, "."+overlayDefaultNamespace+".") {
+			tlsCfg["serverName"] = rewritten
+			changed = true
```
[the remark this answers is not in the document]
sukumars321: Fixed. The rewrite guard now requires the 2nd DNS label to be the overlay default namespace (`opendatahub`), instead of a substring `Contains` check.

Added unit coverage so hosts like `prometheus.monitoring.opendatahub.svc` are left unchanged, while `maas-api-metrics.opendatahub.svc` still remaps correctly.

sukumars321:
@ishitasequeira can you take a look at this as well.

deployment/base/maas-api/default/kustomization.yaml
ahadas: ok, still not sure it's actually needed but we can always drop that later on if we find it redundant

deployment/base/maas-api/monitoring/servicemonitor.yaml
```diff
@@ -11,7 +13,9 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: maas-api
-  podMetricsEndpoints:
-  - port: metrics
-    path: /metrics
-    interval: 30s
+      maas.opendatahub.io/metrics: "true"
```
ahadas, line 16: this needs to be added to the deployment of maas-api, no?

ahadas (commented):
overall lgtm, asked about where the `maas.opendatahub.io/metrics: "true"` label is set - if monitoring of maas-api works with these changes then I probably missed that

deployment/base/maas-api/core/deployment.yaml
```diff
@@ -31,6 +31,9 @@ spec:
           value: "models-as-a-service"
         - name: METRICS_PORT
           value: "9090"
+        - name: METRICS_SECURE
+          value: "false"
+        # METRICS_CERT_DIR defaults in Go (DefaultMetricsCertDir); set explicitly in overlays/tls.
```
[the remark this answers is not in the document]
sukumars321: already aligned; base is http, TLS overlay patches https.

deployment/base/maas-api/monitoring/servicemonitor.yaml
sukumars321: Not really the annotation actually needs to be in the service, which already has it.

ishitasequeira (commented):
Overall aligned with the changes. @sukumars321, can you help validate or confirm below observations?

This PR replaces PodMonitor with ServiceMonitor in the base, but the xKS overlay still deletes a PodMonitor. That delete is now a no-op; `kustomize build maas-api/deploy/overlays/xks` still succeeds, but it renders OCP-only resources that break non-OCP tenant deploys.

On xKS there is no service-ca, and certificate.yaml only provisions `maas-api-serving-cert`,  not `maas-api-metrics-tls`. The pod will not start waiting for that Secret, and secure metrics init would fail without certs anyway.

Should we mirror what #[1278](https://github.com/opendatahub-io/models-as-a-service/pull/1278) did for controller metrics in `deployment/overlays/xks/kustomization.yaml`:
1. Delete Service/maas-api-metrics
2. Delete ServiceMonitor/maas-api-metrics
3. Patch Deployment/maas-api: set `METRICS_SECURE=false` and make the metrics cert volume `optional: true`

maas-api/deploy/overlays/xks/kustomization.yaml
```diff
@@ -70,3 +82,30 @@ patches:
...
+              env:
+              - name: METRICS_SECURE
+                value: "false"
+  # Make metrics-certs optional (index 2 after overlays/tls volume list).
+  - target:
+      kind: Deployment
+      name: maas-api
+    patch: |
+      - op: replace
+        path: /spec/template/spec/volumes/2/secret/optional
```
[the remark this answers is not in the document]
sukumars321: Not needed. metrics is not at all supported in xks.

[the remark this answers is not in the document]
sukumars321: Correct but insecure metrics serving is the default behavior before this change on all platforms and addressing this issue is out of scope for this PR.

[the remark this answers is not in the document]
sukumars321: No thanks. I will create a separate issue.

sukumars321:
@ishitasequeira Addressed your concern, PTAL.

ishitasequeira (dismissed):
Overall PR looks good!!

ishitasequeira (dismissed):
lgtm

ishitasequeira (approved):
lgtm

PR #1298, opened 2026-07-30
docs: center inference examples on body-based routing endpoints

EgorLu:
## Summary

Validates and updates upstream Body-Based Routing documentation per RHOAIENG-79713. All primary inference examples now use the OpenAI-compatible body-based endpoint (`POST /v1/chat/completions` with model in the request body) instead of path-based per-model URLs.

## Changes across 7 files

- **inference.md**: Restructured to lead with body-based endpoint as the primary teaching path. Moved path-based routing to a dedicated "Path-Based Routing (Legacy)" section with a comparison table. All examples (basic, streaming, multi-turn, rate limit retry) now use `${MAAS_API_URL}/v1/chat/completions`.
- **validation.md**: Test steps (inference, auth enforcement, rate limiting) use body-based endpoint as primary, with a note that path-based still works.
- **model-setup.md**: Verification section uses body-based endpoint with a legacy note.
- **model-discovery.md**: Clarified `url` field as path-based (legacy). Added a body-based inference example in "Using Model Information". Renamed "Get the Model URL" to "Get the Path-Based Model URL (Legacy)".
- **external-model-setup.md**: Validation curl commands (Run Inference, Verify Auth Enforcement) use body-based endpoint.
- **tier-to-subscription.md**, **upgrade-to-3.4.md**: Added informational admonition notes explaining that path-based URLs are used for migration context continuity, with a link to the updated Inference guide.

All files include a note that legacy path-based endpoints still work, linking to the Path-Based Routing (Legacy) section in inference.md.

## What is NOT changed

- No code changes, no API changes
- Migration docs keep their path-based examples (contextually appropriate) with a note
- The `/maas-api/v1/models` response format is unchanged (`url` field still returns path-based URLs)
- No new doc files created (updates only, per doc policy)

## Risk analysis

- **Risk rating**: 0
- **Why**: Documentation-only changes with no runtime impact. All changes are to markdown files under `docs/content/`. No Go code, scripts, manifests, or CI logic is affected.

## Summary by CodeRabbit

* **Documentation**
  * Updated inference guidance to recommend a unified OpenAI-compatible endpoint with the model specified in the request body.
  * Refreshed setup, validation, migration, and upgrade examples to use body-based routing.
  * Clarified path-based routing as a supported legacy option.
  * Updated model discovery guidance and examples to distinguish current and legacy endpoint formats.
  * Revised authentication, authorization, rate-limit, streaming, retry, and multi-turn examples accordingly.
  * Added guidance on selecting ready models and using their identifiers for requests.

### DIFF
--- a/docs/content/install/external-model-setup.md
+++ b/docs/content/install/external-model-setup.md
@@ -287,7 +287,7 @@ curl -sS "https://${GW_HOST}/v1/chat/completions" \
 # Bogus key - expect 403
 curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/v1/chat/completions" \
   -H "Content-Type: application/json" \
-  -H "Authorization: Bearer bogus-key" \
+  -H "Authorization: Bearer sk-oai-FAKE-KEY" \
   -d '{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}]}'
 
 # No auth - expect 401
--- a/docs/content/install/external-model-setup.md
+++ b/docs/content/install/external-model-setup.md
@@ -272,23 +272,26 @@ echo "MaaS API key: ${KEY:0:20}..."
 ### Run Inference
 
 ```bash
-curl -sS "https://${GW_HOST}/llm/gpt-4o/v1/chat/completions" \
+curl -sS "https://${GW_HOST}/v1/chat/completions" \
   -H "Content-Type: application/json" \
-  -H "Authorization: Bearer $KEY" \
+  -H "Authorization: Bearer ${KEY}" \
   -d '{"model":"gpt-4o","messages":[{"role":"user","content":"say hello"}]}'
 ```
+
+!!! note "Legacy path-based endpoint"
+    You can also use the path-based URL: `https://${GW_HOST}/llm/gpt-4o/v1/chat/completions`. Both routing modes are supported. See [Inference - Path-Based Routing](../user-guide/inference.md#path-based-routing-legacy) for details.
 
 ### Verify Auth Enforcement
 
 ```bash
-# Bogus key — expect 403
-curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/llm/gpt-4o/v1/chat/completions" \
+# Bogus key - expect 403
+curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/v1/chat/completions" \
   -H "Content-Type: application/json" \
-  -H "Authorization: Bearer sk-oai-FAKE-KEY" \
+  -H "Authorization: Bearer bogus-key" \
   -d '{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}]}'
 
-# No auth — expect 401
-curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/llm/gpt-4o/v1/chat/completions" \
+# No auth - expect 401
+curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/v1/chat/completions" \
   -H "Content-Type: application/json" \
   -d '{"model":"gpt-4o","messages":[{"role":"user","content":"hi"}]}'
 ```
--- a/docs/content/install/validation.md
+++ b/docs/content/install/validation.md
@@ -70,11 +70,10 @@ Each API key is bound to one MaaSSubscription at creation time. `GET /v1/models`
 ```bash
 MODELS=$(curl -sS ${HOST}/maas-api/v1/models \
     -H "Content-Type: application/json" \
-    -H "Authorization: Bearer $API_KEY" | jq -r .) && \
+    -H "Authorization: Bearer ${API_KEY}" | jq -r .) && \
 echo $MODELS | jq . && \
-MODEL_NAME=$(echo $MODELS | jq -r '.data[0].id') && \
-MODEL_URL=$(echo $MODELS | jq -r '.data[0].url') && \
-echo "Model URL: $MODEL_URL"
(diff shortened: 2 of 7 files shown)


EgorLu:
@coderabbitai review

EgorLu:
Still testing on my cluster

docs/content/install/external-model-setup.md
```diff
@@ -272,23 +272,26 @@ echo "MaaS API key: ${KEY:0:20}..."
...
 ### Verify Auth Enforcement
 
 ```bash
-# Bogus key — expect 403
-curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/llm/gpt-4o/v1/chat/completions" \
+# Bogus key - expect 403
+curl -sS -w "HTTP: %{http_code}\n" "https://${GW_HOST}/v1/chat/completions" \
   -H "Content-Type: application/json" \
-  -H "Authorization: Bearer sk-oai-FAKE-KEY" \
+  -H "Authorization: Bearer bogus-key" \
```
jland-redhat: We have some logic around `sk-oai` prefix. So wondering if we want toe keep that here. Also small nit but maybe we stick with FAKE over bogus

( Think the outcome is the same so maybe this is fine)

docs/content/install/validation.md
```diff
@@ -79,23 +79,26 @@ echo "Model URL: $MODEL_URL"
...
-curl -sS -H "Authorization: Bearer $API_KEY" \
+curl -sS -H "Authorization: Bearer ${API_KEY}" \
   -H "Content-Type: application/json" \
   -d "{\"model\": \"${MODEL_NAME}\", \"messages\": [{\"role\": \"user\", \"content\": \"Hello\"}], \"max_tokens\": 50}" \
-  "${MODEL_URL}/v1/chat/completions" | jq
+  "${HOST}/v1/chat/completions" | jq
 ```
 
+!!! note "Legacy path-based endpoint"
+    You can also use the per-model URL from step 3: `${MODEL_URL}/v1/chat/completions`. Both path-based and body-based endpoints are supported. See [Inference - Path-Based Routing](../user-guide/inference.md#path-based-routing-legacy) for details.
```
jland-redhat: So the "MODEL_URL" is the BBR model now, so this is effectively the same as ${HOST}/v1/chat/completions".

To that point there is not real point in capturing the MODEL_URL anymore. But we should still note the legacy path.

jland-redhat (changes requested):

ishitasequeira (changes requested):
A few things mostly from cluster testing and doc accuracy:

- **IPP**: The PR drops the IPP prerequisite, but BBR really does depend on it. Without ext_proc (`model` → `X-Gateway-Model-Name`), pure `POST …/v1/chat/completions` 404s. Can we put the admin note back and restore the "Requires IPP" row in the comparison table?

- **Model ID in the body**: For LLMIS models, clients need the publisher ID from `/v1/models` (e.g. `publishers/llm/models/...`), not the MaaSModelRef name or a short name like `facebook/opt-125m`. Worth calling that out explicitly in inference.md and fixing the examples in model-discovery.md that still show `llama-2-7b-chat`. External models are the exception (`gpt-4o` etc.), a quick note so people don't over-apply the publisher-ID rule would help.

- **jq examples**: CodeRabbit's right. `.data[0]` can grab a Pending model. Filtering with `select(.ready==true)` (same pattern you already have in model-discovery) would make the validation flow less flaky.

- **Multi-tenant**: One sentence in inference.md would go a long way.. non-default tenants hit `{tenant-gateway}/v1/chat/completions` with a tenant API key, not the default `maas.*` URL. Link to multi-tenant validation and we're good.

- validation.md step 3: Since steps 4/6/7 now use `${HOST}/v1/chat/completions`, we can drop `MODEL_URL` from step 3 and leave path-based URL discovery in the legacy note at step 4 (Jamie's point).

- external-model-setup.md — Small one: please revert bogus-key back to `sk-oai-FAKE-KEY`, same as Jamie's suggestion

- **Not covered in this PR but may need changes**: 
  - model-listing-flow.md will have a broken anchor after the section rename and still says `id` is `metadata.name`; 
  - maas-model-ref.md is missing tenantRef / resolvedModelAlias 
  - multi-tenant validation has no BBR inference step
  - validation scripts are still path-based. We can separate this into a different PR

docs/content/install/external-model-setup.md
EgorLu: That's an interesting find / side-effect of Opus 5 over 4.6.
I vividly remember writing "bogus placeholder value" and it took my words literally.

And now on the point, fixed back to `sk-aoi-`.

EgorLu:
All review feedback addressed in commits 809fccf6 and 604a9efc:

- Reverted bogus-key to `sk-oai-FAKE-KEY` (jland)
- Dropped `MODEL_URL` from validation step 3 (jland)
- Restored IPP prerequisite note + `Requires IPP` row in comparison table (ishitasequeira)
- Added publisher ID model format tip for LLMIS vs external models (ishitasequeira)
- Added multi-tenant note for non-default gateway URLs (ishitasequeira)
- All jq filters use `select(.ready==true)` with `-e` flag for fail-fast (CodeRabbit)
- Fixed fenced code block language (CodeRabbit)

**Cluster-validated** on `bbr-test` ROSA HCP cluster: both path-based (`/llm/simulator/v1/chat/completions`) and body-based (`/v1/chat/completions`) routing return 200 OK with valid chat completion responses.

Ready for re-review.

ishitasequeira (approved):
lgtm

Thanks for addressing all the feedback!!

jland-redhat (approved):
lgtm

PR #1301, opened 2026-07-30
feat: add failed requests column and filter auth-failure rows from loki dashboards

tgitelman:
## Summary

Add a **Failed requests** column to the usage breakdown table on both Loki dashboards and filter out auth-failure noise (401/403 with unresolved identity) from dropdowns and table.

### Failed requests column (Q4)

Both dashboards gain a fourth table query counting `response_type="error"` entries. The column shows how many requests failed per user/model/subscription combination — errors that occurred **after** successful authentication (e.g. 404 model-not-found, 500 backend error).

### Auth-failure filtering

When authentication fails (401 invalid token, 403 subscription mismatch), the WASM shim is atomic — **no** identity fields are written to FILTER_STATE. These entries have `user_id="-"` and `subscription="-"` with garbage model names (client-provided, never routed).

| Where | Filter | Mechanism |
|-------|--------|-----------|
| Model dropdown (both) | `subscription!="-"` added to matcher | Stream label (proxy for unresolved identity) |
| Admin table Q1–Q4 | `\| user_id!="-"` | Structured metadata pipeline filter |
| User table Q1–Q4 | `subscription!="-"` in stream selector | Stream label (no `user_id` column — proxy handles user scoping) |

Subscription dropdown already had `subscription!="-"` — no change needed.

### Cross-zero-padding

All four table queries use `or (other_query * 0)` to ensure every label combo appears in every query with value `0` instead of empty. This is required for `MergeSeries` to join correctly — without it, error-only combinations produce empty cells that sort before numbers in descending order.

| Query | Zero-padded with |
|-------|------------------|
| Q1 (Tokens used) | `or (error * 0)` |
| Q2 (Successful requests) | `or (error * 0)` |
| Q3 (Rate limited requests) | `or (hit * 0) or (error * 0)` |
| Q4 (Failed requests) | `or (hit * 0)` |

## Testing

Verified on `amit.dev` cluster against live Loki data:
- All 4 queries return identical label sets (20 combos each) — confirmed via direct Loki API
- Auth-failure rows (`GARBAGE-401-MODEL`, `WRONG-MODEL-NAME-403`, etc.) excluded from table
- Hit-only combos show `0` in Failed column; error-only combos show `0` in Tokens/Successful/Rate limited
- Descending sort on Tokens places real values first (no empty-before-numbers issue)
- Dropdowns exclude models that only appear in auth-failure logs

## Risk analysis

- **Risk rating**: 2
- **Why**: LogQL query changes only — no controller, Go, or infrastructure changes. Dashboard YAML modifications are straightforward stream selector and pipeline filter additions. Cross-zero-padding pattern is well-understood (`or (query * 0)` is standard LogQL). Verified end-to-end on live cluster. Risk is moderate because the queries are more complex and dashboard rendering depends on Perses `MergeSeries` behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a **Failed requests** column to the Usage breakdown dashboard table.
  * Preserved dashboard rows with zero values when certain response types have no matching data.

* **Bug Fixes**
  * Updated model selections to exclude error responses.
  * Excluded unassigned subscriptions from total request, success-rate, and usage breakdown metrics.
  * Improved consistency of usage queries and zero-valued fallback results for tokens, successful requests, rate-limited requests, and failed requests.

### DIFF
--- a/deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
+++ b/deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
@@ -75,7 +75,7 @@ spec:
               name: usage-logs-all
             labelName: model
             matchers:
-              - '{service_name="models-as-a-service", model!="-"}'
+              - '{service_name="models-as-a-service", model!="-", subscription!="-"}'
     - kind: ListVariable
       spec:
         name: view_by
@@ -343,6 +343,13 @@ spec:
                 format:
                   unit: decimal
                   decimalPlaces: 0
+              - name: "value #4"
+                header: "Failed requests"
+                align: "right"
+                enableSorting: true
+                format:
+                  unit: decimal
+                  decimalPlaces: 0
         queries:
           - kind: TimeSeriesQuery
             spec:
@@ -356,8 +363,13 @@ spec:
                     sum by (user_id, model, subscription) (sum_over_time(
                     {service_name="models-as-a-service",
                     subscription=~"$subscription", model=~"$model", response_type="hit"}
-                    | user_id=~"$user"
+                    | user_id!="-" | user_id=~"$user"
                     | unwrap tokens_total [$__range]))
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", model=~"$model", response_type="error"}
+                    | user_id!="-" | user_id=~"$user" [$__range])) * 0)
           - kind: TimeSeriesQuery
             spec:
               plugin:
@@ -370,7 +382,12 @@ spec:
                     sum by (user_id, model, subscription) (count_over_time(
                     {service_name="models-as-a-service",
                     subscription=~"$subscription", model=~"$model", response_type="hit"}
-                    | user_id=~"$user" [$__range]))
+                    | user_id!="-" | user_id=~"$user" [$__range]))
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", model=~"$model", response_type="error"}
+                    | user_id!="-" | user_id=~"$user" [$__range])) * 0)
           - kind: TimeSeriesQuery
             spec:
               plugin:
@@ -383,12 +400,35 @@ spec:
                     sum by (user_id, model, subscription) (count_over_time(
                     {service_name="models-as-a-service",
                     subscription=~"$subscription", model=~"$model", response_type="rate_limit"}
-                    | user_id=~"$user" [$__range]))
+                    | user_id!="-" | user_id=~"$user" [$__range]))
                     or
                     (sum by (user_id, model, subscription) (count_over_time(
                     {service_name="models-as-a-service",
                     subscription=~"$subscription", model=~"$model", response_type="hit"}
-                    | user_id=~"$user" [$__range])) * 0)
+                    | user_id!="-" | user_id=~"$user" [$__range])) * 0)
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", model=~"$model", response_type="error"}
+                    | user_id!="-" | user_id=~"$user" [$__range])) * 0)
+          - kind: TimeSeriesQuery
+            spec:
+              plugin:
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs-all
+                  query: >-
+                    sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", model=~"$model", response_type="error"}
+                    | user_id!="-" | user_id=~"$user" [$__range]))
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", model=~"$model", response_type="hit"}
+                    | user_id!="-" | user_id=~"$user" [$__range])) * 0)
 
   layouts:
     - kind: Grid
--- a/deployment/components/observability/usage-logs/usage-logs-dashboard.yaml
+++ b/deployment/components/observability/usage-logs/usage-logs-dashboard.yaml
@@ -56,7 +56,7 @@ spec:
               name: usage-logs-multi-tenancy
             labelName: model
             matchers:
-              - '{service_name="models-as-a-service", model!="-"}'
+              - '{service_name="models-as-a-service", model!="-", subscription!="-"}'
     - kind: ListVariable
       spec:
         name: view_by
@@ -293,6 +293,13 @@ spec:
                 format:
                   unit: decimal
                   decimalPlaces: 0
+              - name: "value #4"
+                header: "Failed requests"
+                align: "right"
+                enableSorting: true
+                format:
+                  unit: decimal
+                  decimalPlaces: 0
         queries:
           - kind: TimeSeriesQuery
             spec:
@@ -305,8 +312,15 @@ spec:
                   query: >-
                     sum by (model, subscription, key_name) (sum_over_time(
                     {service_name="models-as-a-service",
-                    subscription=~"$subscription", model=~"$model", response_type="hit"}
+                    subscription=~"$subscription", subscription!="-",
+                    model=~"$model", response_type="hit"}
                     | unwrap tokens_total [$__range]))
+                    or
+                    (sum by (model, subscription, key_name) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", subscription!="-",
+                    model=~"$model", response_type="error"}
+                    [$__range])) * 0)
           - kind: TimeSeriesQuery
             spec:
               plugin:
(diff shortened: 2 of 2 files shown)


deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -356,8 +363,13 @@ spec:
                     sum by (user_id, model, subscription) (sum_over_time(
                     {service_name="models-as-a-service",
                     subscription=~"$subscription", model=~"$model", response_type="hit"}
-                    | user_id=~"$user"
+                    | user_id!="-" | user_id=~"$user"
```
ahadas: we wanted to enable showing data per subscription and model, also when the user chooses not to capture users - for this, we'd need to show usage also when `user_id=="-"`. do we still need this change now that user info is retrieved from headers (rather than FILTER_STATE) ?

tgitelman: captureUser is metrics-only — no Loki toggle for user_id. With usageLogging on, user_id is always emitted on successful auth. user_id="-" means unresolved identity, not “user capture off”. Agree we should remove | user_id!="-" and use subscription!="-" instead to filter auth failures. Will push that change.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -355,9 +362,19 @@ spec:
...
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", subscription!="-", model=~"$model", response_type="error"}
+                    | user_id=~"$user" [$__range])) * 0)
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", subscription!="-", model=~"$model", response_type="rate_limit"}
+                    | user_id=~"$user" [$__range])) * 0)
```
ahadas, line 377: this query is for token-used, right? there's no point to include rate-limited calls here, and probably not those that returned error either - what's the reason for this particular change?

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -369,8 +386,18 @@ spec:
...
```
ahadas, line 400: same here for successful-requests

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -382,12 +409,40 @@ spec:
...
                     (sum by (user_id, model, subscription) (count_over_time(
                     {service_name="models-as-a-service",
-                    subscription=~"$subscription", model=~"$model", response_type="hit"}
+                    subscription=~"$subscription", subscription!="-", model=~"$model", response_type="hit"}
+                    | user_id=~"$user" [$__range])) * 0)
+                    or
+                    (sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", subscription!="-", model=~"$model", response_type="error"}
+                    | user_id=~"$user" [$__range])) * 0)
```
ahadas, line 423: and why changing this?

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -382,12 +409,40 @@ spec:
...
+                kind: LokiTimeSeriesQuery
+                spec:
+                  datasource:
+                    kind: LokiDatasource
+                    name: usage-logs-all
+                  query: >-
+                    sum by (user_id, model, subscription) (count_over_time(
+                    {service_name="models-as-a-service",
+                    subscription=~"$subscription", subscription!="-", model=~"$model", response_type="error"}
+                    | user_id=~"$user" [$__range]))
```
ahadas, line 436: isn't filtering by error response-type sufficient here?

tgitelman:
@ahadas Good questions — the or (... * 0) branches are not mixing response types into those columns.

Each table column is a separate LogQL query merged by Perses MergeSeries on (user_id, model, subscription). A query only counts its primary response_type (e.g. Q1 = hits/tokens, Q2 = hits, Q3 = rate_limit, Q4 = error). The or branches are zero-valued fallbacks: same labels, value 0, so every query returns the same label set.

Without that, a combo that only appears as 429s (or only as errors) exists in one query but not the others → MergeSeries leaves empty cells instead of 0, and descending sort breaks (empty sorts before numbers).

So for example on Q1 (tokens): we are not adding rate-limited or error token counts — those branches are * 0 placeholders only. Q4 is the same in reverse: the real count comes from response_type="error"; hit/rate_limit branches are zero-fill for join alignment, not extra filtering.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: There is a general response bellow.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: There is a general response bellow.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: There is a general response bellow.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: There is a general response bellow.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: Same pattern on all of these — the or (... * 0) branches are not mixing response types into the column’s count.

Each query only counts its own response_type (Q1 = hits/tokens, Q2 = hits, Q3 = rate_limit, Q4 = error). The or branches multiply by zero — they only supply matching (user_id, model, subscription) labels so Perses MergeSeries can join the four queries into one row.

Without them, a label combo that only appears under another response_type produces no series in that query → an empty cell, not 0. That also breaks descending sort (empty sorts before numbers).

Q1/Q2: or (error * 0) / or (rate_limit * 0) — so error-only or 429-only rows still show 0 in Tokens / Successful, not empty.
Q3: only added or (error * 0) on top of the existing or (hit * 0) from main — same reason for error-only rows in the rate-limited column.
Q4: primary query is already response_type="error"; or (hit * 0) / or (rate_limit * 0) are zero-fill so hit-only / 429-only combos get 0 in Failed, not empty.
This is the standard LogQL join pattern; Q3 already used it on main for hit-only rows. We extended it when adding the 4th column so every combo gets a full row.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: Same pattern on all of these — the or (... * 0) branches are not mixing response types into the column’s count.

Each query only counts its own response_type (Q1 = hits/tokens, Q2 = hits, Q3 = rate_limit, Q4 = error). The or branches multiply by zero — they only supply matching (user_id, model, subscription) labels so Perses MergeSeries can join the four queries into one row.

Without them, a label combo that only appears under another response_type produces no series in that query → an empty cell, not 0. That also breaks descending sort (empty sorts before numbers).

Q1/Q2: or (error * 0) / or (rate_limit * 0) — so error-only or 429-only rows still show 0 in Tokens / Successful, not empty.
Q3: only added or (error * 0) on top of the existing or (hit * 0) from main — same reason for error-only rows in the rate-limited column.
Q4: primary query is already response_type="error"; or (hit * 0) / or (rate_limit * 0) are zero-fill so hit-only / 429-only combos get 0 in Failed, not empty.
This is the standard LogQL join pattern; Q3 already used it on main for hit-only rows. We extended it when adding the 4th column so every combo gets a full row.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: Same pattern on all of these — the or (... * 0) branches are not mixing response types into the column’s count.

Each query only counts its own response_type (Q1 = hits/tokens, Q2 = hits, Q3 = rate_limit, Q4 = error). The or branches multiply by zero — they only supply matching (user_id, model, subscription) labels so Perses MergeSeries can join the four queries into one row.

Without them, a label combo that only appears under another response_type produces no series in that query → an empty cell, not 0. That also breaks descending sort (empty sorts before numbers).

Q1/Q2: or (error * 0) / or (rate_limit * 0) — so error-only or 429-only rows still show 0 in Tokens / Successful, not empty.
Q3: only added or (error * 0) on top of the existing or (hit * 0) from main — same reason for error-only rows in the rate-limited column.
Q4: primary query is already response_type="error"; or (hit * 0) / or (rate_limit * 0) are zero-fill so hit-only / 429-only combos get 0 in Failed, not empty.
This is the standard LogQL join pattern; Q3 already used it on main for hit-only rows. We extended it when adding the 4th column so every combo gets a full row.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
tgitelman: Same pattern on all of these — the or (... * 0) branches are not mixing response types into the column’s count.

Each query only counts its own response_type (Q1 = hits/tokens, Q2 = hits, Q3 = rate_limit, Q4 = error). The or branches multiply by zero — they only supply matching (user_id, model, subscription) labels so Perses MergeSeries can join the four queries into one row.

Without them, a label combo that only appears under another response_type produces no series in that query → an empty cell, not 0. That also breaks descending sort (empty sorts before numbers).

Q1/Q2: or (error * 0) / or (rate_limit * 0) — so error-only or 429-only rows still show 0 in Tokens / Successful, not empty.
Q3: only added or (error * 0) on top of the existing or (hit * 0) from main — same reason for error-only rows in the rate-limited column.
Q4: primary query is already response_type="error"; or (hit * 0) / or (rate_limit * 0) are zero-fill so hit-only / 429-only combos get 0 in Failed, not empty.
This is the standard LogQL join pattern; Q3 already used it on main for hit-only rows. We extended it when adding the 4th column so every combo gets a full row.

deployment/components/observability/usage-logs/usage-logs-admin-dashboard.yaml
```diff
@@ -75,7 +75,7 @@ spec:
               name: usage-logs-all
             labelName: model
             matchers:
-              - '{service_name="models-as-a-service", model!="-"}'
+              - '{service_name="models-as-a-service", model!="-", response_type!="error"}'
```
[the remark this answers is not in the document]
ahadas: that's on purpose, we don't want the models list to be polluted with models from failed requests - we want this list to include only valid models

tgitelman:
Great solution for the column merge issue.

Wanted to raise a concern about the model dropdown change from subscription!="-" to response_type!="error": my original intent was to remove auth-failure junk (subscription="-", garbage model names from 401s). Legit models that only error would still be selectable. With response_type!="error", error-only models are filtered out of the dropdown and may hide the problem — they still show in the table with Model = All, but you can't drill down to a model that's failing 100%.

tgitelman:
@ahadas I'll redeploy the dashboards and check that everything correct.

PR #1323, opened 2026-08-04
feat(ipp): add HPA-based autoscaling for payload-processing pods

EgorLu:
## Summary

Adds CPU/memory-based horizontal pod autoscaling (HPA) for IPP (payload-processing) pods. Today there is no autoscaling infrastructure for IPP -- only static replica counts with a manual annotation override. PSAP performance testing found that IPP needs horizontal scale-out under load.

### New tenant annotations

| Annotation | Default | Description |
|---|---|---|
| `maas.opendatahub.io/payload-processing-autoscaling` | (not set) | Enable/disable HPA (`true`/`false`) |
| `maas.opendatahub.io/payload-processing-max-replicas` | 10 | HPA ceiling (1-100) |
| `maas.opendatahub.io/payload-processing-target-cpu` | 70 | Target CPU utilization % |
| `maas.opendatahub.io/payload-processing-target-memory` | 80 | Target memory utilization % |

### Interaction with existing replica annotation

When autoscaling is enabled, `payload-processing-replicas` sets HPA `minReplicas` (floor) instead of `Deployment.spec.replicas`, preventing the controller and HPA from fighting over replica count. When autoscaling is disabled, existing behavior is unchanged.

### Implementation

- HPA is generated in PostRender (same pattern as TelemetryPolicy/Istio Telemetry) to avoid kustomize name-reference conflicts with the pre-processing Deployment
- Scale-down: 300s stabilization window, 25%/60s policy (prevents flapping)
- Scale-up: immediate (0s stabilization), 100%/15s or 4 pods/15s (whichever is higher)
- Explicit HPA cleanup when autoscaling is disabled (SSA doesn't prune absent resources)
- RBAC: added `autoscaling/horizontalpodautoscalers` verbs to maas-controller ClusterRole
- No new dependencies -- HPA v2 is built into every OpenShift/K8s cluster

### Files changed

- `deployment/base/maas-controller/rbac/clusterrole.yaml` -- HPA RBAC permissions
- `maas-controller/pkg/platform/tenantreconcile/constants.go` -- new annotations, GVK, naming function
- `maas-controller/pkg/platform/tenantreconcile/params.go` -- autoscaling param resolution, deployment replica skip when HPA active
- `maas-controller/pkg/platform/tenantreconcile/postrender.go` -- HPA resource generation in PostRender
- `maas-controller/pkg/platform/tenantreconcile/pipeline.go` -- HPA cleanup when autoscaling disabled
- `maas-controller/pkg/platform/tenantreconcile/params_test.go` -- 9 new test cases
- `docs/content/reference/crds/tenant.md` -- autoscaling documentation and examples

### Testing

**Unit tests:** 9 new test cases covering default values, custom values, invalid inputs, interaction with replica annotation, deployment replica skip. All existing tests pass unchanged.

**Live cluster test (bbr-test.zr64):**
1. Enabled autoscaling via annotation -- HPA created with correct defaults (CPU 70%, memory 80%, min 1, max 10)
2. Set custom thresholds -- HPA updated to min 2, max 15, CPU 60%, memory 75%
3. Disabled autoscaling -- HPA deleted (cleanup confirmed)
4. Re-enabled autoscaling -- HPA recreated with custom values
5. Removed all annotations -- HPA deleted, clean state

Ref: [RHOAIENG-79786](https://redhat.atlassian.net/browse/RHOAIENG-79786)
Related: [PSAP-2826](https://redhat.atlassian.net/browse/PSAP-2826)

[RHOAIENG-79786]: https://redhat.atlassian.net/browse/RHOAIENG-79786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Payload-processing components now support horizontal pod autoscaling with configurable CPU and memory targets.
  * Tenant configuration supports replica limits, autoscaling settings, and scaling stabilization policies.
  * Autoscaling manages deployment replicas automatically, with configurable minimum and maximum bounds.

* **Bug Fixes**
  * Invalid replica settings are constrained and reported with warnings.
  * Disabling autoscaling removes existing autoscalers and restores configured replica management.

* **Documentation**
  * Added configuration guidance, validation details, migration notes for deprecated annotations, and YAML examples.


jland-redhat:
Have we validated this works (and any chance we can get like a script showing the performance impact)?
@EgorLu 

If so I wonder if this is something we want to create a blocker for and include in 3.5 initial release? WDYT @ishitasequeira 

I was thinking HPAs would be something we wanted to backport anyway

maas-controller/pkg/platform/tenantreconcile/postrender.go
```diff
@@ -342,6 +345,107 @@ func configureIstioTelemetryResources(log logr.Logger, tenant client.Object, res
 	return nil
 }
 
+// configurePayloadProcessingHPA appends an HPA for payload-processing when autoscaling is enabled.
+// The HPA is generated in PostRender (not from kustomize) to avoid name-reference conflicts
+// with the pre-processing Deployment that shares the same base name.
+func configurePayloadProcessingHPA(log logr.Logger, resources *[]unstructured.Unstructured, params PlatformParams) error {
```
ishitasequeira, line 351: Instead of hard-coding the logic here, can we consider adding the HPA to payload-processing manifests instead? https://github.com/opendatahub-io/models-as-a-service/tree/main/deployment/base/payload-processing/instance

maas-controller/pkg/platform/tenantreconcile/pipeline.go
```diff
@@ -95,6 +95,14 @@ func RunPlatform(
 		return nil, fmt.Errorf("apply: %w", err)
 	}
 
+	// Clean up orphaned HPA when autoscaling is disabled. SSA only creates/updates
+	// resources in the rendered set; it does NOT delete resources that were previously
+	// rendered but are now absent. Without explicit cleanup, disabling autoscaling
+	// would leave a stale HPA that keeps scaling pods.
+	if err := cleanupPayloadProcessingHPA(ctx, c, params, log); err != nil {
+		return nil, fmt.Errorf("cleanup payload-processing HPA: %w", err)
+	}
```
ishitasequeira, line 100: Any reason we are making this optional and not default HPA installation? The number of pods could stay at 1 by default and we can let customer update the replicas on HPA.

docs/content/reference/crds/tenant.md
```diff
@@ -118,6 +137,25 @@ spec:
...
+kind: MaasTenantConfig
+metadata:
+  name: default-tenant
+  namespace: models-as-a-service
+  annotations:
+    maas.opendatahub.io/payload-processing-autoscaling: "true"
+    maas.opendatahub.io/payload-processing-replicas: "2"       # minimum replicas (floor)
+    maas.opendatahub.io/payload-processing-max-replicas: "15"  # maximum replicas (ceiling)
+    maas.opendatahub.io/payload-processing-target-cpu: "60"    # scale up when avg CPU > 60%
+    maas.opendatahub.io/payload-processing-target-memory: "75" # scale up when avg memory > 75%
```
ishitasequeira: If we do want to go with the route of adding this config on MaasTenantConfig, have we considered making it part of the spec instead of annotations?

maas-controller/pkg/platform/tenantreconcile/postrender.go
EgorLu: Good question. I tried kustomize manifests first. Added `hpa.yaml` to `deployment/base/payload-processing/instance/`. It broke because both `after/` and `pre-processing/` include `../instance`, creating two copies of the HPA resource.
Kustomize name-reference resolver then found two Deployments (`payload-processing` and `payload-pre-processing`) as candidates for the HPA `scaleTargetRef.name` and errored with `too many possible referral targets`.

PostRender generation avoids this entirely. Same pattern used for TelemetryPolicy and Istio Telemetry resources. It also naturally handles per-tenant naming and namespace assignment.

EgorLu:
@jland-redhat Yes, validated on a new cluster.
Tested the full lifecycle: enable autoscaling -> HPA created with correct defaults -> update thresholds -> HPA updated -> disable autoscaling -> HPA deleted -> re-enable -> HPA recreated. All scenarios worked.

For performance impact, do we want to create our own script, or ask PSAP to re-run their load test with the HPA enabled?

maas-controller/pkg/platform/tenantreconcile/postrender.go
ishitasequeira: So, are we trying to have a single HPA for managing both IPP `payload-pre-processing` and `payload-processing` pods deployments?

We can create a single HA overlay and use that instead of base manifests if HA is enabled by customers.

docs/content/reference/crds/tenant.md
ishitasequeira: I would suggest to have it moved to spec with below config:

```
spec:
    haEnabled: true
    payload-processing:
        replicas: 2
        max-replicas: 15
        cpu: 60
        target-memory: 75
```

@jland-redhat any thoughts?

maas-controller/pkg/platform/tenantreconcile/postrender.go
EgorLu: The HPA is only for `payload-processing`, not `payload-pre-processing`.
The kustomize conflict was just because both share the instance / base layer.

Wouldn't HA overlay still need PostRender patching for per-tenant naming?

ishitasequeira (commented):
Overall the change looks good. Couple of final things that caught my attention:

1. configurePayloadProcessingHPA and cleanupPayloadProcessingHPA have no direct tests — the new tests only cover param resolution, not the actual HPA object (scaleTargetRef, minReplicas/maxReplicas) or the delete-on-disable path.
2. The min>max clamp warning still says payload-processing-replicas/payload-processing-max-replicas — should reference the new spec.payloadProcessing.* fields instead.

Also, could we rebase the PR to latest main?

docs/content/reference/crds/tenant.md
EgorLu: Done -- refactored in the latest commit. The spec structure is:

```yaml
spec:
  payloadProcessing:
    replicas: 2
    autoscaling:
      maxReplicas: 15
      targetCPUUtilization: 60
      targetMemoryUtilization: 75
```

Presence of the `autoscaling` section enables HPA (no separate `haEnabled` boolean needed). Removing the section disables it and triggers HPA cleanup. CRD validation enforces ranges via kubebuilder markers. The old annotation-based replicas are kept as a deprecated fallback.

maas-controller/pkg/platform/tenantreconcile/postrender.go
EgorLu: No, the HPA only targets `payload-processing`. `payload-pre-processing` doesn't get an HPA -- it's a separate deployment with different scaling characteristics.

The kustomize conflict was purely structural: both `after/` and `pre-processing/` reference `../instance`, so any resource added to `instance/` appears twice in the kustomize build. The HA overlay approach would still hit the same issue unless the HPA is placed outside the shared `instance/` layer -- and it would still need PostRender patching for per-tenant deployment names and namespaces.

PostRender generation keeps it simple and consistent with how TelemetryPolicy and Istio Telemetry are already handled in this codebase.

ishitasequeira (approved):
Overall change looks good!! Thanks @EgorLu !!

PR #1348, opened 2026-08-07
fix: prefer HTTPS listener in /v1/tenants gateway URL discovery

liangwen12year:
## Summary

Fixes: [RHOAIENG-81214](https://redhat.atlassian.net/browse/RHOAIENG-81214) (second bug)

The `/v1/tenants` endpoint picks the first ready Gateway listener regardless of protocol. When the HTTP (port 80) listener is listed first, it returns `http://` as the gateway URL. The BFF then sends POST requests to the HTTP endpoint:

1. POST to `http://` → gateway 302 redirects to `https://`
2. Go's `http.Client` follows 302 and converts POST to GET (per RFC 7231)
3. GET on `/v1/api-keys/search` doesn't match the POST handler → 404

### Fix

Prefer HTTPS/TLS listeners over HTTP when building the external URL. Fall back to HTTP only if no HTTPS listener exists.

### Evidence

maas-ui logs on `ods-qe-psi-11` (3.6-ea.1):
```
endpoint=http://maas.apps.ods-qe-psi-11.osp.rh-ods.com/maas-api/v1/api-keys/search
statusCode=404 error="API key not found"
```

`/v1/tenants` returns:
```json
{"protocol": "http", "externalUrl": "http://maas.apps...", "port": 80}
```

## Test plan
- [ ] `/v1/tenants` returns `https://` URL when HTTPS listener exists
- [ ] API Keys page loads in the UI
- [ ] Falls back to HTTP correctly when only HTTP listener exists

[RHOAIENG-81214]: https://redhat.atlassian.net/browse/RHOAIENG-81214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Gateway configuration now prioritizes the first ready secure HTTPS/TLS listener.
  * Retains a ready HTTP listener as a fallback when no secure listener is available.
  * Improves listener port and hostname handling across protocol selections.
  * Prevents tenant cleanup from resetting the shared default gateway authentication policy incorrectly.

### DIFF
--- a/maas-api/internal/tenant/handler.go
+++ b/maas-api/internal/tenant/handler.go
@@ -259,7 +259,7 @@ func selectBestListener(specListeners []any, statusByName map[string]map[string]
 			return listenerPort, listenerProtocol, listenerHostname
 		}
 
-		if !foundHTTP {
+		if listenerProtocol == protocolHTTP && !foundHTTP {
 			foundHTTP = true
 			httpPort = listenerPort
 			httpHostname = listenerHostname
--- a/maas-api/internal/tenant/handler.go
+++ b/maas-api/internal/tenant/handler.go
@@ -269,5 +269,6 @@ func selectBestListener(specListeners []any, statusByName map[string]map[string]
 	if foundHTTP {
 		return httpPort, protocolHTTP, httpHostname
 	}
+	// No ready listeners found; return safe defaults.
 	return 443, protocolHTTPS, ""
 }
--- a/maas-api/internal/tenant/handler.go
+++ b/maas-api/internal/tenant/handler.go
@@ -17,6 +17,12 @@ import (
 	"github.com/opendatahub-io/models-as-a-service/maas-api/internal/logger"
 )
 
+const (
+	protocolHTTPS = "HTTPS"
+	protocolTLS   = "TLS"
+	protocolHTTP  = "HTTP"
+)
+
 // Handler handles /v1/tenant endpoint requests.
 type Handler struct {
 	log              *logger.Logger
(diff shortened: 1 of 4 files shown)


jland-redhat (changes requested):
This change seems reasonable but could we do a little more research on this possibly ask some folks from the Gateway Team how to best accomplish this?

Also similar to my other note it looks like this has logic from a different pr:
https://github.com/opendatahub-io/models-as-a-service/pull/1350

yossiovadia (commented):
Two independent bug fixes, both logically correct. Two test gaps worth addressing before merge:

**1. Missing test for the actual bug scenario in `selectBestListener`**
The production bug was: HTTP listener listed before HTTPS → HTTP URL returned → BFF POST follows 302 → Go converts to GET → 404. The fix is correct but there's no test covering this exact ordering. Needed:
```go
// listeners: [{HTTP, port 80, ready}, {HTTPS, port 443, ready}]
// expected: HTTPS wins (port 443)
```
Also missing: TLS protocol case, and HTTP-only-ready when HTTPS has `attachedRoutes: 0`.

**2. Missing test for the controller deletion guard**
The `!isDefaultTenantNamespace && !isNonDefaultTenant && isDefaultGateway` branch in `handleDeletion` has no unit test. The specific race — non-default tenant namespace, `MaasTenantConfig` already deleted, controller falls back to default gateway — needs a test to pin the behavior. Without it a future refactor silently breaks it.

Minor: `selectBestListener` returns `(443, protocolHTTPS, "")` both for "HTTPS listener with no hostname" and "no ready listeners at all" — same defaults, ambiguous contract. A comment on the final return clarifying "no ready listeners found, returning defaults" would help future readers.

yossiovadia (approved):
All findings addressed — 7 new tests covering the HTTP-before-HTTPS ordering, TLS protocol, no-ready-listeners edge cases, and the controller deletion guard. Both fixes are correct. LGTM

lphiri:
@liangwen12year  @jland-redhat the workaround logic looks fine to me,  however I am a bit puzzled by the redirect behaviour.  If we have both TLS and non-TLS listeners configured , both are valid entrypoints.
This actually made me wonder if  /v1/tenants should actually return multiple addresses.
For now we can merge as is, but we should document the root cause of the redirect.

lphiri:
Looked further in the reproducer cluster and it appears the redirect was caused by the OCP route configuration, which is inconsistent with the Gateway API configuration  - the route pointing the gateway has 
```
tls:
     insecureEdgeTerminationPolicy: Redirect
     termination: passthrough
 ```

jland-redhat (approved):
lgtm

PR #1351, opened 2026-08-07
fix: replace poll with warmup + retry for flaky test

ishitasequeira:
## Description
replace poll with warmup + retry for flaky test

## Merge criteria:

## Summary by CodeRabbit

- **Tests**
  - Improved end-to-end test reliability by actively checking gateway readiness before proceeding.
  - Added retries for temporary gateway and authentication failures during startup checks.
  - Tests now verify successful hybrid chat responses before continuing.
  - Startup checks stop promptly when they encounter non-retryable gateway errors.

### DIFF
--- a/test/e2e/tests/test_per_tenant_ipp_isolation.py
+++ b/test/e2e/tests/test_per_tenant_ipp_isolation.py
@@ -89,6 +89,14 @@ def _request_with_gateway_retry(method, url, retries=GATEWAY_PROPAGATION_RETRIES
             )
             time.sleep(GATEWAY_PROPAGATION_DELAY)
             continue
+        if response.status_code not in (200, 201):
+            log.info(
+                "Gateway returned non-retryable %d (attempt %d/%d, body: %.300s)",
+                response.status_code,
+                attempt,
+                retries,
+                response.text[:300],
+            )
         return response
     return response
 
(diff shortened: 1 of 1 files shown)


yu-teo (commented):
The old `_poll_status` polled until it got a specific expected status code (e.g. `200`), retrying everything else. 
But the new loop also retries everything `non-200` - including responses that likely indicate a real problem (`404`, `405`, `422`, etc.). I see a major difference in that `_poll_status` would raise an `AssertionError` with clear diagnostics **after** timeout, while this loop logs each `non-200` as a warning but keeps going for 90 seconds before the assert fires. So, a simple misconfiguration (for example, wrong model path or a broken route) would spin for 90 seconds instead of surfacing the real error. 

Edit: maybe, we should compare this to `_request_with_gateway_retry`, which only retries the two known-transient patterns (empty `403`, `500 AUTH_FAILURE`) and returns immediately on anything else?  The warmup loop could adopt the same pattern. As in, retry only transient statuses, fail fast on obviously wrong ones (`4xx` other than empty-`403`, `5xx` other than `AUTH_FAILURE`)? Just an idea

ishitasequeira:
> The old `_poll_status` polled until it got a specific expected status code (e.g. `200`), retrying everything else. But the new loop also retries everything `non-200` - including responses that likely indicate a real problem (`404`, `405`, `422`, etc.). I see a major difference in that `_poll_status` would raise an `AssertionError` with clear diagnostics **after** timeout, while this loop logs each `non-200` as a warning but keeps going for 90 seconds before the assert fires. So, a simple misconfiguration (for example, wrong model path or a broken route) would spin for 90 seconds instead of surfacing the real error.
> 
> Edit: maybe, we should compare this to `_request_with_gateway_retry`, which only retries the two known-transient patterns (empty `403`, `500 AUTH_FAILURE`) and returns immediately on anything else? The warmup loop could adopt the same pattern. As in, retry only transient statuses, fail fast on obviously wrong ones (`4xx` other than empty-`403`, `5xx` other than `AUTH_FAILURE`)? Just an idea

Good catch. You're right that the warmup loop was retrying  everything blindly, a 404 or 422 would spin for 90s hiding a real misconfiguration.

That said, the reason the outer warmup loop exists at all is that _request_with_gateway_retry alone (6 retries × 5s = 30s) 
wasn't enough — auth propagation after the prior 19 test modules' MaaSAuthPolicy churn can take longer. So we can't    
fail fast on 403/500, those are the actual transient pattern we're waiting out.                                            

Updated the warmup loop to fail fast on terminal status codes (404, 405, 422, etc.) while still retrying the known transient ones (403, 500, 502, 503). The bigger diagnostic win is the warmup is not None fix and the per-attempt logging — the bool(Response) bug was hiding the actual 403 body in every CI run since this test was created, so every failure showed "no response" instead of the real error. Next CI run should finally reveal the root cause.

jland-redhat (approved):
Lgtm

chaitanya1731 (approved):
lgtm

chaitanya1731 (approved):
lgtm

jland-redhat (approved):
lgtm

chaitanya1731 (approved):
lgtm

PR #1363, opened 2026-08-12
feat(observability): add auth decision recording rules, deny-rate alert, and Perses dashboard

asanzgom:
## Summary

- Add recording rules for Authorino auth decision metrics and a high deny-rate alert to the existing PrometheusRule
- Add promtool unit tests (9 test cases) for all recording rules and the alert
- Add auth decisions Perses dashboard (`dashboard-4-maas-auth-decisions`) alongside the existing Usage dashboards
- Document recording rules, alert, alert overlap, and common queries in the observability metrics reference
- Update `install-observability.sh` summary to list new recording rules and alert

### Recording rules

| Rule | Description |
|------|-------------|
| `maas:auth_decisions:rate5m` | Per-policy auth decision rate by status |
| `maas:auth_deny_ratio:rate5m` | Per-policy deny ratio (by authconfig and namespace) |
| `maas:auth_latency_p95:5m` | Per-policy P95 auth decision latency (by authconfig and namespace) |

### Alert

| Alert | Severity | Condition |
|-------|----------|-----------|
| `MaaSHighAuthDenyRate` | Warning | >10% deny ratio sustained for 10 minutes |

### Perses dashboard

Auth decisions dashboard at `deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml`, deployed by maas-controller alongside the existing Usage dashboards. Panels:

- **Overview stats**: decisions/sec, deny rate, avg policy P95 latency, active policies
- **Auth decision trends**: decisions by status (stacked area), decisions by policy
- **Per-policy details**: deny ratio by policy, P95 latency by policy

Metrics sourced from Authorino `/server-metrics` endpoint. Label schema verified on cluster: `{authconfig, namespace, status}`. Metrics are lazy-initialized (only appear after first auth traffic).

Ref: [RHAISTRAT-2418](https://redhat.atlassian.net/browse/RHAISTRAT-2418), [RHAI-493](https://redhat.atlassian.net/browse/RHAI-493), [RHAI-496](https://redhat.atlassian.net/browse/RHAI-496)

## Risk analysis

- **Risk rating**: 3
- **Why**: Additive changes — new recording rules, alert, Perses dashboard, documentation, and script summary. No existing alerts, rules, or dashboards modified. The recording rules and alert are deployed by `install-observability.sh` (external script), not by the operator. The Perses dashboard follows the same deployment model as the existing Usage dashboards (deployed by maas-controller via kustomize component). Risk drivers: (1) the alert could be noisy if the 10% threshold is too sensitive, mitigated by the 10-minute `for:` window and `> 0.001 rps` traffic guard; (2) the dashboard adds a new PersesDashboard CR that the monitoring controller must reconcile.

## Test plan

- [x] `promtool test rules maas-auth-alerting.unit-tests.yaml` passes (9 tests)
- [ ] Recording rules produce correct values on a cluster with active MaaS inference traffic
- [x] Alert fires when deny ratio exceeds 10% for 10 minutes
- [x] Alert does NOT fire below 10% deny ratio
- [x] Alert does NOT fire before 10 minutes sustained
- [x] Alert does NOT fire with negligible traffic (minimum traffic guard)
- [x] Namespace filtering excludes non-Authorino namespaces
- [x] Existing `MaaSAuthorinoMetadataEvaluatorHighFailureRate` alert unaffected
- [x] Kustomize build passes for `deployment/components/observability/observability/dashboards/`
- [ ] Dashboard renders in Perses UI with data populated in all panels
- [x] Dashboard handles cold start (no auth traffic) gracefully — stat panels show 0, time series show no data

[RHAISTRAT-2418]: https://redhat.atlassian.net/browse/RHAISTRAT-2418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


asanzgom:
@coderabbitai review

somya-bhatnagar (commented):
## Review notes:

### 1. Overlap with existing auth failure alert

`MaaSAuthorinoAuthenticationHighFailureRate` (`authorino-maas-oidc-prometheusrule.yaml`) already fires on >10% `UNAUTHENTICATED` (cluster-wide, `for: 5m`).

`MaaSHighAuthDenyRate` fires on >10% `status!="OK"` (per `authconfig`, `for: 10m`).

On sustained UNAUTHENTICATED spikes both can warn. Please either document why both are needed (granularity + broader status set), or consolidate / tighten so they do not double-page for the same failure mode.

### 2. Wrong home for these rules

Auth decision recording rules + deny alert were appended to `authorino-maas-metadata-evaluator-prometheusrule.yaml` (CR name still `authorino-maas-metadata-evaluator-high-failure-rate`). That file is for metadata evaluator → maas-api failures.

These rules fit better in `authorino-maas-oidc-prometheusrule.yaml` (same underlying metrics) or a dedicated PrometheusRule. Header/comments are now misleading; `install-observability.sh` summary still only lists the old three alerts.

### 3. Docs not updated

`docs/content/observability/metrics-and-dashboards.md` and `docs/content/advanced-administration/external-oidc.md` document existing auth alerts/metrics but not the new recording rules or `MaaSHighAuthDenyRate`.

---

**Also worth noting (lower priority):** promtool unit tests look good but are not wired into CI; PR body is missing the repo-required Risk analysis section; Konflux `maas-group-test` failed on this PR (unclear if related).

asanzgom (commented):
Thanks for the thorough review @somya-bhatnagar! Addressing each point:

### 1. Overlap with existing auth failure alert

These alerts intentionally serve different purposes and are not redundant:

| | `MaaSAuthorinoAuthenticationHighFailureRate` | `MaaSHighAuthDenyRate` |
|---|---|---|
| **Scope** | Cluster-wide (all AuthConfigs aggregated) | Per-authconfig (per-policy) |
| **Status** | `UNAUTHENTICATED` only | All non-OK (`UNAUTHENTICATED`, `PERMISSION_DENIED`, `NOT_FOUND`) |
| **Duration** | `for: 5m` | `for: 10m` |
| **Question answered** | "Is the IdP/OIDC provider broken?" | "Which specific policy is denying traffic?" |

On a sustained `UNAUTHENTICATED` spike, both will fire — that's expected and useful. The OIDC alert tells you _something is wrong with authentication_, the deny-rate alert tells you _which policy_ is affected. An operator responding to the OIDC alert would still want per-policy attribution to identify the blast radius.

`PERMISSION_DENIED` and `NOT_FOUND` denials are only caught by `MaaSHighAuthDenyRate` — the OIDC alert doesn't cover these at all.

I'll add a comment in the PrometheusRule YAML explaining this relationship.

### 2. Wrong home for these rules

Agree the CR name `authorino-maas-metadata-evaluator-high-failure-rate` is now misleading. However, moving to a new PrometheusRule file would also require updating `install-observability.sh` to deploy it with dynamic namespace substitution. For this PR I chose to extend the existing file to avoid that additional deployment plumbing — both rule groups use the same underlying metrics from the same `/server-metrics` endpoint and need the same namespace handling.

Happy to split into a dedicated PrometheusRule in a follow-up if the team prefers.

The `install-observability.sh` summary has been updated in commit 38df232 to list the new recording rules and alert.

### 3. Docs not updated

Fixed in commit eb190ed — recording rules, `MaaSHighAuthDenyRate` alert, hash-to-name guidance, namespace filtering notes, and common PromQL queries added to `docs/content/observability/metrics-and-dashboards.md`.

### Lower-priority items

- **Risk analysis**: Added to PR body.
- **Konflux failure**: Current checks show all CI passing, Konflux builds pending (not failed). The `maas-group-test` isn't in the current check set — may have been a transient issue on an earlier commit.
- **promtool in CI**: Good callout — will track as a follow-up.

docs/content/observability/metrics-and-dashboards.md
```diff
@@ -84,6 +84,28 @@ Exposed on `/server-metrics` (port 8080):
...
+| `maas:auth_latency_p95:5m` | P95 auth evaluation latency per `authconfig` |
+
+| Alert | Condition | Severity |
+|-------|-----------|----------|
+| **`MaaSHighAuthDenyRate`** | >10% deny ratio sustained for 10 minutes (with minimum traffic guard >0.001 rps) | warning |
+
+**Remediate:** AuthPolicy configuration; credential validity (expired tokens, revoked API keys); Authorino logs for the specific `authconfig` hash.
+
+!!! note "`authconfig` labels are SHA-256 hashes"
+    Authorino names AuthConfigs as SHA-256 hashes of the policy name. Recording rules and alerts use these hash values directly. Map hashes to policies via `kubectl get authconfig -n <authorino-namespace>`.
```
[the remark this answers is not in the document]
asanzgom: Skipping this finding — the current documentation is correct. The authconfig metric label does contain the AuthConfig resource name, but Authorino names its AuthConfig resources as SHA-256 hashes of the policy. The label values are hex strings like b88ddc37d1..., not human-readable names. This was confirmed on a live cluster. The hash-mapping guidance (kubectl get authconfig -n <authorino-namespace>) is the correct way for operators to resolve these labels to the underlying policies.

docs/content/observability/metrics-and-dashboards.md
```diff
@@ -153,6 +175,20 @@ sum(rate(vllm:request_success_total[5m])) /
...
 
+**Auth decisions:**
+
+```promql
+# Deny ratio per policy
+maas:auth_deny_ratio:rate5m
+
+# Overall deny rate across all policies
+sum(maas:auth_decisions:rate5m{status!="OK"}) /
+  sum(maas:auth_decisions:rate5m)
```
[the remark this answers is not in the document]
asanzgom: Skipping this finding. These are documentation example queries under "Common Queries" — illustrative PromQL snippets, not deployed recording rules or alert expressions. The recording rule maas:auth_deny_ratio:rate5m already handles the all-OK case with a zero fallback in its actual expression (the or 0 * group by (authconfig, namespace)(...) clause added in commit 01d7150). Adding or vector(0) to a documentation example query would make it harder to read without meaningful benefit — an operator copy-pasting this would get an empty result in the all-OK case, which is the correct Prometheus behavior (no denials = no ratio to show).

asanzgom:
### Konflux `maas-group-test` failure analysis

Investigated the [test artifacts](https://app-artifact-browser.apps.rosa.konflux-qe.zmr9.p3.openshiftapps.com/odh-ci-artifacts/maas-group-test-84fql/e2e-maas-openshift/).

**Root cause**: `odh-model-controller` entered `CrashLoopBackOff` (6 restarts, 0/1 ready) during cluster setup. The DSC never reached `Ready`, so the entire MaaS stack never deployed — no maas-api pods, no maas-controller, no CRs, no AuthPolicies. E2E tests had no infrastructure to run against.

**Not related to this PR.** This PR only adds observability manifests (PrometheusRule recording rules, promtool unit tests, documentation). It doesn't touch Go code, controller logic, or deployment manifests that affect operator startup. The failure is an infrastructure/setup issue with the test cluster's ODH operator bootstrap.

somya-bhatnagar:
## Review notes

### 1. File placement (suggestion)

The new `maas.authorino.auth-decisions` group is in `authorino-maas-metadata-evaluator-prometheusrule.yaml`, while it uses the same `auth_server_authconfig_*` metrics as `authorino-maas-oidc-prometheusrule.yaml` / `authorino-maas-authentication-alerts`. Functionally fine (both are applied by `install-observability.sh`), but the CR name/header still read as metadata-evaluator-only. Have you considered moving this group to the authentication PrometheusRule (or a dedicated CR) for discoverability?

### 2. Relationship to existing auth failure alert (docs clarity)

`MaaSAuthorinoAuthenticationHighFailureRate` (cluster `UNAUTHENTICATED`, 5m) and `MaaSHighAuthDenyRate` (per-`authconfig` `status!="OK"`, 10m) look complementary, but can both warn on sustained UNAUTHENTICATED spikes. A short note in the metrics docs on when each is expected (and that dual-firing can be intentional) would help operators.

asanzgom:
Thanks for the follow-up @somya-bhatnagar!

### 1. File placement

Agree the CR name is misleading. For this PR I kept both groups in the same file to avoid adding another file to `install-observability.sh`'s dynamic namespace substitution flow — both groups use the same Authorino `/server-metrics` metrics and need the same namespace handling. Happy to split into a dedicated CR in a follow-up if the team prefers.

### 2. Alert overlap docs

Good call — added a note in `docs/content/observability/metrics-and-dashboards.md` (commit 533eed8) explaining the complementary relationship:

- `MaaSAuthorinoAuthenticationHighFailureRate` — cluster-wide, `UNAUTHENTICATED` only, answers "is the IdP broken?"
- `MaaSHighAuthDenyRate` — per-authconfig, all non-OK statuses, answers "which policy is denying traffic?"

Dual-firing on sustained `UNAUTHENTICATED` spikes is intentional — gives operators both the aggregate signal and per-policy attribution.

deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml
```diff
@@ -0,0 +1,328 @@
+# Auth decision monitoring dashboard — shows per-policy auth decision rates,
+# deny ratios, and P95 latency from Authorino recording rules.
+# Requires recording rules: maas:auth_decisions:rate5m, maas:auth_deny_ratio:rate5m,
+# maas:auth_latency_p95:5m (deployed via install-observability.sh).
+apiVersion: perses.dev/v1alpha1
+kind: PersesDashboard
+metadata:
+  name: dashboard-4-maas-auth-decisions
```
ahadas: I see this dashboard is not defined for admins-only - shouldn't it be as it displays information that is not specific for the logged in user?
I also wonder if it wouldn't make more sense to extend the existing usage dashboard for admins with this data instead, wdyt?

deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml
```diff
@@ -0,0 +1,328 @@
...
+          description: Filter by AuthConfig policy (SHA-256 hash)
+        allowMultiple: true
+        allowAllValue: true
+        customAllValue: ".*"
+        defaultValue: "$__all"
+        plugin:
+          kind: PrometheusLabelValuesVariable
+          spec:
+            datasource:
+              kind: PrometheusDatasource
```
ahadas, line 33: this means we'll use the default datasource which is `cluster-prometheus-datasource` - and it will require enabling observability in kuadrant and also to enable UWM. part of the changes in 3.5 regarding the usage dashboard was to transition to the ODH/RHOAI monitoring stack instead - I think it makes sense to do that for the information here as well

deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml
asanzgom: Renamed to `dashboard-4-maas-auth-decisions-admin` — agreed this data is cross-tenant and should be admin-only. 

Keeping it as a separate dashboard rather than merging into the existing usage dashboard since it has a different data domain (auth decisions vs. token consumption/rate limiting) and its own variable (authconfig filter). 

Happy to merge them if you feel strongly about it.

deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml
asanzgom: Updated all datasource references to use `data-science-prometheus-datasource` to match the 3.5 transition to the ODH/RHOAI monitoring stack. 
The variable query and all 8 panel queries now explicitly reference it.

ahadas: cool, but I think it's not enough since we have no pod/service monitor for authorino that is labeled with `monitoring.opendatahub.io/scrape: 'true'`, or do we?..

deployment/components/observability/observability/dashboards/auth-decisions-dashboard.yaml
ahadas: > Renamed to `dashboard-4-maas-auth-decisions-admin` — agreed this data is cross-tenant and should be admin-only.

thanks

> Keeping it as a separate dashboard rather than merging into the existing usage dashboard since it has a different data domain (auth decisions vs. token consumption/rate limiting) and its own variable (authconfig filter).
> 
> Happy to merge them if you feel strongly about it.

I'm trying to figure out with myself here whether I feel strongly about it :) 
the thing is, that we thought about adding another dashboard for MaaS health / readiness / whatever-we-want-to-call-it, I wonder if it would fit that one eventually.. or we can say that the usage dashboard is not only about token consumption / rate limited but it also show number of requests/calls - and auth decisions are related to these requests.. I don't want to get to philosophical here, I think I'm fine with starting with this dashboard and then see how to sort out the MaaS information between dashboards such that we won't end up with 3+ dashboards (we'd also want to add one for quotas most probably) just for MaaS

somya-bhatnagar (approved):
lgtm

PR #1372, opened 2026-08-14
fix: split config.openshift.io rules into OCP-specific ClusterRole

ugiordan:
## Description

`maas-controller-role` contains `config.openshift.io` rules (`apiservers`,
`authentications`) used by maas-controller on OpenShift for TLS profile and
OAuth/OIDC issuer discovery. On xKS those APIs do not exist and the rules are
dead weight.

The problem surfaces in ai-gateway-operator, which deploys MaaS manifests via
server-side apply with `ForceOwnership`. Kubernetes RBAC escalation prevention
requires the operator SA to hold every permission it grants in a ClusterRole it
applies. With `config.openshift.io` in `maas-controller-role`, the operator's
own `manager-role` must carry those rules cluster-wide on both OpenShift and
xKS — even though xKS never uses them.

### Changes

- Add `maas-controller/ocp_rbac.go` (package main, module root) holding only the
  `config.openshift.io` kubebuilder RBAC markers
- Wire two-pass controller-gen in the Makefile `manifests` target:
  - First pass: `paths=./cmd/... paths=./pkg/... paths=./api/...` → produces
    `rbac/clusterrole.yaml` (maas-controller-role, no config.openshift.io)
  - Second pass: `paths=.` (module root only) → produces
    `rbac/ocp/clusterrole_ocp.yaml` (maas-controller-ocp-role, only
    config.openshift.io)
- Remove the `config.openshift.io` markers from `cmd/manager/main.go`,
  `maasauthpolicy_controller.go`, and `tenant_controller.go` — they now live
  exclusively in `ocp_rbac.go`
- `rbac/ocp/` gets its own `kustomization.yaml`; the xKS overlay references
  `../../rbac` which resolves through `rbac/kustomization.yaml` (does not
  include `ocp/`), so OCP resources are automatically excluded from xKS with no
  overlay changes needed
- `verify-codegen` CI check updated to watch both generated files
- Quote `$(CONTROLLER_GEN)` in all four Makefile invocations

### Result

On OpenShift: maas-controller SA gets `config.openshift.io` permissions via
`maas-controller-ocp-role` as before. Both ClusterRoles are now fully generated
— `verify-codegen` will catch any drift.

On xKS: `maas-controller-ocp-role` and its binding are never applied. The
operator SA no longer needs to hold `config.openshift.io` permissions on xKS.

### Follow-up

After this merges, ai-gateway-operator (PR #76) will bump the MaaS commit SHA
and remove `config.openshift.io` from `manager-role`, moving it to the
OpenShift-only overlay where it belongs.

### DIFF
--- a/deployment/base/maas-controller/base/kustomization.yaml
+++ b/deployment/base/maas-controller/base/kustomization.yaml
@@ -4,6 +4,7 @@ kind: Kustomization
 resources:
   - ../crd
   - ../rbac
+  - ../rbac/ocp
   - ../manager
   - ../monitoring
   - ../webhook
--- a/deployment/base/maas-controller/default/kustomization.yaml
+++ b/deployment/base/maas-controller/default/kustomization.yaml
@@ -4,6 +4,7 @@ kind: Kustomization
 resources:
   - ../crd
   - ../rbac
+  - ../rbac/ocp
   - ../manager
   - ../monitoring
   - ../webhook
--- a/deployment/base/maas-controller/rbac/clusterrole.yaml
+++ b/deployment/base/maas-controller/rbac/clusterrole.yaml
@@ -130,15 +130,6 @@ rules:
   - patch
   - watch
 - apiGroups:
-  - config.openshift.io
-  resources:
-  - apiservers
-  - authentications
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
   - discovery.k8s.io
   resources:
   - endpointslices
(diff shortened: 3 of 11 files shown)


lphiri (commented):
This looks good to me in principle. 
Needs a closer look from the repo owners.

liangwen12year (approved):

jland-redhat (approved):
lgtm

PR #1385, opened 2026-08-18
chore: replace blind _wait_reconcile sleeps with condition polls

ishitasequeira:
## Description
Replace ~131 `_wait_reconcile()` calls (fixed 8s time.sleep) across 13 E2E test files with condition-aware polling that exits as soon as the expected state is reached. 

Replacements follow 6 categories:
- After auth policy create → `_wait_for_maas_auth_policy_phase()`
- After subscription create → `_wait_for_maas_subscription_phase()`
- Cleanup/teardown → `_wait_for_cr_absent()` (new helper, 30s timeout, 2s poll)                                                                                                                                                                     
- After CR patch/update → appropriate phase poll                                                                                                                                                                                                
- Namespace discovery/finalizer → `wait_for_finalizer()`                                                                                                                                                                                            
- Negative tests / API key propagation → explicit `time.sleep()` with comment explaining why no K8s condition exists to poll                                                                                                                      

Adds _wait_for_cr_absent() helper to test_helper.py. Removes _wait_reconcile() definition entirely; zero residual callers confirmed via grep.

Fixes https://redhat.atlassian.net/browse/RHOAIENG-82329

## How Has This Been Tested?
Full E2E prow suite (prow_run_smoke_test.sh) on OpenShift cluster (api.isequeir.y5ll.s1.devshift.org:6443) with 7 pytest-xdist parallel workers:

- Result: 189 passed, 45 skipped, 0 failures (17m14s wall clock)
- Verified the one regression we introduced (test_api_key_ignores_subscription_header) was root-caused and fixed — API key propagation requires an explicit sleep since it's an HTTP-level artifact with no K8s resource to poll, and the retry helper only retries on 403/500 status codes, not on 200-with-wrong-data
- Confirmed on main branch that the failing test passes (ruling out pre-existing flake)
- Ran with SKIP_DEPLOYMENT=true against existing cluster state to isolate test-only changes

Risk analysis                                                                                                                                                                                                                                     
- Risk rating: 2                                                                                                                                                                                                                                  
- Why: All changes are in E2E test files only — no production code, no controller logic, no manifests. The replacement patterns are mechanical and use existing poll helpers already battle-tested in CI. The one subtle case (API key propagation sleep) was validated under parallel load. Risk is limited to test flakiness on slow clusters, mitigated by using the existing poll timeouts (60s) which are already tuned for CI.

## Merge criteria:
<!--- This PR will be merged by any repository approver when it meets all the points in the checklist -->
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test synchronization by waiting for specific resources to reach expected states or be fully deleted.
  * Replaced generic reconciliation waits with targeted readiness checks, explicit delays, or reconciliation triggers where appropriate.
  * Added timeout handling when resources do not disappear as expected.
  * Updated test cleanup and setup flows while preserving existing assertions and behavior.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/test/e2e/tests/test_api_keys.py
+++ b/test/e2e/tests/test_api_keys.py
@@ -63,7 +63,7 @@ from test_helper import (
     _scale_controller_up,
     _wait_for_gateway_auth_enforced,
     _wait_for_maas_subscription_phase,
-    _wait_reconcile,
+    _wait_for_cr_absent,
 )
 
 log = logging.getLogger(__name__)
@@ -1417,7 +1417,7 @@ class TestAPIKeySubscriptionPhases:
             _delete_cr("maassubscription", subscription_name, namespace=ns)
             _delete_cr("maasauthpolicy", auth_name, namespace=ns)
             _delete_sa(sa_name, namespace=MODEL_NAMESPACE)
-            _wait_reconcile()
+            _wait_for_cr_absent("maassubscription", subscription_name, namespace=ns)
 
     def test_create_key_for_degraded_subscription(self):
         """API key creation succeeds for Degraded subscription."""
@@ -1438,7 +1438,7 @@ class TestAPIKeySubscriptionPhases:
                 [MODEL_REF, missing_model],
                 users=[sa_user]
             )
-            _wait_reconcile(seconds=10)
+            _wait_for_maas_subscription_phase(subscription_name, expected_phase="Degraded", namespace=ns)
 
             cr = _get_cr("maassubscription", subscription_name, namespace=ns)
             phase = cr.get("status", {}).get("phase")
@@ -1458,7 +1458,7 @@ class TestAPIKeySubscriptionPhases:
             _delete_cr("maassubscription", subscription_name, namespace=ns)
             _delete_cr("maasauthpolicy", auth_name, namespace=ns)
             _delete_sa(sa_name, namespace=MODEL_NAMESPACE)
-            _wait_reconcile()
+            _wait_for_cr_absent("maassubscription", subscription_name, namespace=ns)
 
     def test_create_key_for_failed_subscription(self):
         """API key creation is rejected for Failed subscription to prevent key spam."""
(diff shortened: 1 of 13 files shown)


test/e2e/tests/test_api_keys.py
```diff
@@ -1417,7 +1417,7 @@ def test_create_key_for_active_subscription(self):
             _delete_cr("maassubscription", subscription_name, namespace=ns)
             _delete_cr("maasauthpolicy", auth_name, namespace=ns)
             _delete_sa(sa_name, namespace=MODEL_NAMESPACE)
-            _wait_reconcile()
+            _wait_for_cr_absent("maassubscription", subscription_name, namespace=ns)
```
ahadas, line 1420: can you help me understand this change - as I read it,
previously, we deleted the subscription, AuthPolicy and service-account, and then waited for something to happen (probably Kuadrant to operate on the AuthPolicy removal, I suppose)
now, we delete the subscription, AuthPolicy and service-account, and then wait for the subscription CR to be removed
so it raises two questions to me:
1. why do we need to extra wait for CR absent - it looks like _delete_cr is already operating synchronously with a timeout
2. don't we need to wait for Kuadrant to operate on the AuthPolicy removal before proceeding to the next test?(because if we don't then we can probably just drop that sleep..)

ishitasequeira: Thanks @ahadas for the review. The check itself is a defensive check.

1. "Why do we need extra wait for CR absent — _delete_cr already operates synchronously with a timeout"
   - `_delete_cr` runs `oc delete --timeout=30s`, which waits for the delete command to complete (finalizers run), but doesn't guarantee the resource is fully gone from the API server. With finalizers, the CR may linger briefly after the delete returns. `_wait_for_cr_absent` polls until `_get_cr` returns `None`, confirming it's truly removed. That said, in practice `--timeout=30s` on the delete should be sufficient and the extra wait is defensive.
2. "Don't we need to wait for Kuadrant to operate on the AuthPolicy removal before proceeding?"
   - This is a valid concern. The old `_wait_reconcile()` was a fixed sleep, giving Kuadrant time to process the AuthPolicy removal. The new code waits for the MaaS subscription CR to be absent but doesn't explicitly wait for the AuthPolicy removal to propagate through Kuadrant. Based on the tests, we do not need to access same AuthPolicy and thus waiting on it is a redundant wait.

jland-redhat (approved):
lgtm

PR #1401, opened 2026-08-21
fix: source ServiceMonitor serverName namespace from params

jrhyness:
The kustomize replacement for tlsConfig.serverName was reading metadata.namespace from the metrics Service, which is hardcoded to opendatahub. When AGO deploys to a different namespace (e.g. redhat-ods-applications on RHOAI), the post-build namespace transformer updates metadata.namespace on resources but cannot reach the already-baked serverName string, causing HTTPS scrape hostname verification to fail.

Source the namespace segment from the maas-parameters ConfigMap (data.namespace) instead. AGO already writes ApplicationsNamespace into this key via ApplyParams before kustomize build, so the rendered serverName now reflects the actual deployment namespace.

## How Has This Been Tested?

- Built ai-gateway-operator with local maas manifests
- Deployed rhoai
- Patched ai-gateway-operator
- Observed
```
$ kubectl get servicemonitor maas-controller-metrics -n redhat-ods-applications -o jsonpath='{.spec.endpoints[0].tlsConfig.serverName}'
maas-controller-metrics.redhat-ods-applications.svc
```

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ServiceMonitor TLS configuration to use the configured deployment namespace when constructing the service address.
  * Added an explicit deployment namespace setting to ensure monitoring connections resolve to the correct service endpoint.
  * Updated deployment configuration so namespace changes are consistently applied to monitoring resources.


somya-bhatnagar (approved):
/lgtm

PR #1417, opened 2026-08-31
fix: stop legacy ExternalModel reconciler when inference CRD exists

yossiovadia:
## Summary

- On 3.4→3.5 upgrades with the legacy migration controller, both the maas-controller ExternalModel reconciler and the IPP ExternalModel controller create parallel networking resources for the same model. This causes duplicate HTTPRoutes on the same public path, TRLP ownership conflicts (`SetControllerReference` fails), and policy skew.
- The fix adds a check at the top of the legacy reconciler: if an `inference.opendatahub.io` ExternalModel with the same name exists in the same namespace, tear down the `maas-*` children (Service, ServiceEntry, DestinationRule, HTTPRoute) and return early. The IPP controller owns networking from that point forward.
- Handles pure 3.4 clusters (inference CRD not installed) gracefully via `IsNoMatchError` — the reconciler proceeds normally.
- Respects the existing `opendatahub.io/managed=false` opt-out annotation on child resources during teardown.
- Uses the same unstructured lookup pattern already established in `providers_external.go` for `MaaSModelRef`.

## Risk analysis

- **Risk rating**: 3
- **Why**: Changes the reconcile path for ExternalModel on upgraded clusters. Well-scoped — only affects clusters where both legacy and inference ExternalModels coexist for the same model name. Covered by 3 new unit tests. No RBAC changes. No generated file changes. The teardown path reuses existing `deleteIfExists` and `isManaged` helpers already exercised by existing tests.

## Test plan

- [x] `TestReconcile_SupersededByInference` — verifies legacy children are torn down when inference ExternalModel exists
- [x] `TestReconcile_NoInferenceModel_ProceedsNormally` — verifies reconciler creates children as usual without inference CR
- [x] `TestReconcile_SupersededDoesNotTeardownOptedOut` — verifies `managed=false` children survive teardown
- [x] Full package test suite passes (14/14 tests)
- [x] `golangci-lint` clean
- [ ] E2E: deploy on cluster with both CRD versions, create legacy ExternalModel, run migration, verify maas-* children are removed

## Summary by CodeRabbit

* **New Features**
  * Added automatic transition to inference-based ExternalModels when a matching resource has a published HTTP route.
  * Legacy networking resources are removed during the transition.
  * Resources marked as not managed are preserved.
  * Changes to inference-based ExternalModels now trigger reconciliation.

* **Bug Fixes**
  * Prevented the legacy reconciler from recreating networking resources after supersession.

* **Tests**
  * Added coverage for transitions, normal reconciliation, readiness handling, and opt-out behavior.


yossiovadia:
@coderabbitai review

noyitz:
Reviewed the approach and the diff — this correctly implements the "maas-controller defers to inference" direction, and the happy-path logic plus the three tests are sound.

Flagging two things to address before this is mergeable, **assuming we land on this approach** (the layer is still under discussion with Bartosz/Ishita on [RHOAIENG-89628](https://redhat.atlassian.net/browse/RHOAIENG-89628) — conversion-webhook vs reconciler vs deprecation policy):

**1. The reconciler won't re-trigger when the inference CR appears.**
`SetupWithManager` still only does `For(&maasv1alpha1.ExternalModel{})`. The migration controller creating the `inference.opendatahub.io` ExternalModel fires an event on the *inference* object, which this reconciler doesn't watch — and nothing mutates the legacy CR to requeue it. So `teardownLegacyChildren` only runs on the next legacy-EM event or the periodic resync (~10h default), leaving the duplicate HTTPRoutes (and the TRLP ownership conflict) live in between. Suggest adding a `.Watches(&inference ExternalModel, handler.EnqueueRequestsFromMapFunc(...))` that maps to the same NamespacedName, so teardown fires promptly when migration completes.

**2. Teardown should gate on the inference route being Ready, not just existing.**
`isSupersededByInference` returns true on mere existence of the inference CR. If it exists but its HTTPRoute isn't programmed yet (`status.httpRouteName` empty), we delete the working `maas-*` route before the replacement is live → a brief routing/auth gap during cutover. The MaaSModelRef handler already waits for `status.httpRouteName` before switching (`providers_external.go:83-95`); teardown should mirror that and only supersede once `status.httpRouteName` is set.

Both are additive to the current change — no rework of what's here.


[RHOAIENG-89628]: https://redhat.atlassian.net/browse/RHOAIENG-89628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

noyitz:
Pushed `4a459a54` addressing both review items above:

1. **Watch trigger** — `SetupWithManager` now watches `inference.opendatahub.io` ExternalModel (CRD-gated via the REST mapper, so pure-3.4 clusters where the CRD is absent are skipped and don't fail cache sync) and maps each event to the same name/namespace. Teardown now fires promptly when migration creates the inference CR, instead of waiting for the periodic resync. Reconcile no-ops if no legacy CR exists at that key, so native inference-only models are untouched.
2. **Readiness gating** — `isSupersededByInference` now requires `status.httpRouteName` to be set, so the legacy `maas-*` route is only removed once the inference route is programmed (mirrors the MaaSModelRef handler). Avoids a routing gap during cutover.

Tests updated; added `TestReconcile_InferenceNotReady_ProceedsNormally`. build / vet / lint / race all green.

Still contingent on the reconciler-vs-webhook decision on [RHOAIENG-89628](https://redhat.atlassian.net/browse/RHOAIENG-89628) — this just makes the reconciler-defer path merge-ready if that's the direction.


[RHOAIENG-89628]: https://redhat.atlassian.net/browse/RHOAIENG-89628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

somya-bhatnagar:
Validated the full 3.4→3.5 upgrade flow on OpenShift 4.21 — zero manual intervention after the upgrade was applied.

**Images used:**
- 3.4 baseline: `registry.redhat.io/rhoai/odh-maas-controller-rhel9@sha256:e1dfec...` (from RHOAI 3.4.4 CSV `RELATED_IMAGE_ODH_MAAS_CONTROLLER_IMAGE`)
- 3.5 / this PR: `quay.io/rh-ee-sbhatnag/maas-controller:pr-1417`
- Migration controller: `quay.io/opendatahub/odh-ai-gateway-payload-processing:odh-stable`

**Flow confirmed automatically:**
1. RHOAI 3.4 controller created legacy networking for `maas.opendatahub.io ExternalModel`
2. IPP `legacy-migration` controller auto-created `inference.opendatahub.io ExternalModel` + `ExternalProvider`
3. IPP `external-model-controller` set `status.httpRouteName`
4. PR fix detected the signal and tore down all legacy networking

**One request:** Once an ExternalModel has been migrated, could we reflect that in its status? Something like:

```yaml
status:
  phase: Migrated
  message: "Model migrated to inference.opendatahub.io ExternalModel: test-model"
```

After migration the `maas.opendatahub.io ExternalModel` CR intentionally stays on the cluster (it owns the inference objects via OwnerRef), so without a status signal there's no easy way to tell from the console whether a given ExternalModel is still active or already migrated.

SB159:
Validated the 3.4 → 3.5 upgrade flow on OCP 4.21.

Images used

3.4 baseline: quay.io/rhoai/rhoai-fbc-fragment:rhoai-3.4@sha256:8ba5d46f5ac846b2a8b86ed936031ce82877589c6a5cb88fe48e54b02b50dedd (cluster pre-upgrade)
3.5: quay.io/rh-ee-sbhatnag/maas-controller:pr-1417
Flow validated

Pre-upgrade (3.4): Legacy maas.opendatahub.io ExternalModel deployed with MaaSModelRef, auth policy, and subscription; maas-controller created legacy networking.
Upgrade: Cluster upgraded 3.4 → 3.5 on the same cluster.
Post-upgrade (3.5): inference.opendatahub.io ExternalModel programmed (status.httpRouteName set); inference HTTPRoute present; MaaSModelRef and subscription survived.
All upgrade tests passed.

+1 on this — after migration the legacy maas.opendatahub.io ExternalModel stays on the cluster, so a status.phase: Migrated (or similar) would make it much easier to see in the console which models are already on inference.

jland-redhat (approved):
Lgtm

PR #1447, opened 2026-09-07
fix: reject API key authentication on key mint

jland-redhat:
## Summary

- Add gateway `deny-api-key-mint` AuthPolicy rule so `POST /v1/api-keys` rejects callers authenticated with an existing `sk-oai-*` credential
- Reject the same path in `maas-api` `CreateAPIKey` (covers direct-to-pod bypass) with HTTP 403
- Add unit and e2e coverage for nested mint denial

## Test plan

- [x] `go test ./internal/api_keys/... -run TestCreateAPIKey_RejectsAPIKeyAuth`
- [x] `go test ./pkg/controller/maas/... -run TestBuildGatewayAuthPolicySpec_DenyAPIKeyMint`
- [ ] E2E: `test_negative_security.py::TestHeaderSpoofing::test_api_key_cannot_mint_new_api_key`

## Risk analysis

**Risk rating**: 2

**Why**: Behavior change is narrowly scoped to `POST /v1/api-keys` when the caller presents an API key. Legitimate minting with OpenShift/OIDC tokens is unchanged. Gateway and maas-api both enforce the rule; unit tests cover both layers. E2E validates the gateway path in CI when security tests run.

## Summary by CodeRabbit

- **Security**
  - API keys can no longer be used to create additional API keys.
  - Attempts to mint keys with an existing API key are rejected with a 403 Forbidden response and do not expose key material.
  - The restriction applies to API keys supplied through supported authentication headers.

### DIFF
--- a/maas-api/internal/api_keys/handler.go
+++ b/maas-api/internal/api_keys/handler.go
@@ -196,12 +196,39 @@ type CreateAPIKeyRequest struct {
 	Labels       map[string]string `json:"labels,omitempty"`       // Structured key-value pairs for API key metadata
 }
 
+// rejectAPIKeyMintAuth blocks key minting when the caller authenticated with an existing
+// API key. Minting requires a live OpenShift or OIDC identity token.
+func (h *Handler) rejectAPIKeyMintAuth(c *gin.Context) bool {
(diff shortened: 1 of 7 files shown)


jland-redhat:
@coderabbitai review

PR #1453, opened 2026-09-08
docs: maas flow control design RHAISTRAT-2701

pierDipi:
Introduce requestPriority independently of subscription-selection priority so operators can configure inference scheduling without changing which subscription is selected for a request.

Use the AI tenant as the fairness identity and the selected subscription as the logical objective. This enables configured fairness policies to share dispatch opportunities within a priority band while subscriptions express different scheduling tiers.

Have the subscription controller discover active pools through LLMInferenceService status and reconcile one InferenceObjective per subscription/pool pair. Inject matching objective names only after authentication and authorization so clients cannot assert their own tier.

Leave objective resources absent when requestPriority is unset, relying on the scheduler's default priority of zero. Always inject trusted fairness and objective headers in both canonical and legacy forms. Preserve explicit zero and negative priorities, and reconcile changes without recreating subscriptions.

Define operator UI visibility and editing, header-version compatibility, and lifecycle validation so configured priority can be inspected and maintained as models and subscriptions change.

Rich view document at https://github.com/pierDipi/models-as-a-service/blob/flow-control/docs/content/architecture-internals/flow-control.md

Design doc for https://redhat.atlassian.net/browse/RHAISTRAT-2701

## Summary by CodeRabbit

- **Documentation**
  - Added a proposed flow-control design for managing request priorities and fair tenant allocation.
  - Documented support for canonical and legacy request metadata, pool discovery, and compatibility considerations.
  - Described scheduling policies, trusted request classification, pricing, security considerations, and validation plans.
  - Clarified the distinction between subscription-selection priority and request-scheduling priority, including handling of unset and explicit zero values.

### DIFF
--- a/docs/content/architecture-internals/flow-control.md
+++ b/docs/content/architecture-internals/flow-control.md
@@ -1,0 +1,438 @@
+---
+title: MaaS Subscription Flow Control Design
+description: Proposed integration of subscription request priorities with llm-d flow control.
+---
+
+# MaaS Subscription Flow Control Design
+
+!!! note "Proposed design"
+    This document describes planned behavior and follow-up extensions. It is not documentation of an implemented feature.
+
+**Author:** Pierangelo Di Pilato
+
+**Date:** 2026-09-08
+
+**Scope:** Models-as-a-Service, AI Gateway, and llm-d request scheduling
+
+## Overview
+
+Integrate MaaS with Gateway API Inference Extension flow control. Use the AI tenant as the fairness key and the selected
+MaaS subscription as the logical objective key. Users authorized to manage subscriptions configure a subscription's
+request priority. When explicitly configured, the MaaS subscription controller reconciles an `InferenceObjective` for
+each distinct `InferencePool` backing the models in that subscription. Otherwise, requests use the scheduler's default
+priority without a generated objective.
+
+## Motivation
+
+MaaS subscriptions already define model access, token rate limits, and billing metadata. Token rate limits constrain
+consumption over time, but do not determine which admitted requests should be dispatched first when inference capacity
+is scarce. Flow control adds request queuing, priority, and fairness at the inference scheduler.
+
+Non-MaaS Gateways already inject `x-gateway-inference-fairness-id` and `x-gateway-inference-objective` through
+AuthPolicy templates. Their defaults use the cluster issuer for fairness and the caller's service-account namespace, or
+an authenticated-user fallback, for the objective. MaaS needs these headers to reflect its tenant and subscription
+boundaries. Subscription administrators should not need to discover pools or manually maintain objectives in model
+namespaces.
+
+## Goals
+
+* Configure scheduling priority once per subscription and apply it to every eligible model in that subscription.
+* Group requests from the same AI tenant for fairness within each priority band.
+* Reuse the existing flow-control headers and `InferenceObjective` API.
+* Keep generated objectives synchronized with subscriptions and observed model topology.
+* Derive flow-control identity from authorized tenant, subscription, and model context.
+* Let authorized operators view, set, change, and clear request priority through the subscription UI.
+
+## Non-Goals
+
+* Replace MaaS authorization, subscription selection, token rate limits, or billing.
+* Introduce or change request `service_tier` handling. Its interaction with subscription selection, authorization, and
+  pricing requires a separate proposal; this integration derives scheduling priority from the selected subscription's
+  `requestPriority`, without a `service_tier` override.
+* Introduce a new scheduler, per-user fairness, or per-model priority overrides within a subscription.
+* Provide a global capacity allocation across independent pools or scheduler instances, or guarantee latency or minimum
+  capacity across priority bands.
+* Add flow control to external providers or model routes without an `InferencePool`.
+* Change the non-MaaS Gateway defaults.
+
+## Proposed Design
+
+### Subscription priority and fairness
+
+Add an optional `MaaSSubscription.spec.requestPriority` field as the desired `InferenceObjective.spec.priority`. Use a
+signed `int32`, matching the objective API, and preserve the distinction between an omitted value and an explicit `0`
+(for example, using `*int32` in the Go API). Do not apply a CRD default that fills in this field. Higher values receive
+higher scheduling priority; negative values represent lower-than-default priority. Validate and preserve this value
+consistently across the subscription API and objective reconciliation.
+
+The existing `spec.priority` continues to control automatic subscription selection. `spec.requestPriority` controls
+request scheduling after a subscription has been selected. Updating either field does not change the other field's
+meaning, and request priority does not influence subscription selection. Explicit subscription selection and API-key
+subscription binding continue to follow the existing authorization rules.
+
+Subscriptions without `spec.requestPriority` create no `InferenceObjective`, but both canonical and legacy header pairs
+are still injected. The objective header carries the generated name for the subscription/pool pair. With no matching
+objective, llm-d applies scheduler priority `0`, regardless of subscription-selection priority. An explicit
+`spec.requestPriority: 0` creates an objective with priority `0` under that same name, following the same path as any
+other explicit value.
+
+For example, a subscription can have `spec.priority: 100` to make it preferred during automatic selection and
+`spec.requestPriority: 10` to schedule its requests at priority `10`.
+
+The documented default objective header value is the generated subscription/pool objective name, even when request
+priority is omitted. In that case the resource is intentionally absent and the scheduler defaults to priority `0`. When
+request priority is configured, the same header value matches the generated objective in the target pool's namespace.
+
+| Flow-control input                 | MaaS source                                                              | Meaning                                                                                                |
+|------------------------------------|--------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
+| `x-llm-d-inference-fairness-id`    | Stable, unambiguous identity of the resolved `AITenant`                  | All callers and subscriptions of that tenant share a fairness identity.                                |
+| `x-llm-d-inference-objective`      | Generated objective name for the authorized subscription and target pool | Always injected; resolves the configured priority, or scheduler priority `0` when no objective exists. |
+| `InferenceObjective.spec.priority` | Explicit `MaaSSubscription.spec.requestPriority`                         | Applies the same request priority to every pool covered by the subscription.                           |
+
+MaaS emits the canonical `x-llm-d-*` names defined by llm-d Router's `FlowFairnessIDKey` and `ObjectiveKey` constants
+alongside their legacy aliases to support older deployments. Each alias carries exactly the same trusted value as its
+canonical counterpart:
+
+| Canonical header                | Legacy header                     | Value                                      |
+|---------------------------------|-----------------------------------|--------------------------------------------|
+| `x-llm-d-inference-fairness-id` | `x-gateway-inference-fairness-id` | Resolved AI tenant identity                |
+| `x-llm-d-inference-objective`   | `x-gateway-inference-objective`   | Generated subscription/pool objective name |
+
+The scheduler groups flows by `(fairness ID, priority)`. Two subscriptions belonging to one tenant at the same priority
+share a flow; creating additional subscriptions does not give that tenant additional fairness shares at that priority.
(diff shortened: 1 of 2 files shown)


docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+* Provide a global capacity allocation across independent pools or scheduler instances, or guarantee latency or minimum
+  capacity across priority bands.
+* Add flow control to external providers or model routes without an `InferencePool`.
+* Change the non-MaaS Gateway defaults.
+
+## Proposed Design
+
+### Subscription priority and fairness
+
+Add an optional `MaaSSubscription.spec.requestPriority` field as the desired `InferenceObjective.spec.priority`. Use a
```
jland-redhat, line 62: Could we call this `defaultRequestPriority`?

I know it does not make that much sense in the current design but just thinking about future designs when that priority may change based on defined limits.

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+
+Non-MaaS Gateways already inject `x-gateway-inference-fairness-id` and `x-gateway-inference-objective` through
+AuthPolicy templates. Their defaults use the cluster issuer for fairness and the caller's service-account namespace, or
+an authenticated-user fallback, for the objective. MaaS needs these headers to reflect its tenant and subscription
+boundaries. Subscription administrators should not need to discover pools or manually maintain objectives in model
+namespaces.
+
+## Goals
+
+* Configure scheduling priority once per subscription and apply it to every eligible model in that subscription.
```
jland-redhat: More of a wording thing here but should we just say "every eligible inference request"?

Since it is something that I assume we are applying on every request and we won't really care what model gets applied

jland-redhat (commented):
Have not had a chance to make it through the whole doc. 

This was going to be more than I was originally thinking (I was just going to expose the header functionality and not actually create the InferenceObject) but if we can get this done it would be great. Just means extra work on the controller side


Overall it makes sense to me, I will say this is going to make doing the soft limitings interesting. And I don't know if the document covers this but I do think it would be good to have feedback that this only works for InternalModels.

Note: that this pattern does mean that ExternalModels that point to llm-d instances living on other clusters won't be able to follow this pattern. But I think that is ok for now.

jland-redhat:
Oh and I like the pattern of creating proposals on the repository, I may start getting us to use this more often. 

But could we move it to `docs/proposals` and not include it in the mkdocs as that is what creates our official upstream docs?

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+policy determines how flows share dispatch opportunities. The default global-strict policy does not provide tenant
+fairness; deployments requiring fair sharing must explicitly select an appropriate fairness policy.
+
+### Pool discovery and objective reconciliation
+
+The MaaS subscription controller owns the desired objectives:
+
+1. Resolve the subscription's AI tenant using the existing tenant namespace association and validate that each
+   referenced model is exposed through that tenant's Gateway.
+2. Follow each `spec.modelRefs[]` entry to its `MaaSModelRef`. For `spec.modelRef.kind: LLMInferenceService`, read the
```
jland-redhat, line 115: Should we have some sort of feedback so the user understands that this does not work for ExternalModels (possibly a status that just denotes any model that is ignored).

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+[header constants](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/metadata/consts.go). MaaS emits both
+canonical names and their legacy aliases so existing deployments that consume the older names receive the same
+scheduling identities. Newer routers prefer canonical values, while older routers consume the legacy values. Both pairs
+must be derived from the same trusted metadata and injected together, including when request priority is unset. Header
+compatibility does not replace the requirement for the objective API and scheduler flow-control support.
+
+Report per-model flow-control reconciliation state in subscription status. A missing observed pool for a service that is
+still reconciling is pending and retried; an external model or a route deliberately lacking an inference scheduler is
+not applicable. When request priority is unset, absence of an objective is expected and must not be reported as an
+objective reconciliation failure. Objective creation or update failures for explicit request priorities must be visible
```
jland-redhat, line 214: We have a "degraded" status that I think we could use for this.

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+without preventing reconciliation of the subscription's other models.
+
+The existing scheduler defaults to priority `0` when an objective cannot be resolved. Controller and informer
+propagation are asynchronous, so updates may temporarily use the old priority or that default. This integration does not
+promise atomic changes across all pools. Missing objectives must not bypass MaaS access control or token rate limits,
+and MaaS must not report successful flow-control reconciliation while required objectives are missing.
+
+Deployments without the optional inference APIs must continue to support existing MaaS functionality. Enabling flow
```
jland-redhat, line 222: I know this is overkill, but can we add. "Testing should be included to validate MaaS functions without Optional Interface APIs and prevent regressions"

Just cause I want to capture that and I think that test may be tricky to implement with our current setup.

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+merely place requests later in the queue. This behavior must be explained to subscription operators.
+
+If SLO deadline ordering is enabled, `x-llm-d-slo-ttft-ms` and its legacy alias can influence scheduling order. A
+follow-up must define trusted derivation or validation and bounds for these values before exposing subscription SLO
+controls. Protecting objective and fairness headers does not govern additional scheduling inputs. Tenant
+request-priority ranges complement scheduler policies by restricting which bands a subscription may select.
+
+## Priority-Tier Pricing and Chargeback
+
+The existing per-model `billingRate.perToken` on a subscription can incorporate the price of its priority tier. For
```
jland-redhat, line 297: This is a good idea but just as a note billing or metering is not currently supported. I believe this is only used in a DP capacity curently.

docs/content/architecture-internals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+> proceed without tenant-level priority ranges. When cross-tenant pool sharing is introduced, the
+> [tenant request-priority range follow-up](#follow-up-tenant-request-priority-ranges) should define and enforce these
+> boundaries. Subscription priorities still govern contention within each tenant.
+
+The controller requires read access to services and pools and permission to manage objectives in model namespaces.
+Preserve tenant-to-model Gateway validation before writing these resources, and carry the required RBAC changes into
+both the AI Gateway and ODH parent operators. Restrict direct objective edits to the platform's authorized
+administrators. Internal headers and status mappings must not contain API keys or other credentials.
+
+## Follow-up: Tenant Request-Priority Ranges
```
jland-redhat, line 341: I can add this follow up later but we are also hoping to eventually expand this to have priority set based on actual token "soft-limits".

So a user goes over 10,000 we would need to create an objective with a lower priority. 

Note this could exacerbate some of the items noted below, but I do believe it is a pretty high value item.

jland-redhat (commented):
Question, when originally thinking about this my assumption was that the user would be creating the `InferenceObjective` and MaaS would simply be adding the header.

This is obviously a much  higher value integration but would that be a pattern we would still want to support here? Where we can let the user setup the  InferenceObjectives and we simply populate the header? 

Maybe if we do we want custom headers to just be their own path but wanted to throw that out there.

docs/content/architecture-internals/flow-control.md
pierDipi: shouldn't the case of changing priorities be handled by changing subscription selection?
 
if you switch to a different priority, billing rate for a given subscription is not applicable anymore, let's say when soft limits are breached, switch to a lower priority, then billing rate should be lower but that it's essentially switching to a completely different billing rate and priority.

docs/content/architecture-internals/flow-control.md
pierDipi: does this mean this is not the API / design that will move to stable?

pierDipi:
> This is obviously a much higher value integration but would that be a pattern we would still want to support here? Where we can let the user setup the InferenceObjectives and we simply populate the header?

how would this work in practice? I noted in the alternatives that creating inf objectives manually means a lot of work for the user (you need one inf objective combination for each model, subscription and tenant combination).

In addition, that would create an implicit contract that users will start relying on while currently implementation details would be hidden and we have more flexibility of adapting them in the future.

docs/content/architecture-internals/proposals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+merely place requests later in the queue. This behavior must be explained to subscription operators.
+
+If SLO deadline ordering is enabled, `x-llm-d-slo-ttft-ms` and its legacy alias can influence scheduling order. A
+follow-up must define trusted derivation or validation and bounds for these values before exposing subscription SLO
+controls. Protecting objective and fairness headers does not govern additional scheduling inputs. Tenant
+request-priority ranges complement scheduler policies by restricting which bands a subscription may select.
+
+## Priority-Tier Pricing and Chargeback
+
+The existing per-model `billingRate.perToken` on a subscription can incorporate the price of its priority tier. For
```
jland-redhat: Still in discussion. Right now this is just metadata that we capture

docs/mkdocs.yml
```diff
@@ -113,6 +113,7 @@ nav:
     - Controller Architecture: architecture-internals/controller-architecture.md
     - Reconciliation Flow: architecture-internals/reconciliation-flow.md
     - Authentication Internals: architecture-internals/authentication-internals.md
+    - Flow Control Design: architecture-internals/flow-control.md
```
pierDipi:
```suggestion
```

removing the doc from the website page in the PR so that it doesn't show up for now until implemented

jland-redhat (approved):
lgtm

docs/content/architecture-internals/proposals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+
+With in-flight eviction enabled, the existing sheddable filter admits only negative-priority victims, and reclamation
+requires victim priority to be below the blocked demand's priority. Victims are ordered by lowest priority, then newest
+dispatch time. Setting a negative `requestPriority` therefore permits interruption under that configuration; it does not
+merely place requests later in the queue. This behavior must be explained to subscription operators.
+
+If SLO deadline ordering is enabled, `x-llm-d-slo-ttft-ms` and its legacy alias can influence scheduling order. A
+follow-up must define trusted derivation or validation and bounds for these values before exposing subscription SLO
+controls. Protecting objective and fairness headers does not govern additional scheduling inputs. Tenant
+request-priority ranges complement scheduler policies by restricting which bands a subscription may select.
```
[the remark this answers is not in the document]
pierDipi: `x-llm-d-slo-ttft-ms` is informational only for sheddable requests (priority < 0) and it's meant to be provided by clients

docs/content/architecture-internals/proposals/flow-control.md
```diff
@@ -0,0 +1,438 @@
...
+3. Discover the active pool through `LLMInferenceService.status.router.scheduler.inferencePool`, including its group,
+   kind, name, and namespace. Use this observed reference for both managed pools and explicitly referenced pools; do not
+   infer a generated pool name from the service name or assume that the subscription and pool share a namespace.
+4. When `spec.requestPriority` is explicitly set, reconcile one objective per distinct `(subscription, pool)` pair in
+   the pool's namespace, with `spec.poolRef.name`
+   pointing to that pool and `spec.priority` copied from the subscription's `spec.requestPriority`. Multiple model
+   references resolving to the same pool share that objective.
+5. Reconcile changes to request priority, model membership, tenant association, and observed pool references. Remove
+   obsolete objectives when request priority is cleared, a model leaves the subscription, its pool changes, or the
+   subscription is deleted. With request priority omitted, no objectives are desired.
```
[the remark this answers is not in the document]
pierDipi: As described at line 135+, InferenceObjective is derived from Tenant + Subscription + Pool (Pool ~ model for now)

PR #1455, opened 2026-09-09
fix: retry maasmodelref status update conflicts

ishitasequeira:
## Summary

- return MaaSModelRef status update failures to controller-runtime
- retry optimistic-lock conflicts instead of logging and discarding them
- verify status retry and model-alias sibling convergence

## Testing

- `make -C maas-controller test`

## Risk analysis

- **Risk rating**: 4
- **Why**: This changes production reconciliation error handling. The behavior is covered by focused unit tests and the full race-enabled controller suite, but API-server status-conflict timing is unlikely to be exercised deterministically by the default Prow smoke path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Status update failures are now reported and retried instead of being silently ignored.
  - Reconciliation preserves the original failure when status updates also fail, improving error visibility.
  - Changes to model aliases now correctly reconcile both previously associated and newly associated collision records.

- **Tests**
  - Added coverage for status-update conflicts, retry behavior, and alias changes affecting related records.


ishitasequeira:
@coderabbitai review

ishitasequeira:
Backfilled Jira: https://redhat.atlassian.net/browse/RHOAIENG-96456

PR #1471, opened 2026-09-10
fix: respect captureUser setting for usage logs

ahadas:
## Description
Previously, user identity was included in access logs by default. Now, user identity is included in access logs only when the captureUser setting, which is disabled by default, is set.

## How Has This Been Tested?
Tested manually

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Tenant telemetry settings can control whether user IDs are included in usage logs.
  * When enabled, usage logs include the tenant’s user identity and service namespace.

* **Changes**
  * User IDs are no longer captured in usage logs by default.
  * Subscription, group, and key name information continues to be included in usage logs.


deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
```diff
@@ -2,12 +2,14 @@
...
-#            429 local replies) or from header_to_metadata captured before
-#            payload-processing ext_proc strips the headers. Other identity
-#            fields remain on auth identity FILTER_STATE
-#            (wasm.kuadrant.auth.identity.*).
+# Identity:  subscription and groups from X-MaaS-Subscription / X-MaaS-Group
+#            (still present on Limitador 429 local replies) or from
+#            header_to_metadata captured before payload-processing ext_proc
+#            strips the headers. Other identity fields remain on auth identity
+#            FILTER_STATE (wasm.kuadrant.auth.identity.*).
+#            user_id (X-MaaS-Username) is NOT captured by default; it is
```
tgitelman, line 10: Good — documents that user_id is runtime-patched. Only place in this PR that explains the behavior. CRD/docs still say metrics-only (see follow-up files below).

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
```diff
@@ -458,9 +450,7 @@ spec:
                 # Prefer metadata captured by header_to_metadata after auth.
                 # Fall back to request headers when metadata namespace is unavailable
                 # (e.g., Limitador 429 before header_to_metadata runs).
-                - key: user_id
-                  value:
-                    string_value: '%CEL("envoy.filters.http.header_to_metadata" in metadata.filter_metadata ? metadata.filter_metadata["envoy.filters.http.header_to_metadata"]["user_id"] : request.headers["x-maas-username"])%'
+                # user_id is injected here by the tenant reconciler when captureUser=true.
```
tgitelman, line 453: Fine as a pointer for anyone reading the manifest. The actual user_id attr only appears in the applied CR when captureUser=true — worth knowing if you're grepping the repo for user_id and wonder why it's missing here.

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
```diff
@@ -140,6 +140,15 @@ func (r *TenantReconciler) applyUsageLogsEnvoyFilter(
...
 
+	captureUser := tenant.Spec.Telemetry != nil &&
+		tenant.Spec.Telemetry.Metrics != nil &&
+		ptr.Deref(tenant.Spec.Telemetry.Metrics.CaptureUser, false)
+	if captureUser {
+		if err := tenantreconcile.PatchUsageLogsUserID(ef); err != nil {
+			return false, fmt.Errorf("patch user_id into EnvoyFilter: %w", err)
+		}
+	}
+
```
tgitelman, line 151: Logic looks correct: read base manifest → patch user_id only when captureUser → then collector address + workloadSelector. Reconcile re-reads the base file each time so toggling the flag off drops user_id again. 

Minor: same function still os.ReadFiles the full manifest every reconcile (pre-existing). If you ever refactor, embed base + addon once at startup.

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
```diff
@@ -19,6 +19,90 @@ func PatchUsageLogsEnvoyFilterWorkloadSelector(ef *unstructured.Unstructured, ga
 	return nil
 }
 
+// PatchUsageLogsUserID injects user_id capture into an EnvoyFilter loaded from the
+// base manifest (which omits user_id by default). It:
+//  1. Prepends the X-MaaS-Username → user_id rule to every header_to_metadata
+//     HTTP_FILTER configPatch (both the wasm and wasmplugin variants).
+//  2. Prepends a user_id attribute to the NETWORK_FILTER OTel access log
+//     attributes list.
+func PatchUsageLogsUserID(ef *unstructured.Unstructured) error {
```
tgitelman, line 28: Just suggesting while reading the code.
Consider sourcing these fragments from a small YAML addon (envoy-otel-user-id-capture.yaml) instead of map[string]any below — easier to review next to the big EnvoyFilter. Load once via embed/sync.Once, merge here.

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
```diff
@@ -19,6 +19,90 @@ func PatchUsageLogsEnvoyFilterWorkloadSelector(ef *unstructured.Unstructured, ga
...
+	configPatches, found, err := unstructured.NestedSlice(ef.Object, "spec", "configPatches")
+	if err != nil {
+		return fmt.Errorf("read configPatches: %w", err)
+	}
+	if !found {
+		return errors.New("configPatches not found")
+	}
+
+	userIDRule := map[string]any{
```
tgitelman, line 37: My suggestion/code styling.
This is the Envoy config reviewers care about, but it's buried in Go. A 15-line YAML file would diff cleaner than nested maps + raw CEL string.

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
```diff
@@ -19,6 +19,90 @@ func PatchUsageLogsEnvoyFilterWorkloadSelector(ef *unstructured.Unstructured, ga
...
+			if err := unstructured.SetNestedSlice(al, attrs, "typed_config", "attributes", "values"); err != nil {
+				return fmt.Errorf("set attributes.values: %w", err)
+			}
+			accessLogs[0] = al
+			if err := unstructured.SetNestedSlice(patch, accessLogs, "patch", "value", "typed_config", "access_log"); err != nil {
+				return fmt.Errorf("write back access_log: %w", err)
+			}
+			configPatches[i] = patch
+		}
+	}
```
tgitelman, line 101: If manifest drifts (rename header_to_metadata.identity, reshape OTel patch), this loop skips silently and reconcile still succeeds — captureUser=true would do nothing. Suggest httpPatched / networkPatched flags and return an error if either target wasn't updated when this function is called.

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
```diff
@@ -19,6 +19,90 @@ func PatchUsageLogsEnvoyFilterWorkloadSelector(ef *unstructured.Unstructured, ga
...
+
+	return unstructured.SetNestedSlice(ef.Object, configPatches, "spec", "configPatches")
```
tgitelman, line 103: Same concern — success here doesn't prove both HTTP + NETWORK patches actually ran. Fail loud beats silent user-toggle breakage.

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter_test.go
```diff
@@ -303,6 +303,86 @@ func TestTenantEnsureUsageLogsEnvoyFilter(t *testing.T) {
...
+				for _, v := range values {
+					entry, _ := v.(map[string]any)
+					if entry["key"] == "user_id" {
+						return true
+					}
+				}
+			}
+		}
+		return false
+	}
```
tgitelman, line 335: Helper only scans NETWORK_FILTER for key: user_id. It doesn't check that header_to_metadata got the X-MaaS-Username rule (the bit that matters before ext_proc strips headers). Enabled test could pass with OTel attr only and broken header capture.

Suggest also asserting:

both header_to_metadata.identity patches have X-MaaS-Username in request_rules
CEL string_value matches expected
count: 2 HTTP patches + 1 OTel attr when enabled, 0 when disabled

tgitelman:
Follow up suggestion:
I didn’t find any pytest under test/e2e/ that enables usageLogging, sends inference, and asserts user_id in usage-log output. The only E2E captureUser usage I see is test_aitenant_lifecycle.py L461–515 (telemetry field migration into MaasTenantConfig).
This PR adds unit coverage in tenant_usage_logs_envoyfilter_test.go (EnvoyFilter shape). 

My suggestion is to add new test/e2e/ case (toggle usageLogging + captureUser → inference → check collector/receiver). Optional follow-up, not required to merge. :)

tgitelman:
Docs / CRD follow-up (not blocking)

This PR makes captureUser gate user_id in usage logs, but the API and docs still describe it as metrics-only. Worth aligning in a small follow-up (or extending this PR):

maas-controller/api/maas/v1alpha1/tenant_types.go — TenantMetricsConfig.CaptureUser comment, L109–111: says metrics only; should mention usage-log user_id when usageLogging is on.

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maastenantconfigs.yaml — spec.telemetry.metrics.captureUser description, L88–93: same; regenerate after tenant_types.go change.

deployment/base/maas-controller/crd/bases/maas.opendatahub.io_tenants.yaml — captureUser description, L128–133: same if Tenant still embeds TenantMetricsConfig.

docs/content/observability/setup.md — § Installation → Option 1: Operator-Managed, example L69–75 (captureUser: false # GDPR) and bullet L87: metrics-only context; no usage logs.

docs/content/install/maas-setup.md — § MaasTenantConfig CR field table, L192: “Include user labels on metrics” — incomplete.

docs/content/reference/crds/tenant.md — § TenantMetricsConfig table, L65: “Add a user dimension to telemetry metrics” — incomplete.

Without that update, operators may expect user-scoped usage logs or the admin user dropdown to work while captureUser stays false. That’s intentional: logs won’t carry user_id and those views stay empty until both usageLogging and captureUser are enabled.

deployment/components/observability/usage-logs/envoy-otel-access-log.yaml
ahadas: right

maas-controller/pkg/controller/maas/tenant_usage_logs_envoyfilter.go
ahadas: > Minor: same function still os.ReadFiles the full manifest every reconcile (pre-existing). If you ever refactor, embed base + addon once at startup.

that's unrelated to this PR - @tgitelman it looks like Claude suggests a follow up to a part that was introduced [in your PR](https://github.com/opendatahub-io/models-as-a-service/commit/2da6a5b06047230cf6a9fddf1257e959fef48dcb), would you like to address this (separately)?

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
ahadas: nah, I find it more complicated since the definition of the EnvoyFilter would then scattered across more resources

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
ahadas: looks like this comment duplicates the previous one

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
ahadas: if something like this ever happens, the unit tests would fail

maas-controller/pkg/platform/tenantreconcile/usage_logs_envoyfilter.go
ahadas: same answer - no need to check this on runtime, the yaml is part of the container image and the logic is verified by unit tests

jland-redhat (approved):
lgtm

PR #1477, opened 2026-09-10
fix: add the fallback logic for name pulling

yu-teo:
https://redhat.atlassian.net/browse/RHOAIENG-90254

## Description
When a user sets `spec.model.name` on an LLMInferenceService (e.g., **meta-llama/llama-3.2-1b-instruct**), the MaaS subscription UI still shows the object name (`llama-3-2-1b-instruct`) as the `Model ID` because ResolveModelAlias only reads from `status.addresses[*].models[0].name`, which KServe **may not** populate.

As a fix, I added a fallback in `ResolveModelAlias`, so that when `status.addresses` doesn't contain the model name, it now constructs the canonical alias from `spec.model.name` (or `metadata.name` if unset) in the `publishers/{namespace}/models/{name}` format, instead of returning empty.

## How Has This Been Tested?
Ran maas-controller test suite locally. 
Will try to perform cluster testing too

## Merge criteria:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model alias resolution for inference services.
  - Models now receive a canonical alias when an address does not provide a model name, using the configured model name or service name as a fallback.
  - Preserved the existing preference for model names reported through service status.

- **Tests**
  - Added coverage for status-based aliases, configured-name fallbacks, service-name fallbacks, and precedence behavior.

### DIFF
--- a/docs/content/configuration-and-management/quota-and-access-configuration.md
+++ b/docs/content/configuration-and-management/quota-and-access-configuration.md
@@ -182,6 +182,22 @@ TRLP=$(kubectl get tokenratelimitpolicy -n ${MODEL_NS} -l maas.opendatahub.io/mo
     
     Authentication, subscription validation, and model access controls still apply to all endpoints — only token-based rate limiting enforcement is affected.
 
+### Model server must include usage data
+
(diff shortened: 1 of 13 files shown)


jrhyness (approved):
lgtm

PR #1491, opened 2026-09-14
docs: design doc for responses and guardrails for MaaS

pierDipi:
Define a coherent end-to-end architecture for enabling Responses and enforcing guardrails across tenant, model, and subscription scopes. Clarify resource ownership and policy execution so teams can evolve the control plane without weakening enforcement or coupling the design to Envoy.

Establish a shared direction to break into scoped, independently reviewed deliverables over time, rather than one implementation or release commitment.

Rendered Preview of the high level design https://github.com/pierDipi/models-as-a-service/blob/16365f9add9eecdc54320d524ccc821fd65dc2c5/docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md

## Documentation

* Added architecture proposals for Responses API enablement, persistence, security, authorization, streaming, and lifecycle management.
* Documented tenant, model, subscription, and model-entry ownership and capability modes, including validation and fail-closed behavior.
* Defined AI guardrail attachment, inheritance, deterministic check selection, authorization, and sequential evaluation across scopes.
* Documented integration with guardrail services, request processing, reconciliation, and retention policies.
* Described deferred expansion plans for storage, capability discovery, agentic features, MCP, file search, and advanced guardrail policies.


docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+Responses and binds its storage; the model declares its Responses protocol support. All three can attach reusable
+`AIGuardrail` policies, with an additional attachment on each subscription's model entry. A guardrail policy identifies
+ordered checks and references the `NemoGuardrails` resource that serves them. AI Gateway reconciles that provider
+binding independently and also owns AITenant reconciliation. MaaS validates model/subscription attachments and composes
+them with the accepted tenant baseline.
+
+```mermaid
+flowchart TD
+    T[AITenant] -->|platform default or future override| D[Responses connection and CA Secrets]
+    T -->|guardrail attachments| G[AIGuardrail]
```
mariusdanciu: Since this is a MaaS owned CR, the name should be MaasGuardrail in order to be consistent.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+The proposal connects three configuration scopes: `AITenant`, `MaaSModelRef` and `MaaSSubscription`. The tenant enables
+Responses and binds its storage; the model declares its Responses protocol support. All three can attach reusable
+`AIGuardrail` policies, with an additional attachment on each subscription's model entry. A guardrail policy identifies
+ordered checks and references the `NemoGuardrails` resource that serves them. AI Gateway reconciles that provider
+binding independently and also owns AITenant reconciliation. MaaS validates model/subscription attachments and composes
+them with the accepted tenant baseline.
+
+```mermaid
+flowchart TD
+    T[AITenant] -->|platform default or future override| D[Responses connection and CA Secrets]
```
mariusdanciu: I don't think that the AITenant should point to MaasGuardrails CR. This would create a dependency on the cluster admins. Instead MaasTenantConfig should point to the MaasGuardrail CR as a tenant admin can set this up

pierDipi: that's why AIGuardrail is in the AI GW controller, there is no MaaSGuardrail anymore, MaaS resources point to AIGuardrail

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: no, I'm proposing it to be in AI Gateway, the main reasoning is that MaaS will depend on AI Gateway resources (AI Tenant, AI Guardrail, etc) instead of being the other way around since AI GW controller will have to know about guardrails.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+  semantics.
+- Target the supplied NeMo v1 checks contract with approved, reusable policies.
+- Prevent an unsupported protocol, modality, streaming mode or tool path from silently bypassing an effective check.
+- Make policy provenance, runtime readiness and propagation observable without disclosing private governance or prompt
+  content.
+- Stage implementation so every advertised capability has integration coverage.
+
+## Non-Goals
+
+- Implement the CRDs, controllers or runtime in this documentation change.
```
mariusdanciu: I think we should define in the ADR the structure of MaasGuardrails as as per out group discussion.

```
apiVersion: maas.opendatahub.io/v1alpha1
kind: MaasGuardrails
metadata:
  name: maas-regex-pii
  namespace: team-a
spec:
  configs:
    - name: regex
    - name: pii

```

pierDipi: yes, it's below

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+[TrustyAI integration contract](#trustyai-integration-and-deployment-topology) before the policy becomes ready.
+
+**Define the reusable check separately.** The policy names the NeMo configuration and evaluation model; it does not
+contain the server's configuration files. This example places NeMo in the tenant namespace, so the default `Same`
+consumer permission suffices. A central server can instead use the
+[NeMo consumer-permission contract](#nemo-owned-consumer-permission).
+
+```yaml
+apiVersion: aigateway.opendatahub.io/v1alpha1
+kind: AIGuardrail
```
mariusdanciu: Is this reconciled by MaaS or AIGateway controller ? If it's Maas I think we should be consistent with Maas prefix naming and api group.

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: If we do MaaSGuardrail, we end up creating circular dependencies between AI GW controller and MaaS controller

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: AI GW controller

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+        name: nemo-client-credentials
+        key: token
+      caBundleRef:
+        name: nemo-client-ca
+        key: ca.crt
+    timeout: 5s
+  checks:
+    - name: sensitive-data
+      configId: pii
+      model: approved-check-model
```
mariusdanciu: Isn't the model specified in the actual NeMo configmap ? Isn't it sufficient to have a list of NeMo config IDs ?

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+apiVersion: aigateway.opendatahub.io/v1alpha1
+kind: AIGuardrail
+metadata:
+  name: application-safety-v1
+  namespace: <tenant-namespace>
+spec:
+  provider:
+    nemo:
+      ref:
+        name: tenant-nemo
```
mariusdanciu: This may be a gap. The NemoGaurdrail CR points to the NeMo service deployment. The deployment should not be in the tenant namespace but rather in the `ai-gateway-infra` namespace where the maap-api POD also live. As per previous dicisions infrastructure should be separated from the tenant namespace. We need to clarify in the NemoGuardrails CR can live in the tenant namespace but the actual service to be created in the `ai-gateway-infra` namespace.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+spec:
+  payloadProcessing:
+    type: praxis
+  responses:
+    enabled: true
+    storage:
+      mode: PlatformDefault
+      deletionPolicy: Retain
+    retention:
+      maxAge: 168h
```
mariusdanciu: Is this the responses state retention ? If so, maybe we should be a bit more explicit. 

Again the question here is if the platform admin should set this or the tenant admin. If it is in the tenant admin then response should be applied in the MaasTenantConfig CR.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+  guardrails:
+    required:
```
mariusdanciu: Why is "required" necessary here ?

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+    name: qwen3-instruct
+  capabilities:
+    responses:
+      mode: ChatCompletions
+  guardrails:
+    required:
+      - name: model-safety
+        ref:
+          name: model-safety-v1
+          scope: Tenant
```
mariusdanciu: What is the role of scope here ?

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+```yaml
+kind: MaaSSubscription
+metadata:
+  name: application-subscription
+  namespace: <tenant-namespace>
+spec:
+  modelRefs:
+    - name: granite-7b
+      namespace: <model-namespace>
+      guardrails:
```
mariusdanciu: Shouldn't we (also) have `guardrails` at top level so that the set of guardrails would apply to all models in the subscription ? It is true that different models may require different guardrails but if not, setting this once per subscription would make the api less verbose and easier to manage by admins.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+   accepted configuration. Praxis verifies compatibility with its loaded generation and executes the generated filters,
+   including authorized history retrieval, applicable checks, inference and persistence of approved content.
+4. Retrieval, continuation and deletion use the same ownership contract. Disabling Responses withdraws both API surfaces
+   and retains data according to the lifecycle policy.
+
+**Guardrail publication and execution**
+
+1. A policy administrator publishes an AIGuardrail referencing a NeMo service and ordered checks; resource editors
+   attach the policy at the required scopes.
+2. AI Gateway reconciles AIGuardrail, validates its NeMo reference/consumer permission and publishes its binding status.
```
mariusdanciu: I'm still not clear if Guardrails should be MaaS or AIGateway level. It seems to me that we may be leaking control plane aspects here out of MaaS. I mean AIGuardrails as defined here won't work without MaaS. This is why it feels that we are creating more control plane divergence.

docs/content/architecture-internals/responses-and-guardrails.md
```diff
@@ -0,0 +1,2710 @@
...
+state.
+
+```yaml
+kind: AITenant
+spec:
+  responses:
+    enabled: true
+    storage:
+      mode: ExternalPostgres
+      externalPostgres:
```
mariusdanciu: I wonder if this is needed since the managed Postgres option (the future DB claims approach) is not really managed DB. It is still the client who provides and manages the DB. I'm not sure if this distinction is necessary.

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: because there could be `required` guardrails and `default` guardrails, required cannot be overridden by lower levels (subscription, models) while `defaults` can

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: As discussed, changed to `namespace: <tenant-namespace>`

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: As discussed in other threads and in the meeting, AIGuardrail will be an AI GW control plane responsibility, and MaaS resources will only reference it

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: Removed the distinction, now we have only `guardrails: [ ref: .., checks: [ ... ] ]` which are required checks and add at every layer

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: These modes describe how the runtime obtains its database binding, not necessarily who operates PostgreSQL.

pierDipi: but this is really a future expansion to make the point of having mode being an enum vs boolean, moved to future expansion separate from the main doc(s)

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: As discussed, AI Guardrail in under the AI GW controller umbrella, MaaS resources refer to AIGuardrail

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: yes, this is mentioned in `Ownership and API placement`, the example uses the per-model refinement within a subscription but I'll add that to the API samples/overview

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,610 @@
...
+  guardrails:
+    - ref:
+        name: application-safety-v1
+        namespace: <tenant-namespace>
+      checks: [ application-check ]
+```
+
+**Register the model and declare how it handles Responses.** Create the referenced `LLMInferenceService` first.
+`ChatCompletions` is the default adapter mode; `Native` is explicit opt-in, and embedding-only or reranker-only models
+can declare `Unsupported`. The model already inherits the tenant's required baseline and need not repeat that
```
mariusdanciu, line 346: A few notes:
1. This enforces embeddings, re-ranker and other types of models to explicitly declare Unsupported. I would prefer if the capablities object is missing, the current behavior is applied, This also helps with backward compatibility.
2. What is the difference between `ChatCompletions` and `Native` ? Is this about responses being managed by the gateway agentic loop vs passthrough ? If so can we please clarify the meaning in the doc ?

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
```diff
@@ -0,0 +1,1329 @@
...
+
+All five attachment locations use a list of entries containing `ref` and optional `checks`:
+
+```yaml
+kind: MaaSSubscription
+spec:
+  guardrails:
+    - ref:
+        name: application-safety-v1
+        namespace: <tenant-namespace>
```
mariusdanciu, line 276: Is the namespace needed here ? I mean I assume that for AI tenant namespaces the AIGuardrail CRs must live in the same namespace with MaasSubscription. Sam would apply to MaasTenantConfig CR. Please let me know if I'm missing something here.

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
```diff
@@ -0,0 +1,1329 @@
...
+in the initial API. [Required/default composition](04-guardrails-future-expansion.md) is deferred.
+
+## API validation and compatibility contract
+
+The field sketches use the current `maas.opendatahub.io/v1alpha1` resource model; adding them still requires generated
+CRDs, defaulting/validation and controller support. Do not use an opaque Praxis YAML field as the user-facing policy
+API. Unknown policy fields must produce actionable validation errors rather than being silently interpreted as an
+unguarded request.
+
+| Field or combination                                                                | Proposed validation/default                                                                                              |
```
mariusdanciu, line 310: This is a flat table for fields coming from different CR, Can we please add a column (or something) indicating which CR each field pertains to ?

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
```diff
@@ -0,0 +1,1329 @@
...
+              metrics: false
+            x-aigateway-guardrail-RENDER_MODEL_CHECK_ID:
+              plain:
+                expression: 'auth.metadata.maasDecision.modelSafetySelected ? "true" : "false"'
+              metrics: false
+            X-MaaS-Policy-Revision:
+              plain:
+                selector: auth.metadata.maasDecision.revision
+              metrics: false
+```
```
mariusdanciu, line 699: Seeing 
- RENDER_APPLICATION_CHECK_ID
- RENDER_SUBSCRIPTION_CHECK_ID
- RENDER_MODEL_CHECK_ID
- RENDER_PRIVACY_CHECK_ID

Is a bit confusing to me. These are just placeholders to calculate a digest as:
```
SHA-256('["<namespace>","<AIGuardrail name>","<check id>"]')
```
so all are calculated in exactly the same way in order to construct the header name and the praxis filter config.

```
x-aigateway-guardrail-<digest>
```

But the document can reference this as a single placeholder as RENDER_CHECK_ID, can it not ? ...and simply provide the formula for it. There are still N headers with N separate digests depending on the number of guardrails checks configured.

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
pierDipi: > I would prefer if the capablities object is missing

how do we solve the challenge of rejecting requests for `/v1/responses`, or in general "not allowed API endpoints" for a given model? For this case, we don't really have a current behavior as baseline, it would run the responses flow to then fail the request at the model server level, ideally, requests are rejected at the GW level without wasting cycles on unnecessary work for a model that will fail the request.

> What is the difference between ChatCompletions and Native ? Is this about responses being managed by the gateway agentic loop vs passthrough ?

yes, Native means `/v1/responses` is fully supported, `ChatCompletions` means rehydration and translation needs to be applied at the GW level, I'll clarify the text

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
pierDipi: no it's not needed for MaaSSubscription, it uses the AI Tenant target namespace

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
pierDipi: Done

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
pierDipi: Done

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: Added sample including subscription-wide guardrail alongside per-model overrides

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: Clarified

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: sample now uses separate namespace placeholder for NemoGuardrail CR, allowing it to live in a separate namespace while being referenced in AIGuardrail as full qualified reference (including namespace)

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
mariusdanciu: In what context would be needed? Also for MaasModelRef isn't the expectation that AIGuardrail CR is in the same namespace with MaasModelRef ?

docs/content/architecture-internals/proposals/guardrails-responses/02-responses-low-level-details.md
```diff
@@ -0,0 +1,598 @@
...
+- [Stateful operations and persistence transactions](#stateful-operations-and-persistence-transactions)
+- [Streaming and usage](#streaming-and-usage)
+- [Request-path security and privacy contract](#request-path-security-and-privacy-contract)
+- [Responses and Conversations using existing filters](#responses-and-conversations-using-existing-filters)
+- [Tenant capability state transitions](#tenant-capability-state-transitions)
+- [Reviews](#reviews)
+
+## Responses enablement and lifecycle
+
+The [tenant example](01-guardrails-responses-high-level-design.md#proposal-through-resource-examples) shows the initial
```
jland-redhat, line 34: We already have a convention of storing DB info in a secret and keying off of that.

I understand the idea behind this but why do we want to change that now? 

It means MaaS either needs to be really careful with how it migrates or we need to support two different ways to connect to a database. I don't know I see a strong enough reason here to make that change. At least not in the short term.

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,610 @@
...
+“Tenant” is a logical scope, not a recommendation to extend the legacy `Tenant`
+resource. The current code separates `AITenant` (Gateway, OIDC, payload-processing selection) from `MaasTenantConfig`
+(MaaS-specific settings). Follow that separation:
+
+| Resource                                       | Proposed responsibility                                         | Editor                          |
+|------------------------------------------------|-----------------------------------------------------------------|---------------------------------|
+| `AITenant.spec.responses`                      | Enable Responses and provision its runtime/storage dependencies | Platform administrator          |
+| `AITenant.spec.guardrails`                     | Platform-admin baseline for the tenant                          | Platform administrator          |
+| `MaasTenantConfig.spec.guardrails`             | Tenant-admin baseline across MaaS subscriptions and models      | Tenant administrator            |
+| `AIGuardrail.spec`                             | Reusable ordered checks and a typed NeMo server reference       | Authorized policy administrator |
```
jland-redhat, line 111: May be worth making a note that this will (I assume) live in the tenant relevant namespace. This is how the MaaSTenantConfig/MaaSAuthPolicy/MaaSSubscriptions all work.

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,610 @@
...
+apiVersion: maas.opendatahub.io/v1alpha1
+kind: MaasTenantConfig
+metadata:
+  name: default-tenant
+  namespace: <tenant-namespace>
+spec:
+  guardrails:
+    - ref:
+        name: application-safety-v1
+        namespace: <tenant-namespace>
```
jland-redhat: Why would we need to specify the namespace? 

We don't want cross namespace anyway right? We would want the Guardrail object to be locked to the tenant namespace in order to be recognized

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,610 @@
...
+```yaml
+kind: MaaSModelRef
+metadata:
+  name: granite-7b
+  namespace: <model-namespace>
+spec:
+  modelRef:
+    kind: LLMInferenceService
+    name: granite-7b-instruct
+  capabilities:
```
jland-redhat, line 377: If I am understa

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,610 @@
...
+  name: qwen3
+  namespace: <model-namespace>
+spec:
+  modelRef:
+    kind: LLMInferenceService
+    name: qwen3-instruct
+  capabilities:
+    responses:
+      mode: ChatCompletions
+  guardrails:
```
jland-redhat, line 400: If I am understanding `guardrails` correctly here we are saying these are the guardrails that you can use for this model but won't be applied automatically.

If that is the case I would put `guardrails` section under `capabilities` because I think that it would make that more obvious.

docs/content/architecture-internals/proposals/guardrails-responses/02-responses-low-level-details.md
```diff
@@ -0,0 +1,598 @@
...
+Files, document extraction, vector stores, web search and MCP remain future capabilities requiring separate API and
+binding designs, trusted authentication, egress policy, size/time limits and advertised protocol coverage.
+The [deferred agentic flows](03-responses-future-expansion.md#deferred-agentic-flows)
+preserve future architectural context and existing Praxis building blocks, not initial feature availability.
+
+## Responses database architecture and enterprise isolation
+
+### Decision and workload boundaries
+
+The initial baseline is **a shared Responses database and connection binding for enabled tenants**, separate from the
```
jland-redhat: So to make sure I am clear here, tenant's can share a database but they would not have to right?

jland-redhat:
Overall this looks good to me. I've left a few comments and would like to talk it through to make sure I understand the full plan.

One follow-up I'd like captured in future expansion: rate limiting / quota enforcement inside agentic loops (per-iteration metering, frozen subscription for an admitted loop, guarding against iterative amplification). The v1 doc flags the Responses quota gap, but since agentic flows are deferred, it'd help to name loop-internal enforcement as an explicit future deliverable in 03-responses-future-expansion.md.

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
pierDipi: yes it's clarified below

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
pierDipi: this got truncated?

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
pierDipi: > If I am understanding guardrails correctly here we are saying these are the guardrails that you can use for this model but won't be applied automatically.

can you clarify from where this is coming from?

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
pierDipi: yes, this is to make it explicit how it gets resolved and it is clarified below that cross-namespace is only an option for MaaSModelRef (still locked to either target tenant namespace or MaaSModelRef model namespace).

I'll remove it to reduce confusion

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
jland-redhat: Yeah and I no longer know what I was going for so lets go with this as solved :) 

I think it was just a clarification to on the additive capabilities

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
jland-redhat: Nvm on this, this was on a first pass and I got confused as to the requirement

docs/content/architecture-internals/proposals/guardrails-responses/02-guardrails-low-level-details.md
pierDipi: the idea was that MaasModelRef could also re-use Guardrails provided by the tenant admin(s) sort of "as a service" and in that case MaasModelRef could point to `<tenant-namespace>`

docs/content/architecture-internals/responses-and-guardrails.md
pierDipi: As we discussed, model is not removed and will be "hard coded" for the short term

docs/content/architecture-internals/proposals/guardrails-responses/02-responses-low-level-details.md
pierDipi: Clarified that we will use the same existing pattern, other options moved to future considerations

pierDipi:
@mariusdanciu @jland-redhat any more comments?

mariusdanciu (approved):
lgtm

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,621 @@
...
+
+`Selector` permits AIGuardrails in the named tenant namespace to reference this server across namespaces. The selector
+matches the Namespace object's standard name label. Omitting `allowedConsumers` defaults to `Same` and would reject
+these cross-namespace references. This permission field is proposed, not an existing TrustyAI capability. The server's
+endpoint, authentication and successful configuration loading must satisfy the
+[TrustyAI integration contract](02-guardrails-low-level-details.md#trustyai-integration-and-deployment-topology) before
+the policy becomes ready.
+
+NeMo owns model selection for LLM-based rail tasks. The [checks contract](02-guardrails-low-level-details.md#mapping-to-the-nemo-api)
+uses `/v1/guardrail/checks` with fixed `model: check-model`; AIGuardrail selects configurations and phases without a model field.
```
christinaexyou: this should be `v1/checks` - `v1/guardrails/checks` bc it's expected to be deprecated upstream. the guardrails filter in PraxisAI has already migrated to `v1/checks` https://github.com/praxis-proxy/ai/issues/1044#event-31310856724

pierDipi:
```suggestion
uses `/v1/checks` with fixed `model: check-model`; AIGuardrail selects configurations and phases without a model field.
```

pierDipi: fixed

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,621 @@
...
+
+**Define the reusable check separately.** The policy names the NeMo configuration and check phases; it does not contain
+the server's configuration files. All three AIGuardrails remain in `<tenant-namespace>` and explicitly reference NeMo in
+`<guardrails-namespace>`. The server's selector authorizes that provider reference under the
```
christinaexyou, line 214: should we restrict the guardrails-namespace to be the MaaS `--infra` namespace ? afaik all services need to be deployed here. alt. we can add a new field, `serverNamespace` to the guardrails spec so the server doesn't have to be in the same ns as the CR

pierDipi: I don't think we want to prescribe a specific topology, surely, using the AITenant target namespace for the guardrails namespace is probably what we would recommend

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,621 @@
...
+[NeMo consumer-permission contract](02-guardrails-low-level-details.md#nemo-owned-consumer-permission). Credential and
+CA references remain local to each AIGuardrail's tenant namespace; model and subscription attachments reference
+AIGuardrails by name, with the namespace resolved under
+the [scoped reference rules](02-guardrails-low-level-details.md#scoped-policy-references). AITenant, MaaSSubscription and
+MaasTenantConfig must not specify `ref.namespace`: they always use the accepted AITenant target namespace. MaaSModelRef
+must explicitly select its own namespace or that tenant target namespace; no other namespace is allowed.
+
+```yaml
```
christinaexyou, line 222: having 3 instances of AIGuardrails seems redundant to me. why not have a single instance for application-check, subscription-check, and model-safety e.g ?

```
apiVersion: aigateway.opendatahub.io/v1alpha1
kind: AIGuardrail
metadata:
  name: application-safety-v1
  namespace: <tenant-namespace>
spec:
  provider:
    nemo:
      ref:
        name: tenant-nemo
        namespace: <guardrails-namespace>
      credentialsSecretRef:
        name: nemo-client-credentials
        key: token
      caBundleRef:
        name: nemo-client-ca
        key: ca.crt
    timeout: 5s
  checks:
    - name: application-check
      configId: application-safety
      phases: [ Input ]
    - name: subscription-check
      configId: subscription-safety
      phases: [ Input ]
    - name: model-check
      configId: model-safety
      phases: [ Input ]
```

pierDipi: it's a totally valid option, not something we should have a strict opinion on, they should create their AIGuardrail policies as they prefer to manage them, this is only an example to show the possibilities

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
```diff
@@ -0,0 +1,621 @@
...
+        namespace: <guardrails-namespace>
+      credentialsSecretRef:
+        name: nemo-client-credentials
+        key: token
+      caBundleRef:
+        name: nemo-client-ca
+        key: ca.crt
+    timeout: 5s
+  checks:
+    - name: sensitive-data
```
christinaexyou, line 242: at level of MaaS would `sensitive-data` be applied to ?

docs/content/architecture-internals/proposals/guardrails-responses/01-guardrails-responses-high-level-design.md
mariusdanciu: True but NeMo controller should support this first. This is not a blocker though.

christinaexyou (approved):

jland-redhat (approved):
Lgtm

PR #1505, opened 2026-09-17
fix(tenantreconcile): skip praxis-owned writer pods during IPP migration

jland-redhat:
## Summary

- Fix `ensureIPPWritersStopped` so praxis tenants are not blocked when ai-gateway-controller owns the payload-processing Deployment but pods only carry tenant-instance labels (no `managed-by`).
- Skip pod checks for writer Deployment refs when the Deployment exists and `!isMaaSOwnedIPPResource`, matching the existing deployment-level ownership gate.
- Add regression test for praxis-owned Deployment + payload-processing pod without `app.kubernetes.io/managed-by`.

## Test plan

- [x] `go test ./pkg/platform/tenantreconcile/ -count=1`
- [ ] Verify praxis tenant with `payload-processing-type=praxis` completes IPP migration cleanup without `PlatformReconcileFailed`
- [ ] Confirm legacy IPP tenants still defer route cleanup while MaaS-owned writer pods remain

## Risk analysis

**Risk rating:** 2

**Why:** Small, targeted change in praxis migration cleanup with unit test coverage. Behavior only changes when a non-MaaS-owned writer Deployment is present; absent-deployment pod checks are unchanged. Not exercised by default Prow smoke unless praxis migration path runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Migration no longer remains blocked by legacy writer pods when their Deployment is owned by Praxis or another non-MaaS controller.
  * Writer Deployments that cannot be retrieved due to errors continue to be reported appropriately.

* **Tests**
  * Added coverage for Praxis-owned Deployments containing writer pods.


jland-redhat:
@coderabbitai review

PR #1516, opened 2026-09-20
feat: harden IPP backend swap with ipp-migration-cleanup-complete

jland-redhat:
## Summary

Re-opened copy of https://github.com/opendatahub-io/models-as-a-service/pull/1508 (Ryan's `rq-95503`) from `jland-redhat` so CI can re-run cleanly.

- `payload-processing-status` in MaasTenantConfig is the handoff boundary: set to `cleanup-complete` after IPP cleanup; scraped when tenant reconciler starts deploying.
- For a new tenant, AITenant controller annotates `payload-processing-status=cleanup-complete` so initial IPP deploy is not blocked (new tenants only; not normal reconciliation).
- On IPP cleanup, also delete unmanaged resources.

Related: https://github.com/opendatahub-io/ai-gateway-controller/issues/34

## How Has This Been Tested?

Carried from #1508:

- [x] legacy -> praxis swap
- [x] praxis -> legacy swap
- [x] Rapid flip (legacy -> praxis -> legacy)
- [x] Steady state not restricted by annotation
- [x] New tenant not blocked

## Risk analysis

- **Risk rating**: 4
- **Why**: IPP backend swap / tenant reconcile path is tightly coupled to gateway controller behavior and is unlikely to be fully covered by the default Prow smoke path; migration races and cleanup of unmanaged resources have high blast radius if wrong.

## Test plan

- [ ] Confirm CI (unit / e2e / prow) re-runs green on this fork PR
- [ ] Spot-check IPP legacy ↔ praxis swap scenarios from #1508 if CI alone is insufficient
- [ ] Close or supersede #1508 once this PR is the canonical vehicle

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added coordinated status tracking for switching between legacy IPP and praxis payload processing.
  * Added safeguards to prevent simultaneous ownership during payload-processing transitions.

* **Bug Fixes**
  * Tenant configuration now exclusively controls payload-processing mode.
  * Improved cleanup of unmanaged legacy IPP resources during transitions.
  * Reconciliation now waits when another process is completing a payload-processing handoff.
  * Existing tenant-config status is preserved during subsequent reconciliations.

PR #1506, opened 2026-09-17
test(e2e): enable x-api-key auth in CI smoke suite

jland-redhat:
## Summary

- Add `test_x_api_key_auth.py` to the default `run_e2e_tests.sh` smoke file list.
- Exercises IPP ExternalModel-driven `api-keys-x-api-key` gateway identity discovery and x-api-key header authentication in Prow.
- Test is `@serial` and runs in pass 2 (serial cluster mutators) alongside other API-key flows.

## Test plan

- [ ] Prow E2E pass 2 includes `test_x_api_key_auth.py` (skipped automatically when `externalmodels.inference.opendatahub.io` CRD is absent)
- [ ] `test_x_api_key_authenticates` and rejection cases pass on standard fixture deploy (`simulator-subscription`, `facebook-opt-125m-simulated`)

## Risk analysis

**Risk rating:** 3

**Why:** CI-only change, but adds serial E2E coverage that mutates gateway AuthPolicy via a temporary IPP ExternalModel CR. Depends on IPP ExternalModel CRD and standard deploy-models fixtures; not exercised when the CRD is missing (module skip). Local dev cluster run hit 403 during fixture warmup before this PR — Prow full deploy path is the authoritative check.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Tests**
  * Added API key authentication coverage to the default end-to-end test runs.
  * The test runs in both serial and parallel configurations unless specific test paths are provided.

PR #1517, opened 2026-09-20
fix(e2e): wait for x-api-key auth before asserting

jland-redhat:
## Summary

- Fixes flaky `test_x_api_key_authenticates` empty-body 401 after #1506 re-enabled the suite in CI smoke.
- Root cause: fixture waited for `api-keys-x-api-key` in AuthPolicy YAML and only polled Bearer readiness; Kuadrant/Authorino can still lag, so the first single-shot x-api-key call fails with an empty 401.
- Fixture now waits for gateway AuthPolicy Enforced and polls with `_inference_x_api_key` until 200 (same transient-401 handling as `_poll_status`).

Unrelated to IPP backend-swap work (#1508 / #1516); reproduced on cluster with identity already present while x-api-key still returned empty 401 for ~4s.

## Risk analysis

- **Risk rating**: 1
- **Why**: E2E test-only change; no product runtime behavior change. Slightly longer fixture setup when AuthPolicy propagation is slow.

## Test plan

- [x] Reproduced race: AuthPolicy listed `api-keys-x-api-key` while x-api-key inference returned empty-body 401, then 200 within a few seconds
- [ ] CI serial smoke pass includes `test_x_api_key_auth.py` green
- [ ] Spot-check module fixture still tears down ExternalModel / removes identity source


jland-redhat:
@coderabbitai review

chaitanya1731 (approved):
Lgtm

PR #1520, opened 2026-09-21
feat(RHOAIENG-95879): add optional inferencePriority to MaaSSubscription

vivekk16:
## Summary

Adds optional `spec.requestPriority` (signed int32) to the MaaSSubscription CRD. Field only — stored and validated, not yet read by anything.

## What changed

- `maassubscription_types.go` — `RequestPriority *int32`, int32 `Minimum`/`Maximum`, no default
- CRD YAML + `zz_generated.deepcopy.go` — generated
- `docs/content/reference/crds/maas-subscription.md` — one table row

## Why pointer, why no default

Absent, explicit `0` and non-zero must stay distinguishable. A value type would drop an explicit `0`; a default would make "absent" unreachable. Validation is via the structural schema on CREATE and UPDATE; webhook unchanged.

## Testing

`lint`, `test`, `validate-manifests.sh`, `verify-codegen` pass. No RBAC changes.

## Summary by CodeRabbit

* **New Features**
  * Added the optional `inferencePriority` setting for controlling inference request scheduling.
  * Supports signed 32-bit values, including negative priorities for scheduling below the default.
  * Keeps inference scheduling priority separate from subscription selection priority.
  * When omitted, the scheduler uses priority 0 without creating an inference objective; explicitly setting `0` creates one with priority 0.

* **Breaking Changes**
  * Replaced the `requestPriority` setting with `inferencePriority`.


deployment/base/maas-controller/crd/bases/maas.opendatahub.io_maassubscriptions.yaml
```diff
@@ -143,6 +143,19 @@ spec:
                   Higher numbers have higher priority. Defaults to 0.
                 format: int32
                 type: integer
+              requestPriority:
```
pierDipi: it was suggested to rename the field to `inferencePriority` https://github.com/opendatahub-io/architecture-decision-records/pull/169#discussion_r4044733754

vivekk16: Addressed

pierDipi (approved):
/lgtm

jland-redhat (approved):
lgtm

